PR #1, opened 2018-05-15
[Button] Renamed prop bordered to shadowed and added it to secondary

wakooka:


loick:
Update the changelog please :)

PR #2, opened 2018-05-15
Update Button and ItemChoice API to be more restrictive

PerrineBoissieres:
Still need some advice, and maybe you have opinions on other props to update. Ex: I'm not very fan of the `validated` prop, so maybe another name like `onValidation`?

The main objective of this change is to make sure that different styles won't overlap (a `Button` cannot be `primary` and `secondary` at the same time, for example). This simplifies the external API a bit, and avoid issues like style collision.


pith:
onCheckAnimationEnd ?

src/autoComplete/autoCompleteList.tsx
```diff
@@ -127,7 +127,7 @@ extends Component <AutoCompleteListProps, AutoCompleteListState> {
             loading={index === this.props.loadingItemIndex}
             select={this.props.onSelect}
             valid={index === this.props.loadingItemIndex && this.props.valid}
```
pith: I think you changed valid for status

PerrineBoissieres: Hmm... It's the ItemChoice API, not the Button API. I'm not sure I want to impact too many components at once in this PR.

PerrineBoissieres: Ok, I went into it: it's a property that adds a Button with the `CHECKED` status inside the `ItemChoice` (I think it's for the case with a loading ItemChoice), so I could probably rename it to `checked` or something, for consistency sake.

src/autoComplete/autoCompleteList.tsx
```diff
@@ -118,22 +118,30 @@ extends Component <AutoCompleteListProps, AutoCompleteListState> {
...
+            <AutoCompleteListItem
+              key={this.props.itemKey(item)}
+              item={item}
+              className={this.props.itemClassName}
+              highlighted={index === this.state.highlightedIndex}
+              status={status}
+              select={this.props.onSelect}
+              onDoneAnimationEnd={this.props.onDoneAnimationEnd}
+            >
+              <div>
```
loick: Just wondering, but is the div useful here?

src/autoComplete/story.tsx
```diff
@@ -69,6 +69,7 @@ class AutoCompleteExample extends Component<AutoCompleteExampleProps, AutoComple
           searchForItemsMinChars={number('searchForItemsMinChars', 3)}
           loadingItemIndex={number('loadingItemIndex', -1)}
           valid={boolean('valid', false)}
+          onDoneAnimationEnd={() => console.log('HEEEEY')}
```
loick: A more explicit log would be better ;)

src/button/index.tsx
```diff
@@ -9,26 +9,30 @@ import Loader from 'loader'
...
   readonly href?: string | JSX.Element,
   readonly children?: string | number | React.ReactNode,
   readonly className?: Classcat.Class,
   readonly title?: string,
-  readonly primary?: boolean,
-  readonly secondary?: boolean,
-  readonly loading?: boolean,
-  readonly valid?: boolean,
-  readonly warning?: boolean,
+  readonly status?: ButtonStatus,
```
loick: is it really optional?

src/autoComplete/story.tsx
PerrineBoissieres: Forgot to remove it after my tests, haha ^^"

src/button/index.tsx
PerrineBoissieres: I put it optional to have a default value (which is primary, by a bit of an arbitrary decision). I agree that most of the times you'll precise it, but if you just want a simple Button without any big status changing, it's easier to write with a default value.

PR #4, opened 2018-05-17
[BBCSPA-1302] New Modal Component

nickoola:


package.json
```diff
@@ -114,8 +116,12 @@
       ".(ts|tsx)": "<rootDir>/bin/jest.preprocessor.js",
       "jestScript.js|utils.js": "<rootDir>/node_modules/babel-jest"
     },
-    "modulePaths": ["<rootDir>/src/"],
-    "collectCoverageFrom": ["src/**/index.{ts, js}"],
+    "modulePaths": [
```
loick: why? :)

src/modal/index.unit.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
+import Modal from 'modal'
+import Button from 'button'
+
+describe('<Modal>', () => {
+  it('Should have default props with appropriate values and content', () => {
+    const wrapper = mount(<Modal isOpen={true}><p>My modal</p></Modal>)
+    expect(wrapper.prop('fullscreen')).toEqual(false)
```
loick: if this is a plain value, use `toBe`. Here you can use an object to test all of the props, and use `toEqual`

nickoola: Done

package.json
nickoola: Done

src/modal/index.unit.tsx
```diff
@@ -0,0 +1,50 @@
...
+    expect(wrapper.find('.kirk-modal-dimmer--visible').length).toEqual(1)
+  })
+  it('Should be not visible if isOpen is set to false', () => {
+    const wrapper = mount(<Modal />)
+    expect(wrapper.find('.kirk-modal-dialog').length).toEqual(0)
+    expect(wrapper.find('.kirk-modal-dimmer--inactive').length).toEqual(1)
+  })
+  it('Should be visible if isOpen is set to true', () => {
+    const wrapper = mount(<Modal isOpen={true} />)
+    expect(wrapper.find('.kirk-modal-dialog').length).toEqual(1)
```
loick: toBe

src/transitions/index.tsx
```diff
@@ -0,0 +1,53 @@
+import React, { Fragment } from 'react'
+import Transition  from 'react-transition-group/Transition'
+import cc from 'classcat'
+import { transition } from '_utils/branding'
+import style from './style'
+
+export enum AnimationType {
```
PerrineBoissieres: As it is, those constants are not available when importing the library in another project (because only the `default` export is available in the index file). I think they should be accessible when using the `Transition` component in another project.
You have two solutions:
- add the `AnimationType` in the main `src/index` file
- or attach them as a `static` to the `CustomTransition` class

I went with the second solution on my PR for Button and ItemChoice API, but it's your choice as my PR has not been merged yet.

nickoola: I attach TYPES as `static` and to make it work I was constraint to convert the stateless component to a class. 
In case of simple function, I tried to assign TYPES like this `CustomTransition.TYPES =  AnimationType ` but don't find a way to declare TYPES to typescript. If anyone has ideas to keep it stateless...

src/modal/index.unit.tsx
nickoola: Done.

nickoola:
All feedbacks applied guys !!!

PR #12, opened 2018-06-05
Add props to Autocomplete + handle disabled state on Textfield

Zarthia:


src/textField/index.tsx
```diff
@@ -184,7 +185,7 @@ export default class TextField extends PureComponent <TextFieldProps, TextFieldS
             />
           }
           {
-            !isTextArea &&
+            shouldDisplayButton &&
```
Tuxosaurus: Question about the existing: why do we always show a button component - albeit an empty one - on all inputs?

Zarthia: That is a good question, that's the way it is right now

src/autoComplete/index.tsx
```diff
@@ -90,6 +96,12 @@ export default class AutoComplete extends Component<AutoCompleteProps, AutoCompl
     defaultValue: '',
     showList: true,
     valid: false,
+    autoCorrect: 'off',
```
loick: Wouldn't it be better to directly extend the Textfield props? Or are there some props you don't want to have

src/autoComplete/index.tsx
```diff
@@ -219,11 +231,15 @@ export default class AutoComplete extends Component<AutoCompleteProps, AutoCompl
           defaultValue={String(this.state.query)}
           addon={this.props.inputAddon}
           inputRef={this.inputRef}
-          autoCorrect="off"
-          autoComplete="off"
+          autoCorrect={this.props.autoCorrect}
+          autoComplete={this.props.autoComplete}
```
loick: autoComplete should always be off IMO

Zarthia: On the Autocomplete, I agree. Is that a problem @dagatsoin ?

dagatsoin: No.

src/autoComplete/index.tsx
Zarthia: Do you mean those?
```
export interface CommonFormFields {
  name: string,
  id?: string,
  type?: 'text' | 'password' | 'email' | 'number' | 'search',
  placeholder?: string,
  maxLength?: number,
  autoCorrect?: 'on' | 'off',
  autoComplete?: 'on' | 'off',
  disabled?: boolean,
  readOnly?: boolean,
  autoFocus?: boolean,
  required?: boolean,
  onFocus?: (event:
    React.FocusEvent<HTMLTextAreaElement> | React.FocusEvent<HTMLInputElement>) => void,
  onBlur?: (event:
    React.FocusEvent<HTMLTextAreaElement> | React.FocusEvent<HTMLInputElement>) => void,
}
```

PR #16, opened 2018-06-06
Add callback to PushInfos animation

PerrineBoissieres:
So now we can know when the transition is over :)


src/pushInfo/index.tsx
```diff
@@ -9,26 +9,36 @@ interface PushInfoProps {
...
 
-const PushInfo = ({ className, icon, headline, content }: PushInfoProps) => (
-  <div className={cc(['kirk-pushInfo', className])}>
-    {icon && (
-      <div className="kirk-pushInfo-icon">{icon}</div>
-    )
+class PushInfo extends Component<PushInfoProps> {
+  componentDidMount() {
+    if (this.props.onAnimationEnd) {
+      setTimeout(this.props.onAnimationEnd, 1000)
```
Tuxosaurus: Ok, too bad we can't really have the animation time as a prop, given it's split into 2 steps in the css (700ms + 300ms)

pith: Can you replace this magic number by constants and reuse them in the css ?

Issue #14, opened 2018-06-06
Pass down the bodyClassName to the AutoCompleteList


dagatsoin:
Currently it is not possible to pass a class to the result panel of the AutoCompletion component. So we need to use the `.kirk-autoComplete-list` accessor which may change in the futur and may break the app style.

**Current behavior:**
The property `bodyClassName` is passed down to the loading/noresult panel but not to the AutoCompleteList component.

**Expected behavior:**
The property `bodyClassName` is passed down to the AutoCompleteList component.

Zarthia:
Fixed in v0.2.2

pith:
Fixed here #18

PR #17, opened 2018-06-06
Render emptySearch with an autoCompleteList

pith:
- Reuse the `autoCompleteList` to render the empty search. 
- Fix the condition of display for empty search / item list
- Refactor the loading/checked status of the list items

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-#Unreleased
+# Unreleased
 Breaking changes:
 - [Autocomplete] Render empty search with an autocomplete list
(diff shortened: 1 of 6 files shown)


Zarthia:
Can you update the Changelog as well?
And maybe handle this at the same time :) #14

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
-# Unreleased
-- [...]
+#Unreleased
```
Zarthia: Need a space between `#` and `Unreleased` or else it won't be displayed as a title (markdown stuff)

PR #21, opened 2018-06-13
[BBCSPA-1446] Checkbox position change and more

Tuxosaurus:
Add a prop to invert text and checkbox, defaults to checkbox before text.
Prevent empty nodes when labels are missing.
Unify stories into one entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 - [timePicker] Fix display on small devices
 - Remove deprecated updateColor() method
 - [itemChoice] Fix right content display
+- [Checkbox] Reverse label and input and made it configurable
 
 # v0.3.0 (08/06/2018)
 Breaking changes:
--- a/src/checkbox/index.tsx
+++ b/src/checkbox/index.tsx
@@ -11,7 +11,14 @@ export interface CheckboxProps {
   readonly value?: string,
   readonly checked?: boolean,
   readonly disabled?: boolean,
+  readonly labelDisplay?: labelDisplays,
   readonly onChange?: (obj:onChangeParameters) => void,
+}
+
+export enum labelDisplays {
+  LEFT = 'left',
+  RIGHT = 'right',
+  NONE = 'none',
 }
 
 export interface CheckboxState {
@@ -25,6 +32,7 @@ export default class Checkbox extends PureComponent <CheckboxProps, CheckboxStat
     subLabel: '',
     checked: false,
     disabled: false,
+    labelDisplay: labelDisplays.RIGHT,
   }
   state:CheckboxState = {
     isChecked: this.props.checked,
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 - [timePicker] Fix display on small devices
+- [Checkbox] Reverse label and input and more improvements
```
pith: `Reverse label and input and made it configurable`

src/checkbox/index.tsx
```diff
@@ -11,6 +11,7 @@ export interface CheckboxProps {
   readonly value?: string,
   readonly checked?: boolean,
   readonly disabled?: boolean,
+  readonly labelBeforeInput?: boolean,
```
pith: I would prefer `labelPosition` with an enum left / right

src/checkbox/index.tsx
```diff
@@ -58,9 +60,14 @@ export default class Checkbox extends PureComponent <CheckboxProps, CheckboxStat
   }))
 
   render() {
-    const { className, name, children, value, subLabel } = this.props
+    const { className, name, children, value, subLabel, labelBeforeInput } = this.props
+    const hasLabel = children || subLabel
```
pith: Is there a design where we can have a checkbox without label. Also is it really possible to have a sublabel without label ?

src/checkbox/index.unit.tsx
```diff
@@ -77,3 +77,21 @@ it('should trigger a change event on an async checkbox', () => {
   ))
   checkbox.find('input').simulate('change')
 })
+
+it('should display the label on the left and the checkbox on the right', () => {
```
pith: nit: you could do a snapshot test

src/checkbox/index.tsx
Tuxosaurus: I don't know, and hope that we never have a checkbox without a label, it's only a little change to make sure we don't serve empty markup in this event.
Having a subLabel only is possible with the current implementation so I didn't challenge it.

src/checkbox/index.tsx
Tuxosaurus: I'll do that

CHANGELOG.md
Tuxosaurus: Done

src/checkbox/index.tsx
Tuxosaurus: Done

src/checkbox/index.unit.tsx
Tuxosaurus: Done

src/checkbox/index.tsx
```diff
@@ -58,9 +65,19 @@ export default class Checkbox extends PureComponent <CheckboxProps, CheckboxStat
   }))
 
   render() {
-    const { className, name, children, value, subLabel } = this.props
+    const { className, name, children, value, subLabel, labelPosition } = this.props
+    const hasLabel = children || subLabel
```
loick: hasLabel is always true? 'cause children seems to be mandatory

loick: children.length > 0 || (subLabel && subLabel.length > 0)

nickoola:
Can you add screenshots of both versions ?

Zarthia:
screen shot 2018-06-18 at 10 45 01
Label should be placed right after the checkbox (with the adequate margin)

screen shot 2018-06-18 at 10 48 18
Also, when you want to display the checkbox only, without label/sublabel, I guess we shouldn't have the `<label>` element

Tuxosaurus:
Changed "Position" to "Display" as per @Zarthia's request
Added a "None" display that preserves the existing label structure but makes it "visually hidden"

checkbox_label_left
checkbox_label_right
checkbox_label_none

nickoola:
Just try your update localy on Kiaros before merge.

Tuxosaurus:
Checked on kairos (local), change is OK and doesn't need any adjustment
screen shot 2018-06-19 at 13 47 57
screen shot 2018-06-19 at 14 02 14

PR #22, opened 2018-06-18
Set up array to display places in the Itinerary

Zarthia:


src/itinerary/index.tsx
```diff
@@ -27,49 +26,42 @@ const Itinerary = ({
...
-        <span>{arrival.mainLabel}</span>
-        {
-          (!small && arrival.subLabel) && (
-            <span className="kirk-itinerary-subtext">{arrival.subLabel}</span>
-          )
-        }
-      </div>
-    </li>
+    {
+      places.map((place, index) => (
```
pith: So you display the first and the last twice ?

Why not:

const [first, rest, last] = places

Tuxosaurus: I don't see how it could be displayed twice?

src/itinerary/story.tsx
```diff
@@ -20,20 +20,22 @@ stories.add(
...
-          time: text('Time from', '09:00'),
-        }}
-        arrival={{
-          mainLabel: text('Main label to', 'Bordeaux'),
-          subLabel: text('Sublabel to', 'Gare Bordeaux Saint-Jean'),
-          isoDate: '2017-12-11T12:00',
-          distanceFromPoint: isToDistance && '8km',
-          time: text('Time to', '12:00'),
-        }}
+        places={[
```
loick: Maybe you should add a test including the stopovers

src/itinerary/index.tsx
loick: @pith This wouldn't work :)
`const [first, rest, last] = places`

PR #24, opened 2018-06-19
Fix itemChoice right content display

nickoola:


pith:
Can you add a screenshot please 🙂

src/itemChoice/index.tsx
```diff
@@ -123,10 +123,12 @@ class ItemChoice extends PureComponent <ItemChoiceProps> {
           </div>
         )}
         {children}
-        {rightAddon && (
-          <div className={cc(prefix({ 'itemChoice-rightAddon': true }))}>{rightAddon}</div>
-        )}
-        {rightIcon}
+        <div className="kirk-itemChoice-right">
+          {rightAddon && (
```
wakooka: Is it intended to have an empty `div` in some cases ?

nickoola:
capture d ecran 2018-06-19 a 11 54 39
capture d ecran 2018-06-19 a 11 54 13
capture d ecran 2018-06-19 a 11 53 55
capture d ecran 2018-06-19 a 11 53 39

src/itemChoice/index.tsx
nickoola: It's never empty, we always have the arrow as default.

nickoola: rightAddon is when we want to add a content near the arrow.

wakooka: right, my bad, I didn't see `{rightIcon}` underneath

PR #26, opened 2018-06-21
[BBCSPA-1507] Improve the kirk icons management

loick:
- No more JS error/warnings on Storybook
- Use isEmpty on kirk codebase instead of length calculation everywhere
- Fix the fillRule proptype issue on the kirk icons (no more warning on kairos after that)
- Name the icons function to ease the debug as a client of kirk
- Fix the kirk icon generator accordingly 
- Add the color hexa on the storybook branding page (along with their name)
- Automate the story creation for every icon, with its knobs.
    Benefits: 
        - No more forgotten icons in storybook
        - Everything is fully testable (some knobs were missing)
        - the components now render without any default props in storybook which helps you to see the real default state of it.
    How does it work?
        1 - get the props of the component via its static default props (PureComponent).
        2 - get the type of each prop and create the knob accordingly.
        3 - you can be even more lazier than before and by not updating the story file when creating an icon.


src/autoComplete/autoCompleteList.tsx
```diff
@@ -61,11 +63,10 @@ extends Component <AutoCompleteListProps, AutoCompleteListState> {
   onKeyboardEventArrowDown = (e:Event) => {
     e.preventDefault()
 
-    const itemsLength = this.props.items.length
-    if (!itemsLength) return
+    if (isEmpty(this.props.items)) return
```
pith: nit: avoid oneline if

loick: Another PR :) Here I just replaced elements, and I struggle to merge this PR because it's reinitialising the approvals all the time :D

PR #28, opened 2018-06-25
[BBCSPA-1550] Remove the margin around the profile

loick:


loick:
screen shot 2018-06-25 at 15 06 53
screen shot 2018-06-25 at 15 07 14

PR #31, opened 2018-06-26
[BBCSPA-1560] Add info icon

loick:


bin/kirkHelpers.js
```diff
@@ -32,10 +32,23 @@ const generateDispatcher = () => new Promise((resolve) => {
   resolve()
 })
 
+const generateIconDispatcher = () => new Promise((resolve) => {
```
loick: Sorry about that one. Apparently the getter index for the icon was not done (ever?). So now the icon process is complete, `npm run kirk icon` => create all of the getter and your file. Only copy paste your SVG content and you're good

PR #32, opened 2018-06-27
[BBCSPA-1494] warningModal component

Tuxosaurus:
Normal (mobile)
warningmodal

Large (desktop)
warningmodal_large


src/_utils/branding.ts
```diff
@@ -49,6 +49,10 @@ export const font: Font = {
     size: '30px',
     lineHeight: String(Math.floor((32 / 30) * 100) / 100),
   },
+  xxl: {
```
pith: `xxxxl` coming soon 😉

Tuxosaurus: One Letter Per Screen™

PerrineBoissieres:
I'd do a refactoring with the Modal component, but probably in another PR ^^

src/_utils/branding.ts
Zarthia: xxl is the biggest one

Tip-Sy:
Looks good to me!

PR #36, opened 2018-06-29
Text component

Zarthia:


Tuxosaurus:
Specs, for reference:
typography_design_system_1_8_3

src/text/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+  readonly children: string | number | React.ReactNode,
+  readonly display?: TextDisplayType,
+  readonly div?: boolean,
+}
+
+const Text = ({
+  className,
+  children,
+  display = TextDisplayType.BODY,
+  div = false,
```
loick: can it only be a div? No bold, italic etc.?

src/text/style.ts
```diff
@@ -0,0 +1,59 @@
+import css from 'styled-jsx/css'
+import { color, font, fontWeight } from '_utils/branding'
+
+export default css`
+  .kirk-text {
```
loick: remove this?

src/text/index.tsx
Tuxosaurus: This is handled by "display", which corresponds to the different types of text shown in the design system

src/text/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+export enum TextDisplayType {
+  BODY = 'body',
+  BODYSTRONG = 'bodyStrong',
+  BUTTON = 'button',
+  CAPTION = 'caption',
+  DISPLAY1 = 'display1',
+  DISPLAY2 = 'display2',
+  SUBHEADER = 'subheader',
+  SUBHEADERSTRONG = 'subheaderStrong',
+  TITLE = 'title',
```
loick: Move this in the Title component x2

src/text/style.ts
```diff
@@ -0,0 +1,52 @@
...
+    line-height: ${font.base.lineHeight};
+  }
+
+  .kirk-text-caption {
+    color: ${color.secondaryText};
+    font-size: ${font.s.size};
+    line-height: ${font.s.lineHeight};
+  }
+
+  .kirk-text-display1 {
```
nickoola: If this display class is related to the size it should be named like .kirk-text-xxl and .kirk-text-xl

PR #38, opened 2018-07-05
[BBCSPA-1605] Only active listeners when the modal is opened

loick:


src/modal/index.tsx
```diff
@@ -42,37 +42,58 @@ class Modal extends Component <ModalProps> {
...
+        document.addEventListener('mouseup', this.handleOutsideMouseClick)
+        document.addEventListener('touchstart', this.handleOutsideMouseClick)
+      }
     }
-    if (this.props.closeOnOutsideClick && canUseEventListeners) {
+  }
+
+  removeListeners() {
+    if (!this.props.isOpen && canUseEventListeners) {
+      document.removeEventListener('keydown', this.handleKeydown)
```
wakooka: Should we have an object referencing all the events? So that if we add a listener we don't forget to remove it

loick:
Changelog updated

PR #39, opened 2018-07-05
Bbcspa 1551 add prefs icons

nickoola:


src/icon/musicIcon.tsx
```diff
@@ -0,0 +1,102 @@
...
+            <polyline
+              stroke={color.iconFaded}
+              stroke-linecap="round"
+              stroke-linejoin="round"
+              points="22.5 18 22.5 1.5 7.5 1.5 7.5 20"
+            />
+            <path
+              stroke={color.white}
+              fill={color.iconFaded}
+              fill-rule="nonzero"
```
loick: this is gonna be a React warning

loick: Check every icon on storybook with the console opened to see the warnings

nickoola: Good point! Done!

PR #43, opened 2018-07-13
Add notification component

pith:
image


src/notification/style.ts
```diff
@@ -0,0 +1,40 @@
+import css from 'styled-jsx/css'
+import { color, space, font, fontWeight } from '_utils/branding'
+
+export default css`
+  .notification {
```
Zarthia: Classes must be prefixed with `.kirk-`

src/notification/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+    if (!canUseDOM || !this.portalNode) {
+      return null
+    }
+
+    const modalElement = (
+      <TransitionGroup component="div" className="transition-wrapper notification-container">
+        {this.props.isOpen && (
+          <CustomTransition animationName={AnimationType.SLIDE_UP}>
+            <div className={cc(['notification', this.props.className])}>
+              <p className="notification-content">{this.props.children}</p>
```
Zarthia: This could be a good first use for the `<Text>` component ;)

src/transitions/style.ts
```diff
@@ -20,13 +20,12 @@ export default css`
   }
 
   :global(.slide-up) {
-    transition: ease-in-out ${transition.duration.base};
-    transition-property: opacity, transform;
+    transition: all ${transition.duration.base} ease-in-out;
```
Zarthia: Why put `all` here instead of specifying the different properties as it was before?

src/notification/story.tsx
```diff
@@ -0,0 +1,43 @@
...
+      isErrorOpen: !isErrorOpen,
+    }))
+  }
+
+  render() {
+    return (
+      <div>
+        <button onClick={this.toggleErrorDisplay}>Don't click on this button</button>
+        <Notification isOpen={this.state.isErrorOpen} close={this.toggleErrorDisplay}>
+          I told you not to click on this button !
```
Zarthia: Can you put a `text()` knob here, so we can test the text in Storybook?

src/notification/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+              <Button
+                status={Button.STATUS.UNSTYLED}
+                className={`notification-cross`}
```
Zarthia: Is that different from `className="notification-cross"`?

src/notification/index.tsx
pith: 👍 I did know that it was merged

src/notification/index.tsx
pith: nope, good catch

src/itemChoice/style.ts
```diff
@@ -2,7 +2,7 @@ import css from 'styled-jsx/css'
 import { color, font, space } from '_utils/branding'
 
 const separatorHeight = '16px'
-const intValue = (value:string) => parseInt(value, 10)
+const intValue = (value: string) => parseInt(value, 10)
```
loick: why do you need this?

src/notification/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  constructor(props: NotificationProps) {
+    super(props)
+    if (canUseDOM) {
+      this.portalNode = document.createElement('div')
+      document.body.appendChild(this.portalNode)
+    }
+  }
+
+  componentWillUnmount() {
```
loick: Server side, this function isn't called, but it might be a good practice to "protect" every document code with `if (canUseDOM) {` ?

src/notification/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+    document.body.removeChild(this.portalNode)
+    this.portalNode = null
+  }
+
+  render() {
+    if (!canUseDOM || !this.portalNode) {
```
loick: !this.portalNode should be enough

src/notification/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+            <div className={cc(['kirk-notification', this.props.className])}>
+              <Text display={TextDisplayType.TITLE} className="kirk-notification-content">
+                {this.props.children}
+              </Text>
+              <Button
+                status={Button.STATUS.UNSTYLED}
+                className="kirk-notification-cross"
+                onClick={this.props.close}
+              >
+                <CrossIcon iconColor={color.white} size="18px" />
```
loick: we don't have this size in the design system? Should we let it that way? @Zarthia

src/notification/index.unit.tsx
```diff
@@ -0,0 +1,22 @@
...
+  it('should not show the notification when isOpen is false', () => {
+    const wrapper = mount(<Notification isOpen={false} close={close} />)
+    expect(wrapper.find('.kirk-notification').exists()).toBe(false)
+  })
+  it('should call close method when click on cross button', () => {
+    const wrapper = mount(<Notification isOpen close={close} />)
+    wrapper.find(Button).simulate('click')
+    expect(close).toHaveBeenCalled()
+  })
+})
```
loick: No snapshot? :)

src/notification/story.tsx
```diff
@@ -0,0 +1,43 @@
...
+  readonly isErrorOpen?: boolean
+}
+
+class ErrorOpener extends Component {
+  state: ErrorOpenerState = {
+    isErrorOpen: false,
+  }
+
+  toggleErrorDisplay = () => {
+    console.log(this.state.isErrorOpen)
```
loick: remove this

loick: However you can bind an action to this (to log it in storybook)

src/text/index.tsx
```diff
@@ -12,6 +12,8 @@ export enum TextDisplayType {
   DISPLAY2 = 'display2',
   SUBHEADER = 'subheader',
   SUBHEADERSTRONG = 'subheaderStrong',
+  TITLE = 'title',
```
loick: we removed it from the previous PR. It was supposed to be part of the title. If you need this as a text, maybe this is the title component you are looking for?

src/transitions/style.ts
```diff
@@ -26,7 +26,7 @@ export default css`
 
   :global(.slide-up-entering) {
     opacity: 0;
-    transform: translateY(50%);
+    transform: translateY(100%);
```
loick: are you sure this is not a breaking change?

src/text/index.tsx
Zarthia: Actually, I told him to put it back. This `title` display is not used for titles per se but for Item titles. So we need it here

src/notification/index.tsx
Zarthia: We don't really have sizes for icons in the design system, only the default one

src/itemChoice/style.ts
pith: Good question, I don't this just prettier reformatting

src/itemChoice/style.ts
```diff
@@ -42,7 +42,7 @@ export default css`
   }
 
   :global(.kirk-itemChoice + .kirk-itemChoice::before) {
-    content: '';
+    content : '';
```
Zarthia: No space needed between the property and the colon

src/text/index.tsx
loick: Then I maintain what I said from the beginning, we should rename it. If we don't only use it for titles......

src/notification/index.tsx
loick: done

src/text/index.tsx
Zarthia: Well, technically, those are titles, item titles. IMO, the one that needs renaming is `display1`, the one used for the titles/theVoice in Kairos

src/itemChoice/style.ts
pith: I didn't touch the file, check the final diff

src/text/index.tsx
pith: Guys, a code review is not the good place to discuss the naming of the design system. 

My opinion on this is: 
* we should stick to the Pixar naming, 
* Then challenge the naming in Pixar.

But we shouldn't be unsync

src/notification/index.unit.tsx
pith: No

PR #46, opened 2018-07-19
fixed issue on timepicker when being in a negative timezone

wakooka:


src/timePicker/index.tsx
```diff
@@ -20,52 +20,76 @@ const toLocaleTimeStringSupportsLocales = (() => {
...
-    const steps:Steps = {}
+  generateTimeSteps = ({ minuteStep = 30, locale }: { minuteStep?: number; locale: string }) => {
+    const steps: Steps = {}
+    // Taking unix time as reference to loop through hours
     const dt = new Date(0)
-    dt.setHours(0)
+    dt.setUTCHours(0)
+    dt.setUTCMinutes(0)
+
     const { renderTime = formatTimeLabel } = this.props
```
pith: `formatTimeLabel` seems unused

wakooka: Used only if `renderTime` is `undefined`

wakooka: I renamed `formatTimeLabel`  to `defaultRenderTime`

PR #48, opened 2018-07-23
[BBCSPA-1490] IE 11 fixes

Tuxosaurus:
Fixes for UneditableTextField and PushInfo
Loader is still not working but it will require a complete change (no SVG transformation) and not a fix.

PushInfo BEFORE
bbcspa1490-pushinfo-before
PushInfo AFTER
bbcspa1490-pushinfo-after

UneditableTextField BEFORE
bbcspa1490-utf-text-before
UneditableTextField AFTER
bbcspa1490-utf-text-after

UneditableTextField with icon BEFORE
bbcspa1490-utf-icon-before
UneditableTextField with icon AFTER
bbcspa1490-utf-icon-after

### DIFF
--- a/src/pushInfo/style.ts
+++ b/src/pushInfo/style.ts
@@ -51,8 +51,8 @@ export default css`
     align-items: center;
     justify-content: center;
     border-radius: 50%;
-    min-width: 40px;
-    min-height: 40px;
+    width: 40px;
+    height: 40px;
     margin-right: ${space.l};
     background-color: ${color.white};
   }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # unreleased
 - [Proximity] Adjust margins
+- [PushInfo] Fixed IE11 issues
+- [UneditableTextField] Fixed IE11 issues
 ...
 
 # v0.6.1 (19/07/2018)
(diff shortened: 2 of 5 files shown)


src/pushInfo/style.ts
```diff
@@ -47,13 +47,18 @@ export default css`
...
     min-height: 40px;
     margin-right: ${space.l};
     background-color: ${color.white};
   }
+
+  :global(.kirk-pushInfo-icon .kirk-icon) {
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, 50%);
```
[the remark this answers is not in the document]
pith: Why do you need a transform ?

src/pushInfo/style.ts
```diff
@@ -47,13 +47,18 @@ export default css`
...
     min-height: 40px;
     margin-right: ${space.l};
     background-color: ${color.white};
   }
+
+  :global(.kirk-pushInfo-icon .kirk-icon) {
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
```
pith: I don't understand the change. Why do you need a transform ?

Bidjii: It's only a partial support, many bugs can be encountered, like this one

Tuxosaurus: Yes, although actually this one is due to the use of "min-height", which I can just turn to "height". Since there's only one case of PushInfo and it's always the same icon for now.

Tuxosaurus: Alright I went for the width/height instead of min-width/min-height. Less changes win.

PR #49, opened 2018-07-23
Fix mismatch on SSR for modals and notifications

PerrineBoissieres:
This fixes all errors relative to server-side rendering of modals and notifications.

### DIFF
--- a/src/modal/index.tsx
+++ b/src/modal/index.tsx
@@ -71,8 +71,10 @@ class Modal extends Component<ModalProps> {
 
   componentWillUnmount() {
     this.removeListeners()
-    document.body.removeChild(this.portalNode)
-    this.portalNode = null
+    if (this.portalNode) {
+      document.body.removeChild(this.portalNode)
+      this.portalNode = null
+    }
     this.setDocumentScroll('visible')
   }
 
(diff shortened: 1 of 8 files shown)


src/modal/index.tsx
```diff
@@ -158,6 +158,10 @@ class Modal extends Component<ModalProps> {
       </div>
     )
 
+    if (!canUseDOM || !this.portalNode) {
+      return modalElement
```
pith: I don't think this will work the modal element won't be at the same place as it is with the portal.
Did you check your change in Kairos ?

PerrineBoissieres: I tested everything in Kairos, this works ;)

PerrineBoissieres: From what I understand of the way `portal` works, is that even if the element is in another `<div>`, React still considers it at its initial position in the tree, which explains the rendering issue in the first place.

Bidjii:
Update the Changelog please :)

src/modal/index.tsx
loick: Add some test by moking canUseDom to be sure you would render it properly server side. And avoid future regressions.

PerrineBoissieres: I can't mock it, I tried several times, but it always stays to true...

src/modal/index.tsx
```diff
@@ -158,6 +158,10 @@ class Modal extends Component<ModalProps> {
       </div>
     )
 
+    if (!canUseDOM || !this.portalNode) {
```
pith: `!this.portalNode` should be enough

PR #52, opened 2018-07-24
Set up inner wrapper in topBar

Bidjii:


src/topBar/index.tsx
```diff
@@ -22,38 +23,51 @@ const TopBar = ({
...
+  const Wrapper = Boolean(innerWrapperClassName) ? 'div' : Fragment
+  const wrapperProps: { className?: string } = {}
+  if (Boolean(innerWrapperClassName)) {
+    wrapperProps.className = cc([innerWrapperClassName])
+  }
+  return (
+    <header
+      className={cc([
+        {
+          topBar: true,
```
Tuxosaurus: why not `'kirk-topBar': true,`?

Bidjii: Because I forgot to add it back when I removed the `prefix` helper

PR #53, opened 2018-07-25
Set overflow property on document to avoid scrolling

nickoola:


nickoola:
Tested on Kairos !

src/modal/index.tsx
```diff
@@ -53,23 +53,33 @@ class Modal extends Component<ModalProps> {
...
 
   componentDidUpdate(prevProps: ModalProps) {
     if (!prevProps.isOpen && this.props.isOpen) {
       this.addListeners()
+      this.setDocumentScroll('hidden')
     }
 
     if (!this.props.isOpen && prevProps.isOpen) {
       this.removeListeners()
+      this.setDocumentScroll('')
```
Tuxosaurus: Shouldn't you set the overflow to "visible"?

nickoola: This way it remove the property

nickoola:
Tested on Safari and it works !

PR #60, opened 2018-08-09
Renaming container prop on Text component

Bidjii:


Bidjii:
Changelog updated

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 - **[FIX]** Handling of the colors on the `<PinIcon>`
+- **[UPDATE]** Change name of the `container` prop on the `<Text>` component into `tag` (harmonization)
```
pith: "breaking change" would be more appropriate

loick: Agreed

src/text/index.tsx
```diff
@@ -16,7 +16,7 @@ export enum TextDisplayType {
   TITLESTRONG = 'titleStrong',
 }
 
-export enum TextContainerType {
+export enum TextTagType {
```
loick: TagType might be enough

Bidjii: I'll keep it that way to be consistant with the `TextDisplayType` above

PR #65, opened 2018-08-13
ItemInfo component

Bidjii:


src/itemInfo/story.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { boolean, text, withKnobs } from '@storybook/addon-knobs'
+
+import ItemInfo from 'itemInfo'
+import PetIcon from 'icon/petIcon'
+
+const stories = storiesOf('ItemInfo', module)
+stories.addDecorator(withKnobs)
+
+stories.add(
+  'Test',
```
pith: change the name of the story

src/itemInfo/index.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+
+import Item, { ItemProps } from '_utils/item/index'
+
+interface ItemInfoProps extends ItemProps {
```
pith: It should not extends itemProps as it should be a subset of Item

src/itemInfo/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const ItemInfo = ({
+  mainInfo,
+  className,
+  mainTitle,
+  icon,
+  tag,
+}:ItemInfoProps) => {
+  return (
+    <Item
```
loick: Why is Item not... ItemInfo? 

No CSS, no specfic treatment in this one.. Just wondering

Bidjii: There was no specific CSS treatment for `ItemData` neither. We use `Item` as a common skeleton for the 4 types of Items, we're not supposed to have specific CSS here, everything is handled `Item`-side.
But I don't want to export `Item` to use it as is (don't want to fall into V3's problems again), so I will create those 4 Items (`ItemData`, `ItemInfo`, `ItemAction` and `ItemChoice`), restricting the `Item` in certain configuration to match specs

src/itemInfo/story.tsx
pith: up

src/itemInfo/index.tsx
loick: Okay, fine by me. 

Maybe it would be great to add a readme about this?

PerrineBoissieres: Could be a great idea to have it in the documentation somewhere, mainly to remember why we did this in the future and not going through the same mistake again of having a component way too complex to use :)
This could be in the wiki somewhere.

PR #68, opened 2018-08-21
fix `flush` import from `styled-jsx`

chabou:
`flush` and `flushToHTML` exports are not the same.
See usage example: https://github.com/zeit/styled-jsx#styled-jsxserver

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -35,8 +35,7 @@ import WarningModal from 'warningModal'
 import Why from 'why'
 
 import branding from '_utils/branding'
-import { flushToHTML } from 'styled-jsx/server'
-import flush from 'styled-jsx/server'
+import { flush, flushToHTML } from 'styled-jsx/server'
 
 export * from 'icon/index'
 
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -35,7 +35,8 @@ import WarningModal from 'warningModal'
 import Why from 'why'
 
 import branding from '_utils/branding'
-import { flush, flushToHTML } from 'styled-jsx/server'
+import { flushToHTML } from 'styled-jsx/server'
+import flush from 'styled-jsx/server'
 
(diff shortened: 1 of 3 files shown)


src/index.tsx
```diff
@@ -35,7 +35,8 @@ import WarningModal from 'warningModal'
 import Why from 'why'
 
 import branding from '_utils/branding'
-import { flush, flushToHTML } from 'styled-jsx/server'
+import { flushToHTML } from 'styled-jsx/server'
+import flush from 'styled-jsx/server'
```
wakooka: This file is generated, we just added a comment at the top of the file saying this a couple of days ago. The file to update is here https://github.com/blablacar/ui-library/blob/master/bin/templates/components.tsx#L7
Can you also do the import on one line?
`import flush, { flushToHTML } from 'styled-jsx/server'`
Thanks for spotting this 👍

chabou: Ok, I'll change the template file, but is it attended to have same header in this file?
```
/**
 * THIS FILE IS GENERATED
 */
```
Maybe `THIS FILE IS USED TO GENERATE src/index.tsx`. 

Other thing, maybe it could be better to not have generated files in src folder in repository? It can only lead to mistake and unsynchronized files (between updated template and generated file).

How can I generate src/index.jsx from template to commit both?

wakooka: I agree this is not great. After merging this PR, we'll remove the generated file from versioning.
About the template, I also agree the comment `THIS FILE IS GENERATED` is misleading, it's because we use the raw template file to build the generated file, we'll probably remove it from the template file, and will add to the generated file only.

And to generate the `src/index.jsx` run `npm run kirk init`
Sorry about that, we'll work on a better system right after merging your PR
Thanks again :)

wakooka:
Thanks again 👍

PR #71, opened 2018-08-23
Fix dropdownbutton type

chabou:
Default type is "submit" for most of browsers. This leads to an issue if DropDownButton is in a form.

capture d ecran 2018-08-23 a 15 24 30

My current workaround:
```
  onDropdownRef(ref) {
    const element = ReactDOM.findDOMNode(ref)
    element && element.children[0].setAttribute('type', 'button')
  }
```


src/dropdownButton/index.tsx
```diff
@@ -82,7 +82,7 @@ class DropdownButton extends PureComponent <DropdownButtonProps, DropdownButtonS
 
     return (
       <div ref={this.refButton} className="dropdownButton">
-        <button className={cc(className)} onClick={this.toggle}>
+        <button type="button" className={cc(className)} onClick={this.toggle}>
```
wakooka: Can you also add a unit test in `/src/dropdownButton/index.unit.tsx`?
Something like this:
 ```
it('Should have an attribute type button', () => {
    const wrapper = shallow(<DropdownButton {...defaultProps} />)
    expect(wrapper.find('button[type="button"]').toHaveLength(1)
  })

```

chabou: Done

wakooka:
Thanks for the fix @chabou

PR #73, opened 2018-08-27
changed <Menu> to use <ItemChoice>

wakooka:
So far there's a separator on each item, an update on the ItemChoice is coming to be able to specify which has a separator.
capture d ecran 2018-08-27 a 11 11 03


src/itemChoice/index.tsx
```diff
@@ -107,7 +107,7 @@ class ItemChoice extends PureComponent <ItemChoiceProps> {
     return (
       <Component
         role="option"
-        key={key}
+        key={id}
```
pith: 🤔 why do we have a key here ? IMO you should just remove it

src/menu/index.tsx
```diff
@@ -1,18 +1,23 @@
...
 }
 
 const Menu = ({ className, items }: MenuProps) => (
-  <ul className={cc(className)} role="menu">
-    {items.map(item => <MenuItem {...item} key={item.id} />)}
-    <style jsx>{style}</style>
-  </ul>
+  <nav className={cc(className)} role="menu">
+    {items.map(item => (
+      <ItemChoice
```
pith: Does it use an li tag by default ? If it's the case you might need to keep the ul tag

src/menu/story.tsx
```diff
@@ -12,22 +14,57 @@ stories.addDecorator(withKnobs)
 const menuItems = [
   {
     id: 'menu-item-1',
-    label: 'Menu item 1',
-    url: '#',
+    label: 'Dashboard',
+    leftAddon: <HomeIcon />,
+    href: '#',
```
pith: nit: `#item-1`

src/menu/index.tsx
```diff
@@ -1,18 +1,23 @@
...
-import style from './style'
+import ItemChoice, { ItemChoiceProps } from 'itemChoice'
 
 export interface MenuProps {
-  items: MenuItemProps[],
+  items: ItemChoiceProps[],
   className?: Classcat.Class,
 }
 
 const Menu = ({ className, items }: MenuProps) => (
```
pith: It's true that the added value of this component is not huge

src/menu/index.tsx
wakooka: It's either a `<a>` if there's a `href` prop or a `<li>` otherwise

src/itemChoice/index.tsx
wakooka: ping @Bidjii

Bidjii: Yes you could remove it, I talked about it with Loïck, I just have to make sure that Typescript doesn't throw us an error if we don't allow the props to be put here on the first place

src/menu/index.tsx
Bidjii: If there's a `href` props that is a string, you get a `<a>`, yes

src/itemChoice/index.tsx
wakooka: Ok I removed it and checked, it's all good.

PR #74, opened 2018-08-27
Use Facebook and VK brand color as default icon color

pith:
image
image

I also added a story to see all icons:

image


Bidjii:
Wouldn't it be good to do this for all social media icons at once?

pith:
@Bidjii I asked Dawid for the other icons

PR #75, opened 2018-08-27
Text component is now replacing new line by <br/> tags

pith:
image

### DIFF
--- a/src/text/index.tsx
+++ b/src/text/index.tsx
@@ -27,6 +27,7 @@ interface TextProps {
   readonly children: string | number | React.ReactNode
   readonly display?: TextDisplayType
   readonly tag?: TextTagType
+  readonly newlineToBr?: boolean
 }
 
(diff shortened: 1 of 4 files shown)


src/text/index.tsx
```diff
@@ -29,24 +29,39 @@ interface TextProps {
...
   tag = TextTagType.SPAN,
-}:TextProps) => {
+}: TextProps) => {
   const baseClassName = 'kirk-text'
   const displayClassName = `${baseClassName}-${display}`
   const Tag = tag
 
+  let content = children
+  if (typeof children === 'string') {
+    content = replaceNewLineWithBR(children)
```
wakooka: Can you make it a prop `nl2br` (default to true)? Not sure about the naming though :)

nickoola: @wakooka  in what case do you think we'll need to disable `nl2br` ?

PR #82, opened 2018-09-17
Add new Badge component

Tuxosaurus:
Badge icon, for small notifications such as message count.
Used for HomeIcon and BubbleIcon (for now).
Also replaced the ID Check element in Avatar.


src/avatar/index.tsx
```diff
@@ -2,36 +2,37 @@ import React from 'react'
...
+  bottomBadge,
+  topBadge,
+}: AvatarInterface) => (
+    <div
+      className={cc([
+        prefix({ medium, large, image: !!image }, 'kirk-avatar-'),
+        className, 'kirk-avatar',
+      ])}>
+      {image && <img src={image} alt={alt} />}
+      {topBadge}
```
loick: I didn't know we had two ways of putting badges on the avatar. 
Moroever, you have topBadge and bottomBade, but nothing is actually placing them right?

src/avatar/index.unit.tsx
```diff
@@ -1,26 +1,33 @@
 import React from 'react'
-import Avatar, { IdCheck } from 'avatar'
-import Check from 'icon/checkIcon'
+import Avatar from 'avatar'
+import Badge from 'badge'
+import CheckIcon from 'icon/checkIcon'
+
+const bottomBadge = <Badge className="kirk-badge--idCheck">
```
loick: we shouldn't use kirk classes outside of the components

src/avatar/index.tsx
Tuxosaurus: Agreed, nothing places them automatically to the top or bottom.
As you suggested, I won't allow for any JSX.Element there but preset a `UnreadNotificationsBadge` and `IdCheckBadge` instead.

src/icon/homeIcon.tsx
```diff
@@ -36,6 +39,18 @@ class HomeIcon extends PureComponent<Icon> {
           C24.056,12.632,24.041,12.316,23.836,12.13z" />
       </svg>
     )
+
+    if (!isEmpty(badgeContent)) {
```
wakooka: you can do `if(!badgeContent)` since it's a string

src/profile/index.tsx
```diff
@@ -28,32 +31,32 @@ const Profile = ({
...
-        <Rating ratings={ratings} score={score}>{ratingsLabel}</Rating>
+    <div className={cc(['kirk-profile', prefix({ medium }), className])}>
+      <div className="kirk-description">
+        <span className="kirk-title">{title}</span>
+        {ratings > 0 &&
+          <Rating ratings={ratings} score={score}>{ratingsLabel}</Rating>
+        }
+        <span className="kirk-secondaryInfo">{info}</span>
+      </div>
+      {!isEmpty(picture) &&
```
wakooka: picture is a string `{ picture &&`

src/profile/index.tsx
```diff
@@ -28,32 +31,32 @@ const Profile = ({
...
+        <Avatar
+          image={picture}
+          alt={alt}
+          isIdChecked={checked}
+          className="kirk-picture"
+          medium={medium}
+        />
+      }
+      {!isEmpty(action) &&
```
wakooka: action is string `{ action &&`

src/avatar/index.tsx
```diff
@@ -1,37 +1,54 @@
...
+  isIdChecked,
+  unreadNotificationsCount,
+}: AvatarInterface) => (
+    <div
+      className={cc([
+        prefix({ medium, large, image: !!image }, 'kirk-avatar-'),
+        className, 'kirk-avatar',
+      ])}>
+      {image && <img src={image} alt={alt} />}
+      {!isEmpty(unreadNotificationsCount) && unreadNotificationsBadge(unreadNotificationsCount)}
```
wakooka: `unreadNotificationsCount` is a string { unreadNotificationsCount &&

src/badge/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+  if (isEmpty(children)) {
+    return null
+  }
+
+  return (
+    <div
+      className={cc(['kirk-badge', className])}
+      aria-label={ariaLabel}
+    >
+      <span aria-hidden={!isEmpty(ariaLabel)}>{children}</span>
```
wakooka: `ariaLabel` is a string `{ !!ariaLabel }`

src/badge/index.unit.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+
+import Badge from 'badge'
+
+it('Should read the aria-label atribute and not the children if both are provided', () => {
```
Bidjii: attribute :)

src/avatar/index.tsx
```diff
@@ -1,37 +1,54 @@
...
-const Avatar = ({ medium, large, className, checked, image, alt = '', title }: AvatarInterface) => (
-  <div className={cc([prefix({ medium, large, image: !!image }), className])}>
-    {image && <img src={image} alt={alt} />}
-    {checked && <IdCheck checked={checked} title={title} />}
-    <style jsx>{style}</style>
-  </div>
+const unreadNotificationsBadge = (unreadNotificationsCount: string) => (
+  <Badge
+    className="kirk-avatar-badge--unreadNotifications"
+    ariaLabel="Unread messages">
```
loick: should be a prop to be able to translate it

src/avatar/index.tsx
```diff
@@ -1,37 +1,54 @@
...
-}
-
-export interface AvatarInterface extends IdCheck {
+export interface AvatarInterface {
   readonly className?: Classcat.Class
   readonly image?: string
   readonly alt?: string
   readonly medium?: boolean
   readonly large?: boolean
+  readonly isIdChecked?: boolean
```
loick: `idChecked` instead?

src/avatar/index.tsx
```diff
@@ -1,37 +1,54 @@
...
+  large,
+  className,
+  image,
+  alt = '',
+  isIdChecked,
+  unreadNotificationsCount,
+}: AvatarInterface) => (
+    <div
+      className={cc([
+        prefix({ medium, large, image: !!image }, 'kirk-avatar-'),
```
loick: why using `image: !!image`? And nothing on the other props?

src/avatar/index.tsx
```diff
@@ -1,37 +1,54 @@
...
+    ariaLabel="Unread messages">
+    {unreadNotificationsCount}
+  </Badge>
 )
 
+const Avatar = ({
+  medium,
+  large,
+  className,
+  image,
```
loick: picture?

src/badge/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+import style from 'badge/style'
+
+interface BadgeProps {
+  readonly className?: Classcat.Class,
+  readonly children: string | JSX.Element,
+  readonly ariaLabel?: string,
+}
+
+const Badge = ({ className, children, ariaLabel }: BadgeProps) => {
+  if (isEmpty(children) || children === '0') {
```
loick: Can't it be a number?

src/avatar/index.tsx
Tuxosaurus: why?

src/badge/index.tsx
Tuxosaurus: Strictly speaking, yes, but I'd rather have a string if one wants to use it for stuff like "NEW".

src/avatar/index.tsx
Tuxosaurus: why?

src/avatar/index.tsx
Tuxosaurus: indeed

src/avatar/index.tsx
Tuxosaurus: I haven't changed that part. I guess it's to have everything as booleans, since medium and large are booleans but image is a string.

src/avatar/index.tsx
Tuxosaurus: indeed

src/avatar/index.tsx
loick: It's not that easy to read :)

src/avatar/index.tsx
loick: For an avatar, I would say this is more a profile picture

src/badge/index.tsx
loick: okay, but in the tests you are actually giving a number so it should be managed by Typescript as well I guess?

src/avatar/index.tsx
Tuxosaurus: Well it's a sort of convention that using 'is' or 'has' clarifies that we're dealing with a boolean.
If we don't go by this rule, ok, but it does make things easier to read/understand.

src/badge/index.tsx
Tuxosaurus: Looks like Typescript is set to not agree with `number` for `aria-label`
screen shot 2018-09-26 at 15 06 12

src/avatar/index.tsx
Tuxosaurus: I can agree but, but all in all, we're using an image `` tag, and not (yet?) the `<picture>` tag

src/avatar/index.tsx
Tuxosaurus: It won't use shortcuts anymore since I'll be changing the props, so we'll have `prefix({ medium: isMedium, large: isLarge, image: !!image }, 'kirk-avatar-'),`

src/avatar/index.tsx
loick: Leave it like that then :)

src/badge/index.tsx
loick: I was talking about the children for the number

Tuxosaurus: Hum, I'm not sure I'm actually giving it a number in the test, `12` here should be rendered as a string `<Badge>12</Badge>`

Tuxosaurus: Ok as discussed with @wakooka I'm simply gonna remove the `children === "0"` since this logic should not be handled in `ui-library`

PR #86, opened 2018-09-20
Update Stepper and Button to change value by keeping Buttons pressed

nickoola:
Manage mouseDown and mouseUp events on Stepper

### DIFF
--- a/src/stepper/index.tsx
+++ b/src/stepper/index.tsx
@@ -4,7 +4,7 @@ import Button from 'button'
 import PlusIcon from 'icon/plusIcon'
 import MinusIcon from 'icon/minusIcon'
 
-import { color } from '_utils/branding'
+import { color, delay } from '_utils/branding'
 import style from 'stepper/style'
 
 interface StepperProps {
@@ -17,11 +17,12 @@ interface StepperProps {
   max?: number,
   min?: number,
   format?: (value: string | number) => string | number,
-  onChange?: (obj:onChangeParameters) => void,
 }
 
 interface StepperState {
   value: number,
+  min: number,
+  max: number,
 }
 
 // Support IE. Same value returned with Number.MAX_SAFE_INTEGER / Number.MIN_SAFE_INTEGER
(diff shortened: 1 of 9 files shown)


src/stepper/index.tsx
```diff
@@ -24,6 +24,14 @@ interface StepperState {
   value: number,
 }
 
+const KEYCODES = {
+  ARROW_LEFT: 37,
+  ARROW_RIGHT: 39,
+}
+
+let whileMouseDown:any
+let mouseDownDelay:any
```
loick: this should be a class value

src/stepper/index.tsx
```diff
@@ -43,6 +51,7 @@ export default class Stepper extends PureComponent <StepperProps, StepperState>
 
   componentDidMount() {
     this.update(this.state.value, undefined, false)
+    document.addEventListener('keydown', this.handleKeyDown)
   }
 
   componentWillReceiveProps(nextProps: StepperProps) {
```
loick: btw, can you migrate this into a componentDidUpdate ? :)

src/stepper/index.tsx
```diff
@@ -43,6 +51,7 @@ export default class Stepper extends PureComponent <StepperProps, StepperState>
 
   componentDidMount() {
     this.update(this.state.value, undefined, false)
+    document.addEventListener('keydown', this.handleKeyDown)
```
loick: Can you use a native react event for this?

src/stepper/index.tsx
nickoola: Done

src/stepper/index.tsx
nickoola: There are no other way to manage those events, I plan to remove it cause in case of multiple Stepper it will be a mess.

src/stepper/index.tsx
nickoola: Done

src/stepper/index.tsx
nickoola: Done

src/stepper/index.unit.tsx
```diff
@@ -47,18 +47,18 @@ it('Should be able to increment/decrement the value', () => {
...
       Amount of something
     </Stepper>,
   )
   expect(stepper.state('value')).toBe(3)
-  stepper.find('.kirk-stepper-increment').simulate('click')
+  stepper.find('.kirk-stepper-increment').simulate('mouseUp')
   expect(stepper.state('value')).toBe(6)
-  expect(onChange).toHaveBeenCalledWith({ name: 'testName', value: 6 })
-  stepper.find('.kirk-stepper-decrement').simulate('click')
+  // expect(onChange).toHaveBeenCalledWith({ name: 'testName', value: 6 })
```
Bidjii: Do we keep this?

src/stepper/index.unit.tsx
```diff
@@ -47,18 +47,18 @@ it('Should be able to increment/decrement the value', () => {
...
+  stepper.find('.kirk-stepper-decrement').simulate('mouseUp')
   expect(stepper.state('value')).toBe(3)
-  expect(onChange).toHaveBeenCalledWith({ name: 'testName', value: 3 })
+  // expect(onChange).toHaveBeenCalledWith({ name: 'testName', value: 3 })
```
Bidjii: Do we keep this?

src/stepper/index.tsx
```diff
@@ -34,33 +33,51 @@ export default class Stepper extends PureComponent <StepperProps, StepperState>
...
   state:StepperState = {
     value: this.props.value,
   }
 
+  whileMouseDown:number
+  mouseDownDelay:number
+
   componentDidMount() {
-    this.update(this.state.value, undefined, false)
+    this.update(this.state.value)
```
loick: No need for a componentDidMount, you can initialize the state directly with your Min/Max L.39. That way, it would also be intialized at the right value server side (ComponentDidMount is not called server side)

src/stepper/index.unit.tsx
```diff
@@ -47,18 +47,18 @@ it('Should be able to increment/decrement the value', () => {
       step={3}
       min={2}
       max={10}
-      onChange={onChange}
+      // onChange={onChange}
```
loick: Remove comment

src/stepper/index.unit.tsx
nickoola: Done

PR #92, opened 2018-09-26
Autocomplete triggers a search on mount when a defaultValue is provided

pith:
The autocomplete component should be refactored but it requires too much work to be part of this PR.
It will be done in: BBCSPA-2041

### DIFF
--- a/src/autoComplete/index.tsx
+++ b/src/autoComplete/index.tsx
@@ -106,8 +106,6 @@ export default class AutoComplete extends Component<AutoCompleteProps, AutoCompl
     super(props)
     if (props.debounceTimeout > 0) {
       this.searchForItems = debounce(this.searchForItems, props.debounceTimeout)
-    } else {
-      this.searchForItems = this.searchForItems.bind(this)
     }
     this.currentValue = this.props.defaultValue
     this.state = {
@@ -192,7 +190,7 @@ export default class AutoComplete extends Component<AutoCompleteProps, AutoCompl
     return String(this.currentValue).length >= this.props.searchForItemsMinChars
   }
 
-  searchForItems() {
+  searchForItems = () => {
     // If a long `debounceTimeout` is setup, it may happen that a `searchForItems`
     // is still scheduled to be called while the user has modified the query
     // during that lapse of time. Therefore, the check below verify the real input value
--- a/src/autoComplete/index.tsx
+++ b/src/autoComplete/index.tsx
@@ -13,67 +13,69 @@ import style from './style'
 
 type query = string | number | boolean
 interface AutoCompleteProps {
-  readonly name: string,
-  readonly isSearching: boolean,
-  readonly searchForItems: (query: query) => void,
-  readonly onInputChange?: (params: Partial<onChangeParameters>) => void,
-  readonly searchForItemsMinChars?: number,
-  readonly defaultValue?: string,
-  readonly onSelect?: (obj:AutocompleteOnChange) => void,
-  readonly onFocus?: (event: React.FocusEvent<HTMLInputElement>) => void,
-  readonly onBlur?: (event: React.FocusEvent<HTMLInputElement>) => void,
-  readonly onClear?: () => void,
-  readonly className?: Classcat.Class,
-  readonly inputClassName?: Classcat.Class,
-  readonly itemClassName?: Classcat.Class,
-  readonly bodyClassName?: Classcat.Class,
-  readonly items?: AutocompleteItem[],
-  readonly maxItems?: number,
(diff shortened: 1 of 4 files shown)


wakooka:
@pith I fixed the linter errror. But I didn't investigate much on the warning yet.

src/autoComplete/index.tsx
```diff
@@ -107,33 +109,44 @@ export default class AutoComplete extends Component<AutoCompleteProps, AutoCompl
     } else {
       this.searchForItems = this.searchForItems.bind(this)
```
loick: Wow, can you remove this please? 

No more bind this, use class properties instead

src/autoComplete/index.tsx
```diff
@@ -206,16 +222,19 @@ export default class AutoComplete extends Component<AutoCompleteProps, AutoCompl
...
-      this.state.items
-     ) : this.props.renderEmptySearch
+    const shouldDisplayNoResults =
+      this.hasMinCharsForSearch() && !this.state.busy && this.state.noResults && this.props.showList
+    const shouldDisplayAutoCompleteList =
+      this.hasMinCharsForSearch() &&
+      !isEmpty(this.state.items) &&
+      !this.state.busy &&
+      this.props.showList
+    const listItems = shouldDisplayAutoCompleteList
```
loick: Maybe we should create a get method for this instead of having it in the render

src/autoComplete/index.unit.tsx
```diff
@@ -47,34 +46,33 @@ describe('AutoComplete', () => {
...
     wrapper.instance().onInputChange({ value: 'title' })
 
     const items = fakeSearchForItems()
     wrapper.setProps({ isSearching: true })
     wrapper.setProps({ items, isSearching: false })
-    wrapper.find('li').first().simulate('mousedown')
+    wrapper
+      .find('li')
+      .first()
+      .simulate('mousedown')
```
loick: click? We don't have any mousedown events on the autocomplete

src/autoComplete/index.unit.tsx
```diff
@@ -47,34 +46,33 @@ describe('AutoComplete', () => {
...
+      <AutoComplete
+        {...defaultProps}
+        onDoneAnimationEnd={event}
+        selectedItemStatus={ItemChoiceStatus.CHECKED}
+      />,
+    )
     wrapper.instance().onInputChange({ value: 'title' })
 
     const items = fakeSearchForItems()
     wrapper.setProps({ isSearching: true })
```
loick: Why do you simulate the loading state in the proprs change since it's not the feature you are testing

src/autoComplete/index.unit.tsx
```diff
@@ -47,34 +46,33 @@ describe('AutoComplete', () => {
...
     wrapper.setProps({ items, isSearching: false })
-    wrapper.find('li').first().simulate('mousedown')
+    wrapper
+      .find('li')
+      .first()
+      .simulate('mousedown')
 
     expect(event).not.toBeCalled()
```
loick: I don't understand this test

src/autoComplete/story.tsx
```diff
@@ -37,7 +37,8 @@ class AutoCompleteExample extends Component<AutoCompleteExampleProps, AutoComple
     renderEmptySearch: [],
   }
 
-  searchForItems = (query:string) => {
+  searchForItems = (query: string) => {
+    console.log(query)
```
loick: woops

src/autoComplete/index.unit.tsx
pith: I didn't touch this code...

src/autoComplete/index.tsx
pith: done

src/autoComplete/index.tsx
pith: I created a refactoring ticket for that: BBCSPA-2041

src/autoComplete/index.unit.tsx
pith: Click doesn't work => BBCSPA-2041

src/autoComplete/index.tsx
loick: Deal

src/autoComplete/index.unit.tsx
loick: Yep sorry, just saw that. I put a global comment regarding these tests afterwards

chabou:
Please add a property to disable this feature when you'll refactor this Autocomplete.
To prevent this, I have to add a localstate to force a false `showList` property before a real search. And there is some real use cases where user want to have a default value without handle a search.

Thank you 🙏

PR #97, opened 2018-09-28
[Breaking Change] - DropdownButton no longer handles the dropdown content

wakooka:
This component becomes a controlled component, as the state has been removed. So the behaviour
 is consistent with the HamburgerButton.
animation-dropdown

### DIFF
--- a/src/dropdownButton/index.tsx
+++ b/src/dropdownButton/index.tsx
@@ -16,7 +16,7 @@ const DropdownButton = ({
   open = false,
   children,
   onClick,
-  className,
+  className = '',
   iconPosition = 'right',
 }: DropdownMenu) => (
   <div
@@ -27,7 +27,7 @@ const DropdownButton = ({
       },
     ])}>
     <button
-      aria-expanded={open ? 'true' : 'false'}
+      aria-expanded={open}
       type="button"
       className={className}
       onClick={onClick}>
--- a/src/dropdownButton/index.tsx
+++ b/src/dropdownButton/index.tsx
@@ -4,11 +4,11 @@ import style from './style'
 import { ChevronIcon } from 'icon'
 import { color } from '_utils/branding'
 
-export interface DropdownMenu {
+export interface DropdownButtonProps {
   onClick: (event: React.MouseEvent<HTMLElement>) => void,
   children: JSX.Element | string
-  open?: boolean,
-  className?: string
+  open?: boolean
+  className?: ''
   iconPosition?: 'left' | 'right'
 }
 
@@ -18,7 +18,7 @@ const DropdownButton = ({
   onClick,
   className = '',
   iconPosition = 'right',
-}: DropdownMenu) => (
+}: DropdownButtonProps) => (
   <div
     className={cc([
       'kirk-dropdownButton',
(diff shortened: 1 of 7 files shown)


src/dropdownButton/index.tsx
```diff
@@ -1,111 +1,46 @@
...
-  className?: Classcat.Class,
-}
 
-export interface DropdownButtonState {
-  open: boolean,
+export interface DropdownMenu {
+  onClick: (event: React.MouseEvent<HTMLElement>) => void,
+  children: JSX.Element | string
+  open?: boolean,
+  className?: string
```
loick: optional value but no default value set?

src/dropdownButton/index.tsx
```diff
@@ -1,111 +1,46 @@
...
+}: DropdownMenu) => (
+  <div
+    className={cc([
+      'kirk-dropdownButton',
+      {
+        'kirk-dropdownButton--open': open,
+      },
+    ])}>
+    <button
+      aria-expanded={open ? 'true' : 'false'}
```
loick: Boolean is not working?

src/dropdownButton/index.tsx
```diff
@@ -1,111 +1,46 @@
...
+  iconPosition = 'right',
+}: DropdownMenu) => (
+  <div
+    className={cc([
+      'kirk-dropdownButton',
+      {
+        'kirk-dropdownButton--open': open,
+      },
+    ])}>
+    <button
```
loick: Why not using the Button element with the unstyled value to avoid repeating the CSS?

src/dropdownButton/index.tsx
wakooka: added default value

src/dropdownButton/index.tsx
wakooka: actually I copied this from the HamburgerButton, I made the change and it works just by passing `open`

src/dropdownButton/index.tsx
wakooka: I tried it's causing layout issue.

src/dropdownButton/index.tsx
Bidjii: Why not use `Classcat.Class` type?

src/dropdownButton/index.tsx
```diff
@@ -1,111 +1,46 @@
...
-    )
-  }
-}
+const DropdownButton = ({
+  open = false,
+  children,
+  onClick,
+  className = '',
+  iconPosition = 'right',
+}: DropdownMenu) => (
```
Bidjii: `DropdownMenu` > `DropdownButton`

src/dropdownButton/index.tsx
wakooka: As discussed IRL, it can be used on consumer side because we don't use `cc` in this component

src/dropdownButton/index.tsx
wakooka: Done

src/dropdownButton/index.tsx
loick: Still not convinced, what is the problem about that?

PR #98, opened 2018-10-01
Add touch event on Stepper and Button components

nickoola:


src/stepper/index.tsx
```diff
@@ -29,6 +30,15 @@ interface StepperState {
 // Support IE. Same value returned with Number.MAX_SAFE_INTEGER / Number.MIN_SAFE_INTEGER
 const defaultInteger = (2 ** 53) - 1
 
+const isTouchEventAvailable = () => {
```
loick: Move it in an util

src/stepper/index.tsx
```diff
@@ -68,7 +78,10 @@ export default class Stepper extends PureComponent <StepperProps, StepperState>
     this.props.onChange({ name: this.props.name, value })
   }
 
-  handleMouseDown = (fn: () => void) => () => {
+  handleButtonDown = (fn: () => void, isTouchEvent:Boolean = false) => () => {
```
loick: no need to precise `:Boolean` since you put an explicit default value

loick: And no need to pass this variable as an argument since it's gonna be available everywhere

src/stepper/index.tsx
nickoola: Done

src/stepper/index.tsx
nickoola: Ok for `:Boolean` but I'll keep this argument cause I'm using it depending on callback event. In case of touch screen some mouse event are triggered, so this argument allow me to cancel mouse event on mobile for example.

nickoola: We finally use touchEvent check an other way to apply appropriate Button events.

src/_utils/index.ts
```diff
@@ -6,4 +8,11 @@ export const prefix = (
   return [].concat(mods.map(modifier => `${baseClass}-${modifier}`))
 }
 
+export const isTouchEventAvailable = () => {
+  if (canUseDOM) {
+    return 'ontouchstart' in window
+  }
+  return false
+}
```
wakooka: nit: `isTouchEventAvailable = () => canUseDom && 'ontouchstart' in window`

src/stepper/index.tsx
```diff
@@ -97,6 +101,16 @@ export default class Stepper extends PureComponent <StepperProps, StepperState>
     this.update(this.state.value - this.props.step)
   }
 
+  createListeners(callback: () => void) {
+    return isTouchScreen ? {
+      onTouchStart: this.handleButtonDown(callback),
+      onTouchEnd: this.handleButtonUp(callback),
+    } : {
+      onMouseDown: this.handleButtonDown(callback),
+      onMouseUp:this.handleButtonUp(callback),
```
wakooka: Have you checked on hybrid devices, like a surface? Because `isTouchScreen` will return `true` so I'm thinking it might not even work if you use the mouse.

nickoola: Tested on a windows laptop with tactile screen, mouse + touch events works to increase/decrease value.

src/_utils/index.ts
nickoola: Done

PR #113, opened 2018-10-16
Links in Itinerary

Bidjii:


src/itinerary/index.tsx
```diff
@@ -16,76 +18,116 @@ interface ItineraryProps {
...
-                </Text>
-              )
+          </li>
+        )
+      }
+      {
+        places.map((place, index) => {
+
+          let Component: tag
+          let chevron: boolean = false
```
loick: No need to specify this is a boolean, this is quite explicit here

src/itinerary/index.tsx
```diff
@@ -16,76 +18,116 @@ interface ItineraryProps {
...
+              className: cc(['kirk-itinerary-location-wrapper', className]),
+            }
+          }
+
+          return (
+            <li
+              className={
+                cc(['kirk-itinerary-location', {
+                  'kirk-itinerary--departure': index === 0,
+                  'kirk-itinerary--arrival': index === places.length - 1,
```
loick: create a lastPlace const here

src/itinerary/index.tsx
```diff
@@ -16,76 +18,117 @@ interface ItineraryProps {
...
+      }
+      {
+        places.map((place, index) => {
+
+          let Component: tag
+          let chevron = false
+          let hrefProps
+
+          const link = place.href
+          const lastPlace = places.length - 1
```
loick: isLastPlace = places.length - 1 === index

PR #116, opened 2018-10-17
Add halo on Button focus

zoontek:
Hello!

I realized that your buttons doesn't make the distinction between focused state and hovered / active state. It makes keyboard navigation difficult.

Maybe my implementation is not perfect (for your designers), but it's an invite to think about it. Keyboard navigation is important for accessibility (and not only for pro users!) :)


loick:
Hi @zoontek, 

First of all, thanks for your PR! The topic has been raised and our design team will provide an update about this. 

We will let you know about the updates of this topic!

PR #122, opened 2018-10-19
Fix button regression due to overflow hidden

pith:


src/button/index.tsx
```diff
@@ -46,9 +46,9 @@ export interface ButtonState {
 }
 
 type ButtonActionEvents =
-  React.MouseEvent<HTMLElement> |
-  React.TouchEvent<HTMLElement> |
-  React.FocusEventHandler<HTMLElement>
+  | React.MouseEvent<HTMLElement>
```
Tuxosaurus:
```suggestion
  React.MouseEvent<HTMLElement>
```
For the first line

pith: It's forced by prettier ;)

PR #126, opened 2018-11-12
Checkbox loading state

PerrineBoissieres:
### Specs:
Checkboxes should be able to display API request status when being clicked. It uses the same STATUS prop than Item components and variants for consistency.
Loader replaces the blue check icon, either placed on right or left.

### Screenshots:
capture d ecran 2018-11-12 a 12 01 21
capture d ecran 2018-11-12 a 12 01 09

### DIFF
--- a/src/checkbox/index.unit.tsx
+++ b/src/checkbox/index.unit.tsx
@@ -1,92 +1,143 @@
 import React from 'react'
 import renderer from 'react-test-renderer'
 import Checkbox, { labelDisplays } from 'checkbox'
+import Loader from 'loader'
 
-it('should have the proper text & attributes', () => {
-  const checkbox = shallow((
-    <Checkbox
-      name="checkbox1"
-      value="value"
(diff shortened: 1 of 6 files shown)


src/checkbox/index.unit.tsx
```diff
@@ -1,92 +1,143 @@
...
-  expect(checkbox.state('isChecked')).toBe(false)
-  // Value changing
-  checkbox.setProps({ checked: true })
-  expect(checkbox.state('isChecked')).toBe(true)
-  checkbox.setProps({ checked: false })
-  expect(checkbox.state('isChecked')).toBe(false)
-})
+  it('should have the accessibility attributes', () => {
+    const checkbox = shallow(<Checkbox name="checkbox1" value="value">Label</Checkbox>)
+    expect(checkbox.find('label input')).toHaveLength(1)
```
nickoola: Why not cibling the label directly ?

PR #141, opened 2018-11-30
Add color to Text

Bidjii:


src/text/index.tsx
```diff
@@ -47,19 +48,25 @@ const Text = ({
...
   let content = children
   if (typeof children === 'string' && newlineToBr) {
     content = replaceNewLineWithBR(children)
   }
 
   return (
-    <Tag className={cc([baseClassName, displayClassName, className])}>
+    <Tag
+      className={cc([baseClassName, displayClassName, className])}
+      style={cssColorRegex.test(textColor) ? { color: textColor } : null}
```
pith: You could print a warning if the value doesn't match

src/text/story.tsx
```diff
@@ -4,17 +4,29 @@ import { storiesOf } from '@storybook/react'
 import { withInfo } from '@storybook/addon-info'
 import { selectV2, text, withKnobs, boolean } from '@storybook/addon-knobs'
 
+import { color } from '_utils/branding'
 import Text, { TextTagType, TextDisplayType } from 'text'
 
 const stories = storiesOf('Text', module)
 stories.addDecorator(withKnobs)
 
+const c = Object.keys(color).reduce(
```
pith: To remove ?

pith: Ah ok it's used above. Can you change the name to make it more explicit ?

src/text/story.tsx
```diff
@@ -4,17 +4,29 @@ import { storiesOf } from '@storybook/react'
...
+  (acc, key) => ({
+    ...acc,
+    [key]: color[key],
+  }),
+  {},
+)
+
+console.log(c)
```
pith: You can remove it

src/itinerary/__snapshots__/index.unit.tsx.snap
```diff
@@ -16,6 +16,7 @@ exports[`Itinerary component Should display proximity from departure point 1`] =
       >
         <span
           className="kirk-text kirk-text-titleStrong"
+          style={null}
```
pith: 😬 would be better if we could avoid this in the snapshot

src/text/index.tsx
```diff
@@ -47,19 +48,25 @@ const Text = ({
...
   display = TextDisplayType.BODY,
   tag = TextTagType.SPAN,
+  textColor,
   newlineToBr = true,
 }: TextProps) => {
   const baseClassName = 'kirk-text'
   const displayClassName = `${baseClassName}-${display}`
   const Tag = tag
 
+  const cssColorRegex = /^#([0-9a-f]{3}|[0-9a-f]{6})$/i
```
pith: Can you add a test with valid / invalid colors ?

Bidjii: Done

src/text/story.tsx
pith: We agree that this code doesn't do anything ... Or I didn't get it

Tuxosaurus: Oh yes... the result is... `color`

Tuxosaurus: I commited the change

src/text/index.tsx
wakooka: You mean a `console.warn` or `console.error`?

Tuxosaurus: hey @pith any further opinion on this?

PR #142, opened 2018-12-06
ToggleButton component

PerrineBoissieres:
## What does it do
This component serves as an on/off toggle button. It shares the same design as the `Checkbox` component but is meant to be used on its own (and not inside a form).

## Screenshots
capture d ecran 2018-12-06 a 14 20 39
capture d ecran 2018-12-06 a 14 20 49
capture d ecran 2018-12-06 a 14 21 02

## What changes on existing components
Added a `thin` prop on `CheckIcon` and `CircleIcon` so they can be displayed with a thinner stroke width, to match the `Checkbox` design (which is made with CSS, but will need to be updated).
Should a I realign the `Checkbox` component (`Item` variation, using icons and not CSS, etc...) in this PR as well?


Tuxosaurus:
> Added a thin prop on CheckIcon and CircleIcon so they can be displayed with a thinner stroke width, to match the Checkbox design (which is made with CSS, but will need to be updated).

Can't this thinner look be the new default style? Or should we add it for all icons?

> Should a I realign the Checkbox component (Item variation, using icons and not CSS, etc...) in this PR as well?

I'm fine with a separate PR

PerrineBoissieres:
> > Added a thin prop on CheckIcon and CircleIcon so they can be displayed with a thinner stroke width, to match the Checkbox design (which is made with CSS, but will need to be updated).
> 
> Can't this thinner look be the new default style? Or should we add it for all icons?
> 
> > Should a I realign the Checkbox component (Item variation, using icons and not CSS, etc...) in this PR as well?
> 
> I'm fine with a separate PR

I'm hesitant for default style, unless everyone is ok with a breaking change :)

Bidjii:
51174ced-198e-4205-bbd0-6dd926b2e42f

Here are the rules according to the specs

PR #143, opened 2018-12-07
refactor Profile using Item

Tuxosaurus:
profile_default
profile_medium


src/profile/index.tsx
```diff
@@ -1,58 +1,61 @@
...
-      {action &&
-        <div>
-          <Button status={Button.STATUS.UNSTYLED} title={action}>
-            <ArrowIcon right />
-          </Button>
-        </div>
-      }
-      <style jsx>{style}</style>
-    </div>
+      chevron={!!isLink}
```
wakooka: `{isLink}` instead because it's already a boolean

Tuxosaurus: Fixed

src/profile/index.tsx
```diff
@@ -1,58 +1,61 @@
...
+  picture,
+  alt,
+  isIdChecked,
+  isMedium,
+  isLink,
+  score,
+  ratings,
+  ratingsLabel,
+}: ProfileProps) => {
+  const getLeftBody = () => (
```
pith: nit: doesn't need to be a function

Tuxosaurus: Right, changed

PR #148, opened 2018-12-13
Improve accessibility of the textfield and snackbar

pith:
- Add possibility to have a role attribute on `<Text>`
- Add the role "alert" on `<Snackbar>` for accessibility
- Add aria-hidden on the `<Textfield>` clear button

### DIFF
--- a/src/text/index.tsx
+++ b/src/text/index.tsx
@@ -29,6 +29,7 @@ interface TextProps {
   readonly tag?: TextTagType
   readonly textColor?: string
   readonly newlineToBr?: boolean
+  readonly role?: string
 }
 
 const replaceNewLineWithBR = (str: string): React.ReactNode =>
@@ -50,6 +51,7 @@ const Text = ({
   tag = TextTagType.SPAN,
   textColor,
   newlineToBr = true,
+  role,
 }: TextProps) => {
   const baseClassName = 'kirk-text'
   const displayClassName = `${baseClassName}-${display}`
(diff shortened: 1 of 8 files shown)


src/text/index.tsx
```diff
@@ -58,8 +61,10 @@ const Text = ({
     content = replaceNewLineWithBR(children)
   }
 
+  const attrRole = !isEmpty(role) ? { role } : {}
+
   return (
-    <Tag className={cc([baseClassName, displayClassName, className])}>
+    <Tag {...attrRole} className={cc([baseClassName, displayClassName, className])}>
```
loick: you can put role with null as a default value and put it directly on Tag?

src/textField/index.tsx
```diff
@@ -226,6 +230,7 @@ export default class TextField extends PureComponent<TextFieldProps, TextFieldSt
               onClick={buttonOnClick}
               tabIndex="-1"
               title={buttonTitle}
+              aria-hidden={isEmpty(buttonTitle)}
```
wakooka: is buttonTitle a string? in that case shouldn't you do `{!!buttonTitle}` instead?

pith: `isEmpty` is ok and is more readable I think.

```
> isEmpty('')
true
> isEmpty('plop')
false
```

src/text/index.tsx
Tuxosaurus: If we want to prevent misuses of this component, we might as well have 'alert' as the default value

pith: It's defined as alert for the snackbar, but it shouldn't be the default for the Text component ;)

PR #159, opened 2019-01-08
[BBCSPA-1099] Move keycodes into a util

loick:


src/_utils/keycodes.ts
```diff
@@ -0,0 +1,5 @@
+export const KEYCODES = {
```
Tuxosaurus: We only have one currently but in the event we have many more, shouldn't we import only those we need instead of the whole keycodes set?

loick: it's an object, but yeah I was thikning at first to work with an enum like: 
keyCodeIs(keyCode, ESCAPE). 

But in the end, for now since we only have one, I just moved it in its own folder before upgrading it too much for nothing :/

Tuxosaurus: Ok, let's update it when/if needed

PR #160, opened 2019-01-08
[BBCSPA-2164] Running prettier on all of the codebase

loick:


loick:
Ok, done. Prettier check in CI.

PR #162, opened 2019-01-09
Fix SuccessModal display with small content

nickoola:


nickoola:
capture d ecran 2019-01-09 a 18 08 35
capture d ecran 2019-01-09 a 18 08 12

PR #164, opened 2019-01-11
Add format prop to TextField

PerrineBoissieres:
The new `format` prop allows to update the value each time it changes. This can be helpful for some inputs where you want the value to match a certain pattern (ex: a date, a phone number) and autofill some characters along.

In the story, I've put an example with a date input where you autofill the slash characters when the user is typing. Be mindful that this updates the value (this is the big difference between the TextField `format` and the Stepper `format` props).

### DIFF
--- a/src/textField/index.unit.tsx
+++ b/src/textField/index.unit.tsx
@@ -236,9 +236,15 @@ it('Simulates a blur event.', () => {
   expect(onTextFieldBlur).toHaveBeenCalledTimes(1)
 })
 
-it('Be able to format the value', () => {
+it('Should format the default value', () => {
   const wrapper = mount(
     <TextField name="test" defaultValue="Hello" format={value => `${value} world`} />,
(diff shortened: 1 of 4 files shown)


src/textField/index.unit.tsx
```diff
@@ -235,3 +235,10 @@ it('Simulates a blur event.', () => {
   wrapper.find('input').simulate('blur')
   expect(onTextFieldBlur).toHaveBeenCalledTimes(1)
 })
+
+it('Be able to format the value', () => {
+  const wrapper = mount(
+    <TextField name="test" defaultValue="Hello" format={value => `${value} world`} />,
+  )
+  expect(wrapper.find('input').prop('value')).toBe('Hello world')
+})
```
wakooka: Can you add a second expect with a `setProps()` to test that it's also updating correctly?

PR #168, opened 2019-01-17
Add newlineToBr prop to SuccessModal component

nickoola:


pith:
This behaviour is already available in the Text component why not just reusing it ?

src/text/index.unit.tsx
```diff
@@ -60,14 +60,14 @@ describe('Text', () => {
     ).toBe('span')
   })
 
-  it('Should replace \n by br tags', () => {
+  it('Should replace \n with br tags by default', () => {
```
Tuxosaurus: I guess these tests can be removed since you are testing this behaviour in _utils

src/successModal/index.unit.tsx
```diff
@@ -10,6 +10,10 @@ const defaultProps = {
   large: false,
 }
 
+const multipleLineText = `line1
```
Tuxosaurus: You can remove this const

src/text/index.unit.tsx
nickoola: It's more to test the `newlineToBr` prop

src/successModal/index.unit.tsx
nickoola: Yep done

src/text/index.unit.tsx
Tuxosaurus: ok :)

PR #170, opened 2019-01-17
[BBCSPA-2582] Itinerary component key fixes

mcellier:


src/itinerary/index.unit.tsx
```diff
@@ -123,4 +123,56 @@ describe('Itinerary component', () => {
     const itinerary = renderer.create(<Itinerary places={places} />).toJSON()
     expect(itinerary).toMatchSnapshot()
   })
+
+  it('Should use subLabel in key if it\'s a string', () => {
```
pith: nit: you can use double quotes to avoid escaping single quote

`"Should use subLabel in key if it's a string"`

src/itinerary/index.unit.tsx
```diff
@@ -123,4 +123,56 @@ describe('Itinerary component', () => {
...
+        distanceFromPoint: '1,5km',
+        time: '09:00',
+        isoDate: '2017-12-11T09:00',
+        subLabel: 'rue Ménars',
+        mainLabel: 'Paris',
+      },
+      {
+        time: '12:00',
+        isoDate: '2017-12-11T12:00',
+        subLabel: <Proximity value="FAR" title="Pick up point is quite far fom your place" />,
```
pith: nit: you can add an expectation for this case also

PR #173, opened 2019-02-06
Add ItemAction component

pith:
image


src/itemAction/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  readonly action?: string
+  readonly subLabel?: string
+  readonly leftAddon?: React.ReactNode
+  readonly onClick?: (event: React.MouseEvent<HTMLElement>) => void
+  readonly onBlur?: (event: React.FocusEventHandler<HTMLElement>) => void
+  readonly onFocus?: (event: React.FocusEventHandler<HTMLElement>) => void
+  readonly onMouseDown?: (event: React.MouseEvent<HTMLElement>) => void
+}
+
+class ItemAction extends PureComponent<ItempActionProps> {
```
Bidjii: Shouldn't this be a functional component?

Bidjii: And why is it `ItempActionProps` and not `ItemActionProps`? ;)

src/_utils/item/style.ts
```diff
@@ -1,19 +1,40 @@
...
+import { color, space } from '_utils/branding'
 
 export default css`
   :global(.kirk-item) {
     position: relative;
     display: flex;
     padding: 0;
-    padding-top: ${space.l};
-    padding-bottom: ${space.l};
+    padding: ${space.l} ${space.xl};
```
Bidjii: This is huge, it's going to impact all the Items throughout the lib (not only the `ItemAction`). Not a problem per se, just wanted to point that out.

pith: Yep I know that is why I listed it in the breaking changes

src/itemAction/index.tsx
pith: Fixed thanks for the feedback ! :)

src/_utils/item/style.ts
```diff
@@ -1,19 +1,40 @@
...
     align-items: center;
     flex: 1;
     border: 0;
     background: none;
   }
+  :global(.kirk-item.kirk-item--clickable:hover) {
+    background: ${color.pressed};
```
Tuxosaurus: why 'pressed' and not 'hover' (since it's not pressed - also if it was pressed, the status would be "active")

PR #174, opened 2019-02-07
[BBCSPA-2708] Remove the sublabel padding override

loick:


src/itemChoice/index.tsx
```diff
@@ -75,7 +75,6 @@ class ItemChoice extends PureComponent<ItemChoiceProps> {
       prefix({
         itemChoice: true,
         'itemChoice--highlighted': highlighted,
-        'itemChoice--withSubLabel': !!subLabel,
```
Tuxosaurus: Since we're talking about refactoring ItemChoice, I'd prefer to change the style in the app rather than in the library.
Something like:
`.ItemChoiceList > li:first-child {
    padding-top: 16px;
}`

loick: Acutally, we should never target the classes and change the CSS/design inside the library component from the app

loick:
> `.kirk-itemChoice--withSubLabel` needs to be also removed from the `style.ts` if not used anymore

Removed, thanks!

src/itemChoice/index.tsx
Tuxosaurus: Ok I misinterpreted what you needed the change for.

PR #179, opened 2019-02-14
[BBCSPA-2616] Button API update (icon to bubble, no more span, no more ellipsis)

loick:


package.json
```diff
@@ -116,27 +116,6 @@
     "tslint-config-airbnb": "5.11.1",
     "tslint-config-prettier": "1.17.0"
   },
-  "jest": {
```
mcellier: Ninja refactor in an unrelated PR? :)

src/textField/index.tsx
```diff
@@ -63,18 +65,23 @@ export interface TextFieldState {
...
       className,
     })
   }
-  return (
-    <span role="alert" className={className}>
-      {error}
-    </span>
-  )
+
+  if (typeof error !== 'boolean') {
```
mcellier: So, nothing is done if it's a boolean?

src/textField/index.tsx
```diff
@@ -32,14 +32,16 @@ export interface CommonFormFields {
...
+
 export interface TextFieldProps extends CommonFormFields {
   isTextArea?: boolean
   defaultValue?: string
   labelledBy?: string
   onChange?: (obj: OnChangeParameters) => void
   onClear?: () => void
   className?: Classcat.Class
-  error?: string | JSX.Element
+  error?: errorField
```
mcellier: Shouldn't the boolean behavior be explained in the changelog and visible in the story?

src/uneditableTextField/index.tsx
```diff
@@ -43,7 +48,7 @@ export const UneditableTextField = ({
         {children}
       </div>
       <style jsx>{style}</style>
-    </Component>
+    </Fragment>,
```
mcellier: Why fragment?

package.json
loick: We have a lot of errors on the tslint triggered by the extension on vscode. I don't know why, and I didn't really want to dig into it in this PR but indeed, I fixed some stuffs breaking my files through this :)

src/uneditableTextField/index.tsx
loick: We cannot add several components at once, actually `React.createElement` is expecting only one child

src/uneditableTextField/index.tsx
```diff
@@ -19,20 +19,25 @@ export const UneditableTextField = ({
...
+    componentType = 'a'
     props = { href }
   } else {
-    Component = 'div'
+    componentType = 'div'
   }
 
-  return (
-    <Component className={cc(['uneditableTextField', className])} {...props}>
+  return React.createElement(
```
loick: Btw, I moved the "Component" tag into a React.createElement, since this is recommended, and compliant to Typescript

src/textField/index.tsx
loick: Good point, I'm going to remove this.

Edit: Wow, I just moved back to check the implementation of this component, and Error used to be a boolean to put a red background, not the error message, and then the behavior changed.. I'm gonna update this

package.json
loick: And actually I rolled back the biggest part of it :) About this, it's just moved in the jest.config.js

src/textField/index.tsx
loick: Boolean removed, I guess this is a legacy from a previous implementation of the Textfield... Weird, good catch!

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Unreleased
 
+- **[UPDATE]** `TextField` no more boolean accepted in the error field
```
mcellier: **[BREAKING CHANGE]**?

loick: Done

src/button/index.tsx
```diff
@@ -23,9 +23,8 @@ export interface ButtonProps {
   readonly title?: string
   readonly status?: ButtonStatus
   readonly focus?: boolean
-  readonly icon?: boolean
+  readonly bubble?: boolean
```
Tuxosaurus: I'd prefer `isBubble`

loick: done

PR #182, opened 2019-02-18
Do not return href null when no href

pith:
**[FIX]** `Item` should not have an empty href on non `A` tags.

### DIFF
--- a/src/_utils/item/index.unit.tsx
+++ b/src/_utils/item/index.unit.tsx
@@ -45,10 +45,10 @@ it('Should return a tag A when href is a string', () => {
   expect(wrapper.prop('href')).toBe('#')
 })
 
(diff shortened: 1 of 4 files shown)


src/_utils/item/index.unit.tsx
```diff
@@ -1,9 +1,17 @@
 import React from 'react'
+import renderer from 'react-test-renderer'
 
 import Item from './index'
 import ClockIcon from 'icon/clockIcon'
 import ChevronIcon from 'icon/chevronIcon'
 
+it('Should not have changed', () => {
```
Tuxosaurus: nit: I feel "Should not have changed" is awkward and might get confused with what's expected of a feature.
=> "Should match snapshot"

PR #183, opened 2019-02-21
Set Button to inline-flex to solve alignment issues

PerrineBoissieres:
We had regressions when importing 0.22.1 version as Button used with links were full width and Button used as TextField add-ons had icon-alignment issues.

This fixes both problems, and also added a missing unit test on the TextField component.

### DIFF
--- a/src/textField/index.tsx
+++ b/src/textField/index.tsx
@@ -68,9 +68,13 @@ export interface TextFieldState {
 const DisplayError = (error: errorField) => {
   const className = 'kirk-error-message'
 
-  return React.isValidElement(error)
-    ? React.cloneElement(error, { className } as Object)
-    : <span role="alert" className={className}>{error}</span>
+  return React.isValidElement(error) ? (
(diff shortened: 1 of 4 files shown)


Bidjii:
Changelog update?

src/textField/index.tsx
```diff
@@ -68,9 +68,13 @@ export interface TextFieldState {
 const DisplayError = (error: errorField) => {
   const className = 'kirk-error-message'
 
-  return React.isValidElement(error)
-    ? React.cloneElement(error, { className } as Object)
-    : <span role="alert" className={className}>{error}</span>
+  return React.isValidElement(error) ? (
+    React.cloneElement(error, { className } as Object)
+  ) : (
+    <span role="alert" className={className}>
```
Bidjii: I know that you're only re-indenting stuff here, but shouldn't this be a `<Text>` here?
Just sayin' ;D

PR #186, opened 2019-02-22
[BBCSPA-2778] Add call and send icon

pith:
| Call | Send |
|-----|------|
| image | image |


Bidjii:
I think you need to change all those `fill-rule` and stroke rules into camelCase, you'll get errors in Kairos if not

wakooka:
You probably get those warning in Storybook already.

pith:
Fixed thanks ! ❤️

PR #188, opened 2019-02-26
Create SelectField component

nickoola:


nickoola:
capture d ecran 2019-02-26 a 14 44 03
capture d ecran 2019-02-26 a 14 44 18
capture d ecran 2019-02-26 a 14 44 35
capture d ecran 2019-02-26 a 14 45 06

nickoola:
capture d ecran 2019-02-26 a 14 45 06

Bidjii:
No `autofocus`, `disabled` or `required` props?

Bidjii:
Also wouldn't it be good to have the possibility to have a default option selected?

src/selectField/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const baseClassName = 'kirk-selectField'
+
+  return (
+    <div className={cc([baseClassName, className])} ref={forwardedRef}>
+      <select
+        name={name}
+        onChange={event => onChange({ name, value: event.target.value })}
+        defaultValue={value}
+      >
+        <optgroup label={optGroupLabel}>
```
Tuxosaurus: I guess there should be a `optGroups` prop, an array of objects each with their own label and options (and label should be optional, no `optGroup` tag if no label).

nickoola:
@Bidjii The default option correspond to the value set to the value prop passed to `defaultValue` select attribute.
Good point for event/status attribute. Not sure we need it for now cause we'll use this component as a sub-component for the phoneField component (select phone extensions + input tel).
I'm gonna add some props in case of using SelectField alone.

src/selectField/index.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { Ref } from 'react'
+import cc from 'classcat'
+import { ChevronIcon } from 'icon'
+import { color } from '_utils/branding'
+import style from 'selectField/style'
+
+export interface SelectFieldProps {
```
Tuxosaurus: extends CommonFormFields (from TextField... maybe it should be moved to _utils?)

And for select I would add `multiple`

src/selectField/index.tsx
nickoola: Agree about `optGroup` not required, I added it to manage the fact there is no placeholder and no label bellow select in Pixar library.

Bidjii:
> @Bidjii The default option correspond to the value set to the `value` prop passed to `defaultValue` select attribute.

Yeah but is the value displayed inside the closed select without putting a `selected` attribute onto the right option?

nickoola:
@Bidjii React uses `value` and `defaultValue` on select instead of `selected` attribute on option.
https://stackoverflow.com/questions/44786669/warning-use-the-defaultvalue-or-value-props-on-select-instead-of-setting?rq=1

src/selectField/index.tsx
nickoola: Done for the `CommonFieldsProps`
For the `multiple` attribute I do not add it for now cause it totally break the component display.
Don't find a way to manage the component height according to number of option.
And we don't need this multiple options for now.

src/_utils/interfaces.ts
```diff
@@ -0,0 +1,10 @@
+export interface CommonFieldsProps {
```
Tuxosaurus: You didn't remove it from TextField?

src/selectField/index.tsx
Tuxosaurus: Maybe it's not what you need but there is a `size` attribute you can pass along with `multiple` to choose how many options are displayed when opening the select

src/_utils/interfaces.ts
nickoola: I especially focus on this new component but yes it's not a big deal to use this utils with the TextField component.

src/selectField/index.tsx
nickoola: You right `size` attribute fix the height issue but an other scrollbar display issue need to be fixed before adding `multiple` attribute option.
capture d ecran 2019-02-27 a 15 00 17
I'm creating issues (multiple option for SelectField + CommonFieldsProps refacto for TextField) to keep post about what we plan to improve.

src/selectField/index.tsx
nickoola: At the end I added `ariaLabel` prop to keep accessibility if no `<label>` tag linked ton the `select` tag.

src/selectField/index.tsx
```diff
@@ -0,0 +1,62 @@
+import React, { Ref } from 'react'
+import cc from 'classcat'
+import { ChevronIcon } from 'icon'
+import { color } from '_utils/branding'
+import { CommonFieldsProps } from '_utils/interfaces'
+import style from 'selectField/style'
+
+interface SelectFieldProps extends CommonFieldsProps {
+  readonly ariaLabel?: string
+  readonly options: Object[]
```
wakooka: Can you add a specific object like we did for `autocompleteItem`?
https://github.com/blablacar/ui-library/blob/master/src/typings/autocomplete.d.ts

src/selectField/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+        aria-label={ariaLabel}
+        onChange={event => onChange({ name, value: event.target.value })}
+        onFocus={onFocus}
+        onBlur={onBlur}
+        defaultValue={value}
+        disabled={disabled}
+        required={required}
+        autoFocus={autoFocus}
+      >
+        {options.map(({ value, label }: any) => (
```
wakooka: Like the comment above, use the same Object type.

src/selectField/index.tsx
nickoola: Done!

src/selectField/index.tsx
nickoola: Done!

src/selectField/style.ts
```diff
@@ -0,0 +1,56 @@
...
+    margin: 0;
+    padding: 0;
+    font-size: ${font.base.size};
+    line-height: ${font.base.lineHeight};
+    color: ${color.primaryText};
+    box-shadow: none;
+    background: transparent;
+    -webkit-appearance: none;
+    -moz-appearance: none;
+    -ms-appearance: none;
```
wakooka: remove prefixes

src/selectField/index.tsx
nickoola: Done

src/selectField/style.ts
nickoola: Done

PR #192, opened 2019-03-05
Do not render to/from itinerary addons with whitespace strings

grosbouddha:
Unit tested and tested locally in the storybook demo page by inserting temporary whitespace strings.


Tuxosaurus:
Don't forget the changelog :)

Also, not sure if it is needed but the addon knobs in the story could be using text instead of booleans to reflect this fix.

src/itinerary/index.tsx
```diff
@@ -27,12 +30,12 @@ const Itinerary = ({
 }: ItineraryProps) => {
   return (
     <ul className={cc([className, { 'kirk-itinerary--small': small }])}>
-      {headline && (
+      {isNonEmptyString(headline) && (
```
loick: why not directly using !isEmpty? Since Typescript will enforce the usage of string anyway

grosbouddha: To cover cases with strings made of whitespace spaces (e.g."   ").
I could rename isNonEmptyString to something more explicit to make it clear e.g. isEmptyOrWhitespace ?

loick: Oh, right for the whitespace!

About the naming it's fine like that, thanks for the explanation :)

grosbouddha:
> Don't forget the changelog :)
> 
> Also, not sure if it is needed but the addon knobs in the story could be using text instead of booleans to reflect this fix.

Added the storybook modification suggestions and the changelog entry, thanks!

grosbouddha:
> Don't forget the changelog :)
> 
> Also, not sure if it is needed but the addon knobs in the story could be using text instead of booleans to reflect this fix.

Added the storybook modification suggestions and the changelog entry, thanks!

grosbouddha:
PTAL

PR #193, opened 2019-03-06
Add fade-in animation to drawer dimmer

grosbouddha:
The new animation spec (300ms) comes from the v3/dashvboard dimmer (confimed with Nico Duval).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
-[...]
+- **[FIX]** Fix fade-in animation for Drawer dimmer
+- [...]
 
 # v3.0.0 (12/04/2019)
 
--- a/src/drawer/style.ts
+++ b/src/drawer/style.ts
@@ -15,20 +15,20 @@ export default css`
     overflow-y: hidden;
   }
 
-  .kirk-drawer--open::after {
+  .kirk-drawer::after {
     content: '';
+    font-size: 0;
+  }
+
+  .kirk-drawer--open::after {
     position: absolute;
     top: 0;
     left: 0;
     bottom: 0;
     right: 0;
     z-index: 1;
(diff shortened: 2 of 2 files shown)


src/drawer/style.ts
```diff
@@ -15,20 +15,20 @@ export default css`
...
     bottom: 0;
     right: 0;
     z-index: 1;
-    background-color: rgba(0, 0, 0, 0);
-    transition: background-color 150ms linear;
-  }
-
-  .kirk-drawer--open::after {
     background-color: rgba(0, 0, 0, 0.3);
+    transition: background-color 300ms linear;
```
Bidjii: If I may intervene here, should this `300ms` duration go in the `branding` file?

nickoola: We already have `transition: background-color ${transition.duration.fast} linear;` for example but it setted to `200ms`.

wakooka: Yes can you use one of the value in the branding file? Or if it does'n fit your case add a new one
https://github.com/blablacar/ui-library/blob/master/src/_utils/branding.ts#L82

grosbouddha: Updated to ${transition.duration.base} , thanks!

grosbouddha:
PTAL

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@
 - **[UPDATE]** Removing css linter (legacy, targeting CSS files which does not exist anymore on this repository)
 - **[UPDATE]** Moving to styled-jsx V3
 - **[UPDATE]** Re-organize the dependencies (dev & prod dependencies)
+- **[FIX]** Fix fade-in animation for Drawer dimmer
```
wakooka: Move this to `Unreleased` section

grosbouddha: Done

PR #194, opened 2019-03-07
Add fix for Menu onClick items + tests

grosbouddha:


src/menu/story.tsx
```diff
@@ -84,6 +84,13 @@ const menuItems = [
     leftAddon: <CrossDiscIcon />,
     href: 'menu-item-10',
   },
+  {
+    id: 'menu-item-11',
+    label: 'Menu item with onClick',
+    onClick: () => {
+      window.alert('Item clicked')
```
Bidjii: Maybe you could use the Storybook `action` plugin here :)

wakooka: Yes e.g. `onClick={action('onClick')}`and it will appear in the Action Logger tab in storybook

grosbouddha: Done. Thanks for the tip!

PR #196, opened 2019-03-08
Migration to React Day Picker

PerrineBoissieres:
## Why?
This PR has one big objective: remove our dependency to React Dates and Moment.
The combination of the two adds a ton of weight to the library, and forces the dependency to Moment to any project using the library.

Looking at alternatives, React Day Picker seems like a decent one: very light, good Typescript support (the author even plans to rewrite it entirely in TS), regular updates, no reliance on third-party lib for styling and localization, and similar features than what we really used out of React Dates.

### API - What has changed
- `daySize`, `hideKeyboardShortcutsPanel`, `weekDayFormat` and `monthFormat` have been removed (if there is any issue with their removal, please raise your hand)
- `firstDayOfWeek`, `weekdaysShort`, `weekdaysLong` and `months` are now used for localization since the component doesn't support anything else than english by itself
- `stickyPositionTop` helps customize the position of the sticky weekdays in vertical mode
- `fromMonth` indicates the default month to show by giving "starting date" of the calendar (defaults to today, so current month), it has no impact on `isOutsideRange` prop
- `initialDate` now use standard `Date` object instead of `moment`

### Performance usage
Removing the dependency to moment and React Dates saved around ~0.5MB when importing the library into a project (after compilation). The DatePicker is also slightly faster on mount, and consumes less memory than its former version.


src/datePicker/index.unit.tsx
```diff
@@ -1,22 +1,24 @@
-import React from 'react'
-import moment from 'moment'
-import DatePicker from 'datePicker'
+// import React from 'react'
```
loick: If you comment all of the tests, we cannot know how this is supposed to be working, and how we could migrate it to SPA

PerrineBoissieres: Tests have been added ;)

src/datePicker/index.tsx
```diff
@@ -1,132 +1,194 @@
...
-const navNext = <ArrowIcon right iconColor={color.primary} />
+const defaultWeekdaysLong = [0, 1, 2, 3, 4, 5, 6].map(weekday =>
+  DayPicker.LocaleUtils.formatWeekdayLong(weekday),
+)
+const defaultWeekdaysShort = [0, 1, 2, 3, 4, 5, 6].map(weekday =>
+  DayPicker.LocaleUtils.formatWeekdayShort(weekday),
+)
+const defaultMonths = DayPicker.LocaleUtils.getMonths()
+
+enum Orientation {
```
mcellier: Should probably be exported for external use to prevent clients from duplicating this.

PerrineBoissieres: It's available as a static with `DatePicker.constants` ;)

src/datePicker/index.tsx
```diff
@@ -1,132 +1,194 @@
...
+import Button from 'button'
+
+import style from 'datePicker/style'
+
+const [BASE_CLASSNAME] = prefix({ datepicker: true })
 
-const navPrev = <ArrowIcon iconColor={color.primary} />
-const navNext = <ArrowIcon right iconColor={color.primary} />
+const defaultWeekdaysLong = [0, 1, 2, 3, 4, 5, 6].map(weekday =>
+  DayPicker.LocaleUtils.formatWeekdayLong(weekday),
```
mcellier: Is it using Intl to translate by default? If so, is it using the browser locale?

PerrineBoissieres: It's not using Intl, ReactDayPicker has basic english default settings (with months and weekdays definitions).
So what happens is that our DatePicker is english by default, and you can override everything with the given props.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **[BREAKING CHANGE]** Migrate `DatePicker` to ReactDayPicker and remove React Dates
```
mcellier: It may be useful to develop a bit: 
- Which properties are removed
- Removing forced dependency to moment

src/datePicker/index.tsx
mcellier: nit: Maybe add some doc in the storybook to explain how to use Intl for this.

CHANGELOG.md
PerrineBoissieres: I added a bit more details on what changed.

src/datePicker/index.tsx
PerrineBoissieres: Done, that's a good point even though the story code provides an example, at least it is visible at first glance.

PR #201, opened 2019-03-13
[1.0.0] Changelog updated

loick:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@blablacar/ui-library",
-  "version": "0.24.1",
+  "version": "1.0.0",
```
Tuxosaurus: Should we update the Readme by making clear this is still rather an alpha version?

loick: Why is it an alpha version ?

Tuxosaurus: I think we're missing to many "guidelines" and explanations (and support) to leave this out in the wild.

Tuxosaurus: Ok as discussed, adding "guidelines" is planned.

PR #207, opened 2019-03-20
[BBCSPA-2933] Remove storybook/addon-info plugin

loick:


mcellier:
We don't have any hooks preventing us to commit files with unused vars?

loick:
> We don't have any hooks preventing us to commit files with unused vars?

Supposed to, but I guess it was merged with a no verify or during a rebase...

PR #219, opened 2019-04-03
Refactor and change disabled color for preference icons

Tuxosaurus:
LightGray and Gray are not visible enough for icons with status OFF


Tuxosaurus:
Still need to figure out something wrong in the default color selection in the Story

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - [...]
+- **[UPDATE]** Rafactor and change disabled color for `musicIcon`, `petIcon`, `smokeIcon`
```
Bidjii: `Rafactor` > `Refactor`

src/icon/petIcon.tsx
```diff
@@ -4,14 +4,14 @@ import isEmpty from 'lodash.isempty'
 
 import { color } from '_utils/branding'
 
-export enum PetType {
+enum Status {
   ON = 'on',
   OFF = 'off',
   DEFAULT = 'default',
```
grosbouddha: Why do we need the DEFAULT state?
IMO, the default state should be either ON or OFF and defined at the component call site, not a third state inside the component.
wdyt?

Tuxosaurus: We have 3 states for preferences (actually 4 since it can be "never changed" / "unselected" but in this case we don't display anything):
Screen Shot 2019-04-04 at 14 11 40

Tuxosaurus:
Before:
Screen Shot 2019-04-04 at 15 56 28

After:
Screen Shot 2019-04-04 at 15 53 49

src/icon/petIcon.tsx
grosbouddha: I see, thanks for the context.

src/icon/musicIcon.tsx
```diff
@@ -4,14 +4,14 @@ import isEmpty from 'lodash.isempty'
 
 import { color } from '_utils/branding'
 
-export enum MusicType {
+enum statusList {
```
mcellier: - Theses statuses are used by the 3 icons. Shouldn't it be put somewhere else?
- Are theses constants available outside? Like `MusicIcon.status.DEFAULT` or something?

src/icon/smokeIcon.tsx
```diff
@@ -37,93 +38,51 @@ class SmokeIcon extends PureComponent<SmokeIconProps> {
...
-              fill={color.iconFaded}
-              fillRule="nonzero"
-              strokeLinecap="round"
-              strokeLinejoin="round"
-            />
-          </g>
-        )}
+        <g>
+          <path
+            fill={finalColor}
```
mcellier: Can't the fill be defined in the `<g>`? That'll make us repeat ourselves a bit less

Tuxosaurus: Seems to work 👍

src/icon/musicIcon.tsx
Tuxosaurus: Added a "_utils/icon/status" file with the status and the "Barred" element so that the 3 icons import them from this one source

PR #220, opened 2019-04-04
Added wheelchair icon

mcellier:


src/icon/wheelchair.tsx
```diff
@@ -0,0 +1,43 @@
+import React, { PureComponent } from 'react'
+import cc from 'classcat'
+import isEmpty from 'lodash.isempty'
+
+import { color } from '_utils/branding'
+
+class WheelchairIcon extends PureComponent<Icon> {
+  static defaultProps: Icon = {
```
grosbouddha: Out of scope for this PR but it hurts to see so much duplicated code in the Icon code :/

Tuxosaurus: Indeed...

src/icon/wheelchair.tsx
```diff
@@ -0,0 +1,43 @@
...
+        viewBox="0 0 100 100"
+        xmlns="http://www.w3.org/2000/svg"
+        className={cc(['kirk-icon', className])}
+        width={size}
+        height={size}
+        aria-hidden={isEmpty(title)}
+      >
+        {title && <title>{title}</title>}
+        <g fill={iconColor}>
+          <path
```
Tuxosaurus: Looks like the paths can be further optimized: https://jakearchibald.github.io/svgomg/

mcellier: Done. Thx.

src/icon/wheelchair.tsx
mcellier: #221

PR #223, opened 2019-04-10
[BBCSPA-1985] Added aria attributes and focus trap on Modals

wakooka:


package-lock.json
```diff
@@ -40,18 +40,18 @@
       }
     },
     "@babel/core": {
```
mcellier: Why so many updates???

mcellier:
Do we have a ticket to refactor modals to avoid code duplication?

Bidjii:
Create an issue 😬

wakooka:
> Do we have a ticket to refactor modals to avoid code duplication?

We do and I've already started :)

src/confirmationModal/index.tsx
```diff
@@ -135,7 +147,13 @@ class ConfirmationModal extends Component<ConfirmationModalProps> {
         <TransitionGroup component="div" className="transition-wrapper">
           {isOpen && (
             <CustomTransition animationName={AnimationType.SLIDE_UP}>
-              <div className={classNames}>
+              <div
+                className={classNames}
+                role="alertdialog"
```
grosbouddha: alertdialog should be for simple dialog with only text and buttons (like a window.prompt).
I think we should use the role dialog.

src/confirmationModal/index.tsx
```diff
@@ -135,7 +147,13 @@ class ConfirmationModal extends Component<ConfirmationModalProps> {
...
+                aria-labelledby={ariaLabelledBy}
+                aria-describedby={ariaDescribedBy}
+                aria-modal="true"
```
grosbouddha: FYI, last time I checked the screenreader support for aria-modal was really limited.
I'm not sure we want to aria-hidden background elements either (we used to do that in some past projects and it really confiuses some screenreaders).

src/confirmationModal/index.tsx
```diff
@@ -92,6 +97,11 @@ class ConfirmationModal extends Component<ConfirmationModalProps> {
     }
   }
 
+  onEntered = () => {
+    this.focusTrap = createFocusTrap(this.portalNode)
+    this.focusTrap.activate()
```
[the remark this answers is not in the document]
grosbouddha: Ok I found the doc for focus trap :)
`By default, when a focus trap is activated the first element in the focus trap's tab order will receive focus. With this option you can specify a different element to receive that initial focus. Can be a DOM node, or a selector string (which will be passed to document.querySelector() to find the DOM node), or a function that returns a DOM node.`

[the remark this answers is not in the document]
wakooka: I deactivate here:
https://github.com/blablacar/ui-library/pull/223/files#diff-b4fd3b3ee547244ee7e7f442dd7224cdR70

src/confirmationModal/index.tsx
wakooka: I checked with VoiceOver only, it didn't seem to be confused. You're right, the docs are saying that we should make the other elements "inert" which `aria-modal` does, the old way to do it is to trigger a `aria-hidden` on those elements manually, at least we're future proof

> The aria-modal property was introduced in ARIA 1.1. As a new property, screen reader users may experience varying degrees of support for it.
> Applying the aria-modal property to the dialog element replaces the technique of using aria-hidden on the background for informing assistive technologies that content outside a dialog is inert.

package-lock.json
wakooka: I did a fresh install maybe that's why.

PR #224, opened 2019-04-11
Fix Button broken display for loading and checked status with children

nickoola:


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - **[UPDATE]** Added aria attributes and focus trap on Modals
+- **[FIX]** `Button` do not display children on loading and checked status
```
mcellier: `Button` doesn't display children on `LOADING` and `CHECKED` statuses

src/button/index.tsx
```diff
@@ -189,7 +189,7 @@ export default class Button extends PureComponent<ButtonProps, ButtonState> {
             onDoneAnimationEnd={onDoneAnimationEnd}
           />
         )}
-        {children}
+        {!hasLoader && children}
```
mcellier: hasLoader does the job for LOADING & CHECKED ?

mcellier: Add unit tests

nickoola: Yep Done

CHANGELOG.md
nickoola: Done

Tuxosaurus: Sorry I'm late, I would say "shouldn't display", else I'm not sure if we're talking about the fix or the problem

src/button/index.tsx
nickoola: And should have a discussion about the collapsed display. I agree hasLoader is a bit more than just managing loading state.

PR #227, opened 2019-04-12
Added border on Textfield focus

wakooka:


Bidjii:
Changelog 😱

src/textField/style.ts
```diff
@@ -1,6 +1,9 @@
 import css from 'styled-jsx/css'
 import { color, space, font, radius, transition } from '_utils/branding'
 
+const borderSize = '1px'
+const borderFocusSize = '3px'
```
mcellier: Shouldn't they be in the `_utils/branding` file?

src/button/index.tsx
```diff
@@ -36,6 +36,7 @@ export interface ButtonProps {
   readonly onDoneAnimationEnd?: () => void
   readonly tabIndex?: string
   readonly disabled?: boolean
+  buttonRef?: (button: HTMLButtonElement) => void
```
mcellier: How is it related to textfield focus?

wakooka: There is a ⤫ to clear the field. And I don't want to lose the focus in the field when I click on it, so I need to know if what you click on to blur is this button. See: 
See https://github.com/blablacar/ui-library/pull/227/files/b76cd20dbfc692b88326228ff018030fe9632105#diff-c0fb58a964df8de2d8c9613483ec4cd4R152

src/textField/style.ts
wakooka: Discussed IRL, IMO, LGTM. We use the 1px border to keep the consistent padding to 16px and have the textfield height to 56px, since it's only used in the textfield we don't want to put it in a global branding file.

src/phoneField/index.tsx
```diff
@@ -193,6 +193,7 @@ export default class PhoneField extends PureComponent<PhoneFieldProps, PhoneFiel
           defaultValue={defaultPhoneValue}
           onChange={this.onChange}
           title={textFieldTitle}
+          focusBorder={false}
```
pith: Why do you need to disable focusBorder ?

wakooka: I updated the PR so that SelectField and PhoneField alos use the same border on focus.

PR #233, opened 2019-04-30
Move react into peer dependencies

wakooka:
Since we are now using hooks, we can't have 2 versions of React as a dependencies in a host project.
Moving the dependencies as `peerDependencies` will use only one version of React (the one defined in the host project)
More info explained here:
https://reactjs.org/warnings/invalid-hook-call-warning.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-[...]
+- **[BREAKING CHANGE]** Moved `react` and `react-dom` to `peerDependencies`
 
 # v3.2.0 (30/04/2019)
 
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -71,7 +71,9 @@
     "jsdom": "14.0.0",
     "jsdom-global": "3.0.2",
     "prettier": "1.16.4",
-    "react-test-renderer": "16.8.4",
+    "react": "16.8.6",
+    "react-dom": "16.8.6",
+    "react-test-renderer": "16.8.6",
```
mcellier: Shouldn't we have the same version as `react` and `react-dom`? If it's the case, and we're using peer dependencies, don't you think we can't have a fixed version?

wakooka: We don't need `react-test-renderer` because it's only a `devDependencies`

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Unreleased
 
-[...]
```
Bidjii: `peerDependecies` > `peerDependencies`

PR #234, opened 2019-04-30
Fix phoneField default prefix selection

nickoola:


src/phoneField/index.tsx
```diff
@@ -76,7 +76,7 @@ const mapCountriesPhoneData = (
   countryNames: customCountryNames,
 ): mappedCountryPhoneData[] => {
   return countryData.map(data => ({
-    value: data.dialCode,
+    value: data.iso2,
```
nickoola: To have unique value on <option> DOM elements

PR #238, opened 2019-05-28
Fix phoneField payload value

nickoola:


src/phoneField/index.tsx
```diff
@@ -39,7 +39,7 @@ interface PhoneFieldProps {
 interface PhoneFieldState {
   countryData: mappedCountryPhoneData[]
   countryWhitelist: string[]
-  phoneRegion: string
+  phonePrefix: string
   completePhoneNumber: string
   // Make the type of [key: string]: string etc... assignable for setState()
```
Tuxosaurus: nit: `etc...` => `etc.`

nickoola: Done

PR #250, opened 2019-06-07
ItemChoice migration and Items specs update

PerrineBoissieres:
It's back!

Now that we've added a couple of `Item` variants and `Item`-based components, `ItemChoice` is severely outdated and doesn't match the rest of its brethren. This is a massive breaking change.
Also, our latest design specs included more details and variants for Items using inputs (checkboxes and radios), so we need to put some order into what we have (and because we used the old `ItemChoice` for nearly everything).

### What's in there for you (TODO-list)
- [x] `ItemChoice` refactoring to be an `Item`-based component
- [x] `AutoComplete` refactoring to use the new `ItemChoice` component
- [x] `Menu` refactoring to use the new `ItemChoice` component
- [x] Removing old `Radio` component in favor of `ItemDeclaredChoice`
- [x] New component/props to handle the separators logic between `ItemChoice` components (instead of being automatic)
- [x] Use the new separator logic for `AutoComplete` and `Menu` components
- [x] Rename `ItemDeclaredChoice` to `ItemRadio`
- [x] Change `ItemRadio` input visual design to match latest specs
- [x] New `ItemCheckbox` component
- [x] Add specs to all `Item`-based components


src/_utils/radio/index.unit.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import renderer from 'react-test-renderer'
+
+import Checkbox from './index'
```
Tuxosaurus: For clarity, rename Checkbox to Radio

PerrineBoissieres: Missed this one, thanks!

mcellier:
😱 🙀

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 # Unreleased
 
 - [...]
+- **[BREAKING CHANGE]** Complete refactoring of `ItemChoice` to now use `Item` as its base component.
+- **[REMOVED]** `Checkbox`, `Radio` and `RadioGroup` have been removed.
```
mcellier: `**[BREAKING CHANGE]**`

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
...
+- **[RENAMING]** `ItemDeclaredChoice` and `ItemDeclaredChoiceGroup` have been renamed to `ItemRadio` and `ItemRadioGroup`
```
mcellier: `**[BREAKING CHANGE]**`

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
...
+- **[NEW]** `ItemsList` is a new component that's in charge of handling separators and listing of `ItemChoice`-like components.
+- **[NEW]** `ItemCheckbox` replaces the old `Checkbox` component and uses `Item` as its base component. Its API is similar to `ItemRadio`.
+- **[BREAKING CHANGE]** `AutoComplete`, `Menu` and `Drawer` now use `ItemChoice` and `ItemsList`, their API have been edited accordingly to `ItemChoice` API changes.
+- **[REMOVED]** `renderItem` prop on `AutoComplete` has been removed.
```
mcellier: `**[BREAKING CHANGE]**`

src/_utils/item/index.tsx
```diff
@@ -107,20 +122,30 @@ const Item = ({
...
               {rightBody}
             </Text>
           )}
         </div>
       )}
       {rightAddon && <div className="kirk-item-rightAddon">{rightAddon}</div>}
-      {chevron && <div className="kirk-item-rightAddon">{<ChevronIcon />}</div>}
+      {chevron && (
+        <div className="kirk-item-rightAddon">
+          {<ChevronIcon iconColor={!isClickable ? color.fadedText : color.secondaryText} />}
```
mcellier: Are these colors compatible with the left/right colors? We may end up with a rainbow-like item with this

src/_utils/radio/index.unit.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+describe('Checkbox', () => {
```
mcellier: nit: I think i'll be better to test explicitely what should be rendered, something like:

```
const radio = shallow(<Radio isChedked />)
expect (radio.find(CircleIcon).toExist()).toBe(true)
```

src/itemCheckbox/index.tsx
```diff
@@ -0,0 +1,86 @@
+import React, { Component, Fragment } from 'react'
+import cc from 'classcat'
+
+import Item from '_utils/item'
+import Checkbox from '_utils/checkbox'
+import { TextDisplayType } from 'text'
+
+import style from './style'
+
+export enum ItemCheckboxStatus {
```
mcellier: Can't you use the `ItemStatus` enum? 
In every component we have the `CHECKED` status here too, so for consistency' sake, it may be better to have this here.

src/itemRadio/index.tsx
```diff
@@ -2,17 +2,17 @@ import React, { Component, Fragment } from 'react'
...
 import Item from '_utils/item'
-import Checkbox from '_utils/checkbox'
+import Radio from '_utils/radio'
 import { TextDisplayType } from 'text'
 
 import style from './style'
 
-export enum ItemDeclaredChoiceStatus {
+export enum ItemRadioStatus {
   DEFAULT = 'default',
```
mcellier: Same thing as the CheckboxItem component: it may be better to use ItemStatus (or at least a put these in common)

src/itemsList/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { ItemRadioProps } from 'itemRadio'
+
+import style from './style'
+
+export const ItemsListDivider: FunctionComponent = () => null
+
+type ItemsListChild =
+  | React.ReactElement<ItemChoiceProps>
+  | React.ReactElement<ItemRadioProps>
+  | null
```
mcellier: Missing: `React.ReactElement<ItemCheckboxProps>`

src/itemsList/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+    className: '',
+    role: '',
+    keyGenerator: index => index,
+  }
+
+  static Divider = ItemsListDivider
+
+  render() {
+    const { children, className, withSeparators, keyGenerator, ...otherProps } = this.props
+    const { list } = React.Children.toArray(children).reduce(
```
mcellier: The following block is kinda hard to read. Could you split it a bit?

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 # Unreleased
 
 - [...]
+- **[BREAKING CHANGE]** Complete refactoring of `ItemChoice` to now use `Item` as its base component.
```
mcellier: It's internal, it shouldn't be declared as BREAKING (you already said that the props have been changed)

src/itemCheckbox/index.tsx
PerrineBoissieres: There is no `CHECKED` state for checkboxes and radios where we display the green check after loading, since the change in the component's look already indicates that the value has been correctly changed.

src/_utils/item/index.tsx
PerrineBoissieres: The chevron can only be two colors, from what I've seen: `secondaryText` by default, and a lighter grey when the Item is disabled. Even when the text is blue (recommended choice for example), the chevron stays grey.

I agree that there are a lot of color variations on Items, but I'm just following the specs here =/

src/_utils/radio/index.unit.tsx
PerrineBoissieres: Fixed!

CHANGELOG.md
PerrineBoissieres: Fixed!

CHANGELOG.md
PerrineBoissieres: Fixed!

CHANGELOG.md
PerrineBoissieres: Fixed!

src/itemsList/index.tsx
PerrineBoissieres: Fixed!

src/itemsList/index.tsx
PerrineBoissieres: I changed the approach (less `.reduce()` which I know isn't easy to read for a lot of people).

CHANGELOG.md
PerrineBoissieres: Fixed!

src/_utils/item/index.tsx
```diff
@@ -65,7 +74,7 @@ const Item = ({
       tagProps = { ...tagProps, href }
     }
   }
-  const isClickable = !!href || !!onClick
+  const isClickable = !!href || !!onClick || !!onMouseDown || tag.type === 'label'
```
wakooka: not sure to understand why adding `onMouseDown` and `tag.type === label`. It feels hacky, and easy to forget a use case.

src/_utils/radio/__snapshots__/index.unit.tsx.snap
```diff
@@ -0,0 +1,112 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Radio Should render a check icon if isChecked is set to true 1`] = `
+<svg
```
wakooka: Is it expected that there is no input radio in the snapshot?

wakooka: Oh wait `Radio`  is not an `input type="radio` I think this is confusing.

src/_utils/radio/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { color } from '_utils/branding'
+import Loader from 'loader'
+import CircleIcon from 'icon/circleIcon'
+
+interface RadioProps {
+  readonly isChecked?: boolean
+  readonly isLoading?: boolean
+}
+
+const Radio = ({ isChecked = false, isLoading = false }: RadioProps) => {
```
wakooka: I think we should rename this, see my other comment in the snapshot

src/drawer/story.tsx
```diff
@@ -112,7 +46,19 @@ class DrawerDemo extends Component {
...
+                leftAddon={<NewspaperIcon />}
+                status={ItemChoice.STATUS.LOADING}
+                href="#"
+              />
+              <ItemChoice label="Rides booked" leftAddon={<TicketIcon />} href="#" />
+              <ItemChoice label="Messages" leftAddon={<BubbleIcon />} href="#" />
+              <ItemChoice label="Ride alerts" leftAddon={<BellIcon />} href="#" />
+              <ItemChoice label="Menu item with onClick" href="#" />
+            </Menu>
           </Drawer>
```
wakooka: Nice! 👍

src/itemRadioGroup/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  onChange = ({ name, value }: OnChangeParameters) => {
+    this.props.onChange({ name, value })
+    this.setState({ value })
+  }
+
+  render() {
+    const { children, status, name, className, withSeparators, withChevrons } = this.props
+    return (
+      <ItemsList withSeparators={withSeparators} className={className}>
```
wakooka: Would be good to add a `role="radiogroup"`, and maybe I'll add a ticket to check on keyboard support
https://www.w3.org/TR/2017/WD-wai-aria-practices-1.1-20170628/examples/radio/radio-1/radio-1.html

src/_utils/item/index.tsx
PerrineBoissieres: Yeah, I was hesitating to create a new prop for handling the hover effect (instead of using other props).

`onMouseDown` is for the autocomplete items (that do not use `href` or `onClick`) and `label` is for checkboxes and radios.

src/_utils/radio/__snapshots__/index.unit.tsx.snap
PerrineBoissieres: Yeah, it's just the Radio icon lookalike ^^
Same thing for Checkbox, it just handles the different status.

src/_utils/item/index.tsx
PerrineBoissieres: I switched to the new isClickable prop. More stable than relying on something else.

src/_utils/radio/index.tsx
PerrineBoissieres: Renamed ;)

src/itemRadioGroup/index.tsx
PerrineBoissieres: Added the role attribute ;)

src/_utils/radio/index.unit.tsx
PerrineBoissieres: Snapshots are fine for html-rendering checks. I like both solutions, so, dunno, I'll do a poll to get others opinions ^^

PR #251, opened 2019-06-07
Add error property to phonefield component

nickoola:


nickoola:
Capture d’écran 2019-06-07 à 16 14 41
Capture d’écran 2019-06-07 à 16 15 03

src/phoneField/index.tsx
```diff
@@ -180,39 +194,44 @@ export default class PhoneField extends PureComponent<PhoneFieldProps, PhoneFiel
...
-          name={FIELDS.PHONENUMBER}
-          placeholder={textFieldPlaceholder}
-          defaultValue={defaultPhoneValue}
-          onChange={this.handleChange}
-          title={textFieldTitle}
-          onFocus={this.onFocus}
-          onBlur={this.onBlur}
-          focusBorder={!isInline}
-        />
+      <div className={cc([baseClassName, prefix({ error: !!error })])}>
```
grosbouddha: Boolean(error) instead of !!

src/phoneField/index.tsx
```diff
@@ -180,39 +194,44 @@ export default class PhoneField extends PureComponent<PhoneFieldProps, PhoneFiel
...
+        <div id={id} className={classNames} aria-labelledby={ariaLabelledBy}>
```
grosbouddha: I'm not sure how this aria label will work in screen readers, we will have bropwse focus on this container but most likely on a nested focusable inputs inside.

grosbouddha: My gut feeling will be to just remove it.

src/phoneField/index.tsx
PerrineBoissieres: For my part I prefer the short syntax for boolean casting, but that's a matter of preferences.

src/phoneField/index.tsx
```diff
@@ -180,39 +194,44 @@ export default class PhoneField extends PureComponent<PhoneFieldProps, PhoneFiel
...
+            name={FIELDS.PHONEREGION}
+            options={this.state.countryData}
+            ariaLabel={selectFieldLabel}
+            defaultValue={this.props.defaultRegionValue}
+            onChange={this.handleChange}
+            onFocus={this.onFocus}
+            onBlur={this.onBlur}
+            focusBorder={!isInline}
+          />
+          <TextField
```
grosbouddha: IF we want to do a11y properly, we will need a aria-invalid="true" on the textfield input
It might be out of scope for this PR but having a TODO/issue would be nice.

src/phoneField/index.tsx
```diff
@@ -21,6 +21,8 @@ interface PhoneFieldOnChangeParameters {
     completePhoneNumber: string
   }
 }
+
+type errorField = string | JSX.Element
```
grosbouddha: Could it be string only ? Do we need some crazy error message customization?

src/phoneField/index.tsx
PerrineBoissieres: Aria labels don't impact focus, it just provides additional informations when the user focus/interacts with the item. I'm more concerned about putting it on a `<div />`, since it can't be interacted with, I'm afraid the aria attribute won't have any effect =/

src/phoneField/index.tsx
grosbouddha: I'm ok with both but we already use Boolean below in the same file, being consistent in the same file would be nice. But it's just a nit, I don't care too much too.

src/phoneField/index.tsx
nickoola: So I created a dedicated ticket to check/fix A11y
>> BBCSPA-3355

PR #255, opened 2019-06-12
Handle 5+ passengers display in TripCard

mcellier:


src/tripCard/index.tsx
```diff
@@ -45,6 +55,18 @@ export interface TripCardProps {
   title?: string
 }
 
+const renderPassenger = (passenger: User) => (
+  <li className="kirk-tripCard-avatar" key={passenger.firstName}>
```
PerrineBoissieres: I'd use something else as key, just to avoid the case where two passengers have the same first name.

mcellier: I'll update the key to `${passenger.firstName}-${passenger.avatarUrl}`

PR #264, opened 2019-06-18
Removed large prop and added size prop with preset values instead

wakooka:


src/_utils/branding.ts
```diff
@@ -96,7 +96,12 @@ export const delay = {
 export const componentSizes = {
   timeWidth: '64px',
   buttonIconSize: '48px',
-  modalMaxWidth: '662px',
+}
+
+export const modalSize = {
```
mcellier: where are they coming from

PR #268, opened 2019-06-20
Icons refactor

mcellier:
Before:
screencapture-blablacar-github-io-ui-library-iframe-html-2019-06-21-10_46_48

After:
screencapture-localhost-63043-iframe-html-2019-06-21-10_46_15


src/_utils/icon/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+import style from 'icon/style'
+import { color } from '_utils/branding'
+
+export interface IconProps extends Icon {
+  readonly children: JSX.Element
+  readonly viewBox?: string
+}
+
+const BaseIcon = ({
+  className = '',
```
mcellier: I probably don't need to declare default here as they're set in `defaultProps`

mcellier:
Some icons:

Kapture 2019-06-21 at 10 49 26
Kapture 2019-06-21 at 10 50 37
Kapture 2019-06-21 at 10 51 23

I checked other components that use icons too (Loader, ItemRadio...)

PR #269, opened 2019-06-21
Add EmptyState component

mcellier:
New `EmptyState` component. To be used when empty search results, empty rides lists, ...

Capture d’écran 2019-06-21 à 11 55 13


mcellier:
Note: I'm using object-fit to keep image ratio, but it won't work on IE11. https://caniuse.com/#feat=object-fit

src/emptyState/specifications/emptystate.md
```diff
@@ -0,0 +1,33 @@
...
+
+> Illustrations
+
+Illustrations should have the max possible size (full width) and keep the W:H ratio when scalling WHILE not exceeding 1/3 of the screen height.
+
+> Content
+
+DO
+- Write two sentences.
+-  The first sentence refers to the missing element.
```
nickoola: White space

src/emptyState/specifications/emptystate.md
```diff
@@ -0,0 +1,33 @@
...
+- The second one guides the member to the action she can make.
+
+DO NOT
+- Excuse yourself.
```
nickoola: Is it necessary ? ;-)

mcellier: It's in the spec ;)

Tuxosaurus:
> Note: I'm using object-fit to keep image ratio, but it won't work on IE11. https://caniuse.com/#feat=object-fit

I'm not very comfortable with this, but I won't block this PR given that we would use it in a limited width layout (not fullscreen width I mean)

src/emptyState/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+  readonly className?: Classcat.Class
+  readonly image: string
+  readonly text: string
+  readonly button?: JSX.Element
+}
+
+const EmptyState = ({ className, image, text, button}: EmptyState) => (
+  <div className={cc(['kirk-empty-state', className])}>
+    <img src={image} alt="" />
+    <Title>{text}</Title>
```
wakooka: I would put `text` prop as a `children`, that's what we normally do in our component: main test is the `children` prop, wdyt?

src/emptyState/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+    {button}
```
wakooka: `button` prop, but this is not enforced I could pass a Modal in there and it will still render.

src/emptyState/index.tsx
mcellier: I thought about that part a bit: children is too permissive with what we can give it, allowing nonsenses in the `Title` component.
But... as the button can be whatever we want, i'll be more open to use it as children.

wakooka: Ok my bad, that's not what we do anymore in components apparently, checked with Item(s) components.

PR #276, opened 2019-06-25
Add pattern and inputMode as props

nickoola:


src/textField/index.tsx
```diff
@@ -46,6 +46,8 @@ export interface TextFieldProps extends CommonFormFields {
   label?: string
   buttonTitle?: string
   focus?: boolean
+  inputMode?: string
```
mcellier: You should be a bit more restrictive on this field, as we do for types. 
https://developer.mozilla.org/fr/docs/Web/HTML/Attributs_universels/inputmode

src/textField/story.tsx
```diff
@@ -13,7 +13,7 @@ import primaryDoc from '../button/specifications/primary.md'
 const stories = storiesOf('TextField', module)
 stories.addDecorator(withKnobs)
 
-const inputTypes = ['text', 'email', 'number', 'password']
+const inputTypes = ['text', 'email', 'number', 'password', 'number', 'tel']
```
mcellier: Maybe we should export these as `TextField` statics (and same for inputMode)

src/textField/index.tsx
wakooka: and export `inputMode` as static like @mcellier was saying for `type`
https://github.com/blablacar/ui-library/pull/276/files#r297204113

src/textField/story.tsx
```diff
@@ -4,24 +4,23 @@ import { storiesOf } from '@storybook/react'
 import { action } from '@storybook/addon-actions'
 import { withKnobs, text, boolean, number, select } from '@storybook/addon-knobs'
 
-import TextField from 'textField'
+import TextField, { inputTypes, inputModes } from 'textField'
```
mcellier: You can also use them as class constants (ex: `TextField. INPUT_TYPES`) and not export the enums.

PR #284, opened 2019-06-28
Fix EmptyState margin & centering

mcellier:


wakooka:
Screenshot before/after?

mcellier:
> Screenshot before/after?

Not visible in Storybook, since we don't reset browsers' default. We should probably implement a reset css in the storybook

PR #286, opened 2019-06-28
Display itinerary as small if we don't provide time or subLabel 

mcellier:
Before:
Capture d’écran 2019-06-28 à 11 43 30

After:
Capture d’écran 2019-06-28 à 11 43 37

In both cases, we don't have the `small` property, but we don't have either time or subLabel in places. Especially used for `TripCard` component that don't give us full control over the itinerary.

### DIFF
--- a/src/itinerary/index.unit.tsx
+++ b/src/itinerary/index.unit.tsx
@@ -203,4 +203,34 @@ describe('Itinerary component', () => {
       .key()
     expect(key2).toBe('route-end-tours')
   })
+
+  describe('small', () => {
+    it('Should be displayed as small if required in props', () => {
+      const itinerary = shallow(<Itinerary places={places} small />)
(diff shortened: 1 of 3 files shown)


src/itinerary/index.unit.tsx
```diff
@@ -203,4 +203,34 @@ describe('Itinerary component', () => {
...
+      const itinerary = shallow(<Itinerary places={places} small />)
+      expect(itinerary.hasClass('kirk-itinerary--small')).toBe(true)
+    })
+
+    it('Should not be displayed as small if not in props and time or sublabel exists', () => {
+      const itinerary = shallow(<Itinerary places={places} />)
+      expect(itinerary.hasClass('kirk-itinerary--small')).toBe(false)
+    })
+
+    it('Should be displayed as small if no time or subLabel', () => {
```
Tip-Sy: nit: `nor subLabel` ;)

PR #288, opened 2019-07-05
Allow SelectField and PhoneField to focus on mounting

nickoola:


nickoola:
Capture d’écran 2019-07-08 à 10 35 58
Capture d’écran 2019-07-08 à 10 34 32
Capture d’écran 2019-07-08 à 10 33 31

nickoola:
> Few high level questions:
> 
> * you added a autoFocus on selectField but not on the phoneField. I suppose the autofocus was added to autofocus something inside the phonefield. But I'm not sure how you re gonna use that select param if it's not propagated from phonefield
> * autofocus only works on page load AFAIK but you can access pages with a phonefield after a client-side navigation (and then autofocus will not work). Why not rely on a programmatic focus behavior that will work all the time and will be needed anyway for SPA navigation ?
> * For Phonefield, do we plan to have a focus style when either the nested select or textinput is selected inside the phonefield ? ( there is none for the moment IIRC)

I finally add `focus` and `autoFocus` for both components.
And I added attached screens to display the render on mounting `focus` is setted to `true`.

src/phoneField/index.tsx
```diff
@@ -181,6 +186,20 @@ export default class PhoneField extends PureComponent<PhoneFieldProps, PhoneFiel
...
+    if (this.ref && this.props.focus) {
+      this.ref.current.focus()
+      this.setState({ hasFocus: true })
+    }
+  }
+
+  componentDidUpdate(prevProps: PhoneFieldProps) {
+    if (this.ref && prevProps.focus !== this.props.focus) {
+      this.ref.current.focus()
+      this.setState({ hasFocus: true })
```
wakooka: Remove this line since it will trigger the `onFocus`

src/phoneField/index.tsx
```diff
@@ -181,6 +186,20 @@ export default class PhoneField extends PureComponent<PhoneFieldProps, PhoneFiel
     return null
   }
 
+  componentDidMount() {
+    if (this.ref && this.props.focus) {
+      this.ref.current.focus()
+      this.setState({ hasFocus: true })
```
wakooka: Remove this line since it will trigger the `onFocus`

src/phoneField/index.tsx
nickoola: Done!

src/phoneField/index.tsx
nickoola: Done!

PR #292, opened 2019-07-11
Example of component migrated to styled-components

PerrineBoissieres:
### Full migration of a component: the steps

- Rename the `index.tsx` file using the component name (ex: title/Title.tsx), same for associated unit test file.
- Since the `className` prop is used by `styled-component` to pass generated class names, we need another prop to pass any custom class names the user of the library would like to apply to the component (prop name TBD).
- Create a new `index.ts` that now serves the styled version of the component (and different variants if any) alongside other things you want to publicly export.
- Remove the `style.ts` file and use `styled-component` to apply the styles to the component (in Title case, styles are declared in the `index.ts` file, but it can be in another file, as long as `index.ts` exports the styled component by default). Styles can mostly be copy-pasted, and `global` styles can safely be removed.
- Remove all static properties of a component (ex: `Button.STATUS`) and prefer exporting them as a separate variable.
- Update imports in other files/components.

### Index changes

The index file has been simplified a lot and is now more flexible in what you can export (no more repetition of components with import/export syntax, we now use `export ... from ...` syntax). The unit tests have been updated to be more scalable, while still checking that we export all the main components and icons that we create.
You now have the possibility to export more than just the main component (ex: `export { default as Button, ButtonStatus } from 'button'`), which will help exporting type declaration, enums and other configuration outside of the library.

### How to import and use

There aren't many changes in the way you import components in another project, as paths and names haven't change. What changes is if you use SSR, you would now need to use the `styled-component` API to get styles. Also, if you used static properties for some components, you would need to update your code to now use different variables (ex: `Button.STATUS` becomes a brand new variable `ButtonStatus` that you need to import). Don't forget the jest plugin if you have snapshots, so you do not get huge conflicts on classnames each time a component is updated.


Tuxosaurus:
package.json deleted?

PR #295, opened 2019-07-17
Tabs update

wakooka:
Tabs

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "react-dom": "16.8.6",
     "react-test-renderer": "16.8.6",
     "rimraf": "2.6.3",
-    "storybook-readme": "5.0.1",
+    "storybook-readme": "5.0.2",
     "tslint": "5.14.0",
     "tslint-config-airbnb": "5.11.1",
     "tslint-config-prettier": "1.18.0",
--- a/src/tabs/index.tsx
+++ b/src/tabs/index.tsx
@@ -1,54 +1,65 @@
-import React, {createRef, Fragment, PureComponent, Ref, RefObject} from 'react'
+import React, { createRef, Fragment, PureComponent, Ref, RefObject } from 'react'
 import style from './style'
-import cc from "classcat"
+import cc from 'classcat'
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -75,7 +75,7 @@
     "react-dom": "16.8.6",
     "react-test-renderer": "16.8.6",
     "rimraf": "2.6.3",
-    "storybook-readme": "5.0.1",
+    "storybook-readme": "5.0.2",
```
wakooka: Fixing this https://github.com/tuchk4/storybook-readme/issues/132

src/tabs/index.tsx
```diff
@@ -62,88 +72,125 @@ const generateTabPanelId = (tab: Tab):string => {
...
+                    id={tab.id}
+                    ref={this.state.tabIdToRefs.get(tab.id)}
+                    onClick={this.handleTabClicked}
+                    onKeyDown={this.handleTabKeyDown}
+                    className={cc(['kirk-tab'])}
+                  >
+                    {tab.label}
+                  </button>
+                  {tab.badgeContent && (
+                    <Badge ariaLabel={tab.badgeAriaLabel} className="kirk-tab-badge">
```
grosbouddha: Since the Tabs is using programmatic focus management on <button>s, this badge count will most likely never be announced for screen readers (Unless the screenreader user is using virtual cursor to access the badge content, most likely this will not happen).

We could put the badge inside the button and absolutely position it ? When the button get focused, the badge will be announced and maybe will could use aria labels to improve the announcements e.g.

<button>Messages <badge aria-label="3 new">3</badge></button> => "Messages 3 new" will be announced when button is focused.

wdyt ?

wakooka: Unfortunately we can't put the `Badge` inside the button because `<button>` does not accept block element, but what we could do is to enrich the button title attribute, and concatenate button label with `badgeAriaLabel`. What do you think?

grosbouddha: ^ sgtm or we could hav

PR #302, opened 2019-07-19
Add missing icons for messaging notifications

grosbouddha:
Add new `CreditCardIcon`, `ExclamationIcon`, `GiftIcon`, `idCheckIcon`

Tested locally in the story book (changing the size and color of each icon).


src/icon/exclamationIcon.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react'
+import BaseIcon from '_utils/icon'
+
+// tslint:disable:max-line-length
+export const ExclamationIcon = (props: Icon) => (
```
Tuxosaurus: Maybe the Exclamation icon "within a circle" can be challenged since we have the Warning icon which is an Exclamation within a triangle

Tuxosaurus: In fact I remember I asked a designer about that a few months ago for a different feature and they agreed.

grosbouddha: If we challenge that, we will have to update ios/android for consistency.
I would prefer to keep it simple for now and avoid UX/mobile frictions.

Tuxosaurus: Ok

PR #303, opened 2019-07-22
update tabs (isWrapped, extend click area, Badge position)

Tuxosaurus:
No visual change but allows for simpler customisation when used.
Screen Shot 2019-07-22 at 18 38 13

### DIFF
--- a/src/tabs/index.tsx
+++ b/src/tabs/index.tsx
@@ -1,4 +1,4 @@
-import React, { createRef, Fragment, PureComponent, Ref, RefObject } from 'react'
+import React, { createRef, PureComponent, RefObject } from 'react'
 import style from './style'
 import cc from 'classcat'
 import Badge from 'badge'
@@ -22,6 +22,8 @@ interface TabsProps {
   readonly onChange?: Function
   readonly status?: TabStatus
   readonly className?: Classcat.Class
+  readonly tablistWrapperClassName?: Classcat.Class
+  readonly isWrapped?: boolean
 }
 
 interface TabsState {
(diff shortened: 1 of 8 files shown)


src/tabs/index.tsx
```diff
@@ -22,6 +22,7 @@ interface TabsProps {
   readonly onChange?: Function
   readonly status?: TabStatus
   readonly className?: Classcat.Class
+  readonly isWrapped?: boolean
```
Tuxosaurus: I admit I'm not a big fan of my naming here. If you have other suggestions...

src/tabs/style.ts
```diff
@@ -1,11 +1,24 @@
...
 const bottomBorderWidth = '2px'
 const bottomBorderOffset = '1px'
-const badgeOffset = '2px'
 
 export default css`
+  .kirk-tabs-wrapper {
+    position: relative;
+  }
+
+  .kirk-tabs-wrapper::before {
```
wakooka: Can't we use a normal border now that we this extra class? Or maybe we need an extra element again:
Main container could extend to 100%. And you could have a inner container with a max-width just for placing the tabs which would be the current `.wrapped` class

wakooka: That way we make sure we don't break the layout if tabs need to be in a column for example

Tuxosaurus: Alright, will try that way.

Tuxosaurus: Done, also fixed the selected tab border issue in scrollable mode so now borders look the same as in fixed.
Screen Shot 2019-07-23 at 16 34 55

PR #308, opened 2019-07-29
Migrate Avatar and badge to styled-component

pith:
Avatar
image

Badge
image


src/avatar/Avatar.tsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react'
+import cc from 'classcat'
+
+import prefix from '_utils'
+import { color } from '_utils/branding'
+import Badge from 'badge'
+import CheckIcon from 'icon/checkIcon'
+
+export interface AvatarInterface {
+  readonly className?: Classcat.Class
```
PerrineBoissieres: Don't forget the `extraClassName` attribute ;)

src/badge/Badge.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import cc from 'classcat'
+import isEmpty from 'lodash.isempty'
+
+interface BadgeProps {
+  readonly className?: Classcat.Class
```
PerrineBoissieres: Don't forget the `extraClassName` attribute ;)

src/messagingSummaryItem/__snapshots__/index.unit.tsx.snap
```diff
@@ -42,11 +42,10 @@ exports[`Should render properly with read messages 1`] = `
     className="jsx-1542968643 kirk-item-rightAddon"
   >
     <div
-      className="jsx-1920606029 kirk-avatar--image kirk-avatar"
+      className="kirk-avatar--image sc-bwzfXH fylziy kirk-avatar"
```
PerrineBoissieres: nit: Add the jest-styled-components library to avoid having the generated class names in the snapshots

Bidjii:
Yo
Quick question here. I'm also using `styled-components` in my component lib and I was wondering: Why don't you use props directly into your CSS, instead of having modifier classes (for example) in your styled components?

PerrineBoissieres:
> Yo
> Quick question here. I'm also using `styled-components` in my component lib and I was wondering: Why don't you use props directly into your CSS, instead of having modifier classes (for example) in your styled components?

This is part of the improvements we would do later, for now we're just focused on quickly switching from a library to another, with as few impacts as possible. We have quite a big amount of components, and we do not want this basic library switch to drag on for too long. There are a lot of improvements we can do thanks to styled-components, but this is probably a part 2 that we can do progressively when we have time :)

Bidjii:
👌

src/messagingSummaryItem/__snapshots__/index.unit.tsx.snap
pith: Done, I pushed it in a separate commit. I applied it in the Jest config so it should be global.

PR #309, opened 2019-07-29
Proximity migration

Tuxosaurus:
Screen Shot 2019-07-29 at 16 41 38


src/proximity/index.tsx
```diff
@@ -1,63 +1,21 @@
-import React from 'react'
```
pith: You can rename this file `Proximity.unit.tsx`

package.json
```diff
@@ -87,6 +87,7 @@
     "country-telephone-data": "0.6.0",
     "exenv": "1.2.2",
     "focus-trap": "5.0.0",
+    "jest-styled-components": "6.3.3",
```
pith: You can rebase on the `styled-components` branche. I already imported it and setup the config

src/proximity/index.tsx
Tuxosaurus: hum, this is the new index file. The unit test file has already been renamed to `Proximity.unit.tsx`

Tuxosaurus: @pith is it OK?

pith: Yes, sorry stupid comment.

PR #311, opened 2019-07-29
Migrate Caption to styled-component

pith:
image


pith:
> Do not target master 🙏

Sorry

PR #313, opened 2019-07-29
Migrate _utils/Item to Styled Components

PerrineBoissieres:
A few snapshots updates ^^"

But it works, no regression on Item components :)


package.json
```diff
@@ -69,6 +69,7 @@
     "gh-pages": "2.0.1",
     "husky": "1.3.1",
     "jest": "24.8.0",
+    "jest-styled-components": "6.3.3",
```
pith: 🤔 I already committed it normally

src/_utils/item/Item.unit.tsx
```diff
@@ -1,8 +1,10 @@
 import React from 'react'
 import renderer from 'react-test-renderer'
-import { shallow , mount } from 'enzyme'
+import 'jest-styled-components'
```
pith: It's configured globally, so you don't need to import it here

src/itemAction/index.unit.tsx
```diff
@@ -54,26 +54,26 @@ describe('ItemAction', () => {
 
   describe('status', () => {
     it('Should not render Loader when status is not defined', () => {
-      const wrapper = shallow(<ItemAction>...</ItemAction>).dive()
+      const wrapper = mount(<ItemAction>...</ItemAction>)
```
pith: Why `mount` here ?

src/toggleButton/index.tsx
```diff
@@ -1,14 +1,16 @@
 import React, { PureComponent } from 'react'
 import cc from 'classcat'
 
-import Item, { ItemProps } from '_utils/item/index'
+import Item from '_utils/item/index'
+import { ItemProps } from '_utils/item/Item'
```
pith: nit: you also export it on the index

src/itemAction/index.unit.tsx
PerrineBoissieres: We need to mount to access inside the `Item` from a parent component (before we used a `dive()` but since `Item` is wrapped by a hoc, we cannot do this anymore). I'm not usually a fan of diving into children, but for now it fixes the test.

Would be best to update the tests a bit once we migrate those components, to remove tests that are checking inside the children components, and thus testing the child behavior (which should already be done by `Item` itself).

src/toggleButton/index.tsx
PerrineBoissieres: It's not for use outside of the library, so we should keep it internally. Also I ran into a weird issue with typescript when trying to import/export interfaces 🤔

package.json
PerrineBoissieres: It's not up to date, on it ^^

pith: Can you remove my previous `jest-styled-components` import in the deps ? Thanks :)

src/_utils/item/Item.tsx
```diff
@@ -0,0 +1,172 @@
...
+  readonly onClick?: (event: React.MouseEvent<HTMLElement>) => void
+  readonly onBlur?: (event: React.FocusEventHandler<HTMLElement>) => void
+  readonly onFocus?: (event: React.FocusEventHandler<HTMLElement>) => void
+  readonly onMouseDown?: (event: React.MouseEvent<HTMLElement>) => void
+}
+
+const Item = ({
+  chevron,
+  className,
+  extraClassName,
```
pith: You can now remove `extraClassName ` ;)

src/_utils/item/index.tsx
```diff
@@ -1,171 +1,83 @@
...
-        <div className="kirk-item-rightAddon">
-          {<ChevronIcon iconColor={!isClickable ? color.fadedText : color.secondaryText} />}
-        </div>
-      )}
-      <style jsx>{style}</style>
-    </Tag>
-  )
-}
+  /* Text areas */
+  .kirk-item-leftText {
```
pith: no `&` prefix ?

src/_utils/item/Item.tsx
PerrineBoissieres: Done :)

src/_utils/item/index.tsx
PerrineBoissieres: Fixed :)

(they aren't necessary for children, but it's more consistent to have it everywhere)

package.json
PerrineBoissieres: Done

PR #314, opened 2019-07-29
[CSS in JS] BaseIcon in utils/icon

grosbouddha:


grosbouddha:
Tested locally: loader, all icons, a few random icons

src/_utils/icon/index.tsx
```diff
@@ -1,63 +1,28 @@
...
+`
 
-BaseIcon.defaultProps = {
-  className: '',
-  iconColor: color.icon,
-  size: 24,
-  title: '',
-  badgeAriaLabel: '',
-  badgeContent: '',
+StyledBaseIcon.defaultProps = {
```
PerrineBoissieres: Keep the defaultProps on the BaseIcon component, no need add them on the styled one.

grosbouddha: Done, PTAL

src/typings/icons.d.ts
```diff
@@ -1,6 +1,7 @@
 declare interface Icon {
   readonly size?: number | string
   readonly className?: Classcat.Class
+  readonly extraClassName?: Classcat.Class
```
Tuxosaurus: to be removed after the newest discoveries

PR #316, opened 2019-07-29
Textfield component

wakooka:


src/autoComplete/index.unit.tsx
```diff
@@ -343,18 +343,18 @@ describe('AutoComplete', () => {
 
   describe('#focus', () => {
     it('Can add focus to TextField', () => {
-      const wrapper = shallow(<AutoComplete {...defaultProps} focus />)
+      const wrapper = mount(<AutoComplete {...defaultProps} focus />)
```
pith: You need to update the `AutoComplete` import that is why your test fails

src/autoComplete/index.unit.tsx
```diff
@@ -343,18 +343,18 @@ describe('AutoComplete', () => {
...
       expect(wrapper.find('TextField').prop('focus')).toBe(true)
     })
     it('Should not add border focus on Textfield', () => {
-      const wrapper = shallow(<AutoComplete {...defaultProps} />)
+      const wrapper = mount(<AutoComplete {...defaultProps} />)
```
pith: Why `mount` ?

src/textField/Textfield.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+type errorField = string | JSX.Element
+
+export interface TextFieldProps extends CommonFormFields {
+  isTextArea?: boolean
+  defaultValue?: string
+  labelledBy?: string
+  onChange?: (obj: OnChangeParameters) => void
+  onClear?: () => void
+  className?: Classcat.Class
```
pith: You forgot to add the `extraClassName`.

src/autoComplete/index.unit.tsx
wakooka: I can't, it's imported from the component Autocomplete itself, it's not imported in the test

src/autoComplete/index.unit.tsx
wakooka: Because wrapper.find('TextField') can only work with mount since now it's wrapped in the Styled HOC

Tuxosaurus:
Rename files with "Textfield" to "TextField"

src/textField/Textfield.tsx
Tuxosaurus: Not needed anymore :)

wakooka: Removed :)

wakooka:
PTAL

PR #317, opened 2019-07-29
migrate Itinerary to styled-components

Tuxosaurus:
Screen Shot 2019-07-29 at 17 03 09
Screen Shot 2019-07-29 at 17 02 51
Screen Shot 2019-07-29 at 17 03 01


src/itinerary/index.tsx
```diff
@@ -1,150 +1,196 @@
...
-export default Itinerary
+import styled from 'styled-components'
+import { color, space, componentSizes } from '_utils/branding'
+
+import Itinerary from './Itinerary'
+
+const distanceFromHeight = '40px'
+
+const StyledItinerary = styled(Itinerary)`
+  li {
```
pith: no `&` ?

Tuxosaurus: `&` is on the `<ul>` but the base class is never used directly, only with modifiers (`--small`)

Tuxosaurus:
I replaced the only snapshot test, thus removing the file

PR #321, opened 2019-07-29
Convert SelectField to CSS-in-JS

grosbouddha:


src/selectField/SelectField.unit.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import { mount } from 'enzyme'
 import renderer from 'react-test-renderer'
 
-import SelectField from 'selectField'
+import SelectField from './SelectField'
```
pith: To test also the CSS with your snapshot you can use the following import for the snapshot test

`import StyledSelectField from './index'`

The current import is fine for the other tests.

grosbouddha: Done, thanks!

PR #330, opened 2019-07-30
Toggle button styled components

wakooka:
Capture d’écran 2019-07-30 à 16 14 06


src/toggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,66 @@
...
+}
+
+export default class ToggleButton extends PureComponent<ToggleButtonProps> {
+  static defaultProps: Partial<ToggleButtonProps> = {
+    status: ToggleButtonStatus.DEFAULT,
+    checked: false,
+    disabled: false,
+  }
+
+  static STATUS = ToggleButtonStatus
```
PerrineBoissieres: Not need anymore, remove it ;)

src/_utils/item/index.tsx
```diff
@@ -79,5 +79,5 @@ const StyledItem = styled(Item)`
   }
 `
 
-export { ItemStatus } from './Item'
+export { ItemStatus, ItemProps } from './Item'
```
PerrineBoissieres: Unless you intent to export them outside of the lib, don't expose the props in this file. Also Typescript raises a warning why trying to export interfaces this way, dunno why =/

PR #337, opened 2019-07-30
[BBCSPA-3528] Update Node version to 10.15.0

Tip-Sy:
Node 10.x.x uses a new implementation for date formatting, so now we're forcing the 24-hour format by default, and remove the unused locale props.
When using this component, you still have the possibility to pass your own "renderTime" method with your own implementation (depending on the locale or not)


wakooka (approved):

grosbouddha (approved):

pith (changes requested):
Also change the version in `travis.yml`

pith (approved):

src/timePicker/index.tsx
```diff
@@ -41,12 +41,15 @@ export const getTodayDate = () => {
 
 /**
  * Format given dateTime with `Date#toLocaleTimeString`.
+ * Node 10.x.x uses a new implementation for date formatting,
+ * so we don't rely on it and check the locale for 24-hour format.
```
grosbouddha: Please add a comment to TimePickerProps#renderTime:
'Custom renderTime. If not provided a default one without proper i18n will be used.'

grosbouddha (approved):

PR #348, opened 2019-07-31
Fix ItemChoice static const export

wakooka:
Was going to migrate `Menu` component, but there was no style on this component. However I notice some issues with the export of STATIC consts in `ItemChoice`


pith:
You need to rebase

PR #353, opened 2019-07-31
PhoneField styled-component migration

wakooka:
image


src/phoneField/PhoneField.tsx
```diff
@@ -0,0 +1,266 @@
...
+  }
+}
+
+type errorField = string | JSX.Element
+interface PhoneFieldProps {
+  readonly name: string
+  readonly onChange: (obj: PhoneFieldOnChangeParameters) => void
+  readonly id?: string
+  readonly className?: Classcat.Class
+  readonly extraClassName?: Classcat.Class
```
grosbouddha: extraClassName makes me think that is will be on the same element as className (i.e. the root).
Can we use a naming that tries to identify where the class is added ?
In that case, maybe innerWrapperClassName ?

src/autoComplete/AutoCompleteList.unit.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import { shallow, mount } from 'enzyme'
 
-import AutoCompleteList from './AutoCompleteList'
+import AutoCompleteList from './autoCompleteList'
```
pith: It brokes the test
```
Test suite failed to run
    Cannot find module './autoCompleteList' from 'AutoCompleteList.unit.tsx'
      1 | import React from 'react'
      2 | import { shallow, mount } from 'enzyme'
    > 3 | 
        | ^
      4 | import AutoCompleteList from './autoCompleteList'
      5 | import ItemChoice, { ItemChoiceStatus } from 'itemChoice'
      6 | 
```

wakooka: Oops bad conflict resolving.

src/phoneField/PhoneField.tsx
wakooka: Done PTAL

src/autoComplete/AutoCompleteList.unit.tsx
wakooka: Fixed

PR #357, opened 2019-07-31
Fix case issue with ItemRadio

PerrineBoissieres:
Wrong case in ItemsList import ^^"


src/phoneField/story.tsx
```diff
@@ -34,7 +34,7 @@ stories.add(
       <PhoneField
         id={text('id', 'phoneFieldId')}
         name={text('name', 'phoneField')}
-        extraClassName={text('className', 'additionalClass')}
+        innerWrapperClassName={text('className', 'additionalClass')}
```
wakooka: already fixed if you rebase :)

PR #370, opened 2019-08-16
Reinforce style cibling to avoid children component to override it

nickoola:


Tuxosaurus:
You should explain why it is a problem.

Nit: "cibling" => specificity

nickoola:
Fix tested on Kairos
Capture d’écran 2019-08-19 à 10 43 58

PR #372, opened 2019-08-19
Refactor out textarea from textfield

grosbouddha:
- Split textfield into texfield (i.e. inputs) and textarea dedicated components
- add new options for adding a button icon and button callback to the textarea. It will be positioned in the bottom-right corner. Some padding is added to avoid overlap of the text and button.
- add new fitContent attribute that will allow to fit the height of the textarea to its content dynamically.


src/textarea/Textarea.unit.tsx
```diff
@@ -0,0 +1,152 @@
...
+  expect(textarea.prop('value')).toBe('blabla')
+})
+
+it('Should have a valid button with title', () => {
+  const mockOnButtonClick = jest.fn()
+  const props = {
+    name: 'test',
+    defaultValue: 'blabla',
+    buttonTitle: 'button title',
+    buttonIcon: <EyeIcon />,
```
Tip-Sy: nit: you could also add an "expect" for the icon , it's not tested elsewhere

src/textarea/Textarea.tsx
```diff
@@ -0,0 +1,250 @@
...
+  buttonRef: HTMLButtonElement = null
+  textareaRef: RefObject<HTMLTextAreaElement> = React.createRef()
+
+  componentDidMount() {
+    if (this.textarea && this.props.focus) {
+      this.textarea.focus()
+    }
+  }
+
+  componentWillReceiveProps({ defaultValue, focus }: TextAreaProps) {
```
pith: Can you add a todo here, to refactor this method ? `componentWillReceiveProps` is now deprecated.

See: https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops

Also this method is probably buggy: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#anti-pattern-erasing-state-when-props-change

src/textarea/Textarea.tsx
```diff
@@ -0,0 +1,250 @@
...
+
+  onChange = () => {
+    this.props.onChange({
+      name: this.props.name,
+      value: this.state.value,
+    })
+
+    // Fit to content.
+    if (this.props.fitContent && this.textareaRef && this.textareaRef.current) {
+      this.textareaRef.current.style.height = '0'
```
pith: Why do you need to reset the height before updating it ?

src/textarea/Textarea.tsx
```diff
@@ -0,0 +1,250 @@
...
+      const verticalPadding = 16
+      this.textareaRef.current.style.height =
+          `${this.textareaRef.current.scrollHeight - 2 * verticalPadding}px`
```
pith: Nice trick :)

src/textarea/Textarea.tsx
```diff
@@ -0,0 +1,250 @@
...
+      onChange: this.onTextAreaChange,
+      ref: this.ref,
+    }
+
+    if (error) {
+      attributes['aria-invalid'] = 'true'
+    }
+
+    const textareaFieldClassNames = ['kirk-textarea-element']
+    const shouldDisplayButton = buttonIcon && onButtonClick && !isEmpty(buttonTitle)
```
pith: `!isEmpty(buttonTitle)` => that's fine to enforce a11y, but this could be documented in the specs otherwise it's a bit hidden.

src/textarea/Textarea.tsx
```diff
@@ -0,0 +1,250 @@
...
+      textareaFieldClassNames.push('kirk-textarea-hasButton')
+    }
+    if (fitContent) {
+      textareaFieldClassNames.push('kirk-textarea-fitContent')
+    }
+    attributes.className = cc(textareaFieldClassNames)
+
+    return (
+      <div className={cc(['kirk-textarea', prefix({ error: !!error, disabled }), className])}>
+        {label && <label htmlFor={id}>{label}</label>}
```
pith: Do we want to have a label ? I don't see it in the specs.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Unreleased
 
+- **[BREAKING CHANGE]** Extract `Textarea` from `TextField`
+- **[NEW]** Add ability to add a bottom/right button icon to `Textarea`
+- **[NEW]** Add fitToContent feature to `Textarea`
```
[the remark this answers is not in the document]
grosbouddha: Added documentation about fitContent into the changelog and the Textarea props.

src/textarea/Textarea.tsx
grosbouddha: Done. BBC-498 created too.

src/textarea/Textarea.tsx
grosbouddha: Then the scrollHeight is the height of the whole (overflowing) content.

src/textarea/Textarea.tsx
grosbouddha: It was already there so I kept it.
Do we have specs to have a label for teatfields ?

src/textarea/Textarea.unit.tsx
grosbouddha: I added a check for the icon in the button. Thanks!

src/textarea/Textarea.tsx
grosbouddha: Added doc

PR #384, opened 2019-09-02
Adapt Textarea height when changing programmatically its content in 'fitContent' mode

grosbouddha:
When used with fitContent mode, the Textarea is adapting its height as the user typs in the field. We need to update the height as well when the value of the field is updated programmatically using the defaultValue props.

TESTED=Locally in story book with programmatic and manual modifications of Textarea content.

### DIFF
--- a/src/icon/sendIcon.tsx
+++ b/src/icon/sendIcon.tsx
@@ -1,3 +1,4 @@
+// tslint:disable:max-line-length
 import React from 'react'
 import BaseIcon from '_utils/icon'
 import { BaseIconDefaultProps } from '_utils/icon/BaseIcon'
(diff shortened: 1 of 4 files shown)


src/icon/sendIcon.tsx
```diff
@@ -1,3 +1,4 @@
+// tslint:disable:max-line-length
```
Tip-Sy: Oh thanks

src/textarea/Textarea.tsx
```diff
@@ -100,6 +100,11 @@ export default class Textarea extends PureComponent<TextAreaProps, TextAreaState
       this.setState({
         value: defaultValue,
       })
+      // Update height in next tick to give a chance to the contained text to
+      // give a proper layout to the surrounding parents.
+      setTimeout(() => {
+        this.maybeAdaptHeightToContent()
```
Tip-Sy: Sounds hacky but seems to work 😛

PR #395, opened 2019-09-06
Set default horizontal paddings for Item to 24px

grosbouddha:


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **[BREAKING CHANGE]** Add default horizontal padding in base `Item`.
```
pith: It reminds me a PR that I did and rollback ;)

grosbouddha: ahah, yep it's a dead PR. I realized after that it was much more complicated than just that.
I will decline it now.

PR #396, opened 2019-09-10
add highlight animation to Tabs

Tuxosaurus:
kirk_tabs_highlight


src/tabs/Tabs.tsx
```diff
@@ -119,11 +122,39 @@ export class Tabs extends PureComponent<TabsProps, TabsState> {
...
+
+    // tablist parent needed to get the height of the whole component
+    // (tabs can have different heights)
+    const parentTablist = currentTab.closest('.kirk-tablist')
+    const parentBounds = parentTablist.getBoundingClientRect()
+
+    // adapt to current tab width
+    this.highlightRef.current.style.width = `${tabBounds.width}px`
+    // top position, 1px = wrapper gray border
+    this.highlightRef.current.style.top = `${parentBounds.height - 1}px`
```
wakooka: Can you get that `1px` dynimacally? If not, add a const above.

Tuxosaurus: I changed it (changed the parent element I used)

src/tabs/Tabs.tsx
```diff
@@ -119,11 +122,39 @@ export class Tabs extends PureComponent<TabsProps, TabsState> {
     }
   }
 
+  moveHighlight = (activeTabId: string) => {
+    const currentTab = this.state.tabIdToRefs.get(activeTabId).current
+    const tabBounds = currentTab.getBoundingClientRect()
+
+    // tablist-wrapper parent needed to get the height of the whole component
+    // because tabs can have different heights
+    const parentTabWrapper: any = currentTab.closest('.kirk-tablist-wrapper')
```
Tuxosaurus: `any` because if I type it as what it is (HTMLDivElement) TypeScript won't let me commit:
```
error TS2740: Type 'Element' is missing the following properties from type 'HTMLDivElement': align, accessKey, accessKeyLabel, autocapitalize, and 110 more.

131     const parentTabWrapper: HTMLDivElement = currentTab.closest('.kirk-tablist-wrapper')
```

If I don't provide type, my linter complains that I can't use `offsetHeight` from that element...

grosbouddha: Use HTMLElement instead of HTMLDivElement, that should work.
This is static type-checking, the fact that it is a div in particular is runtime and can't be enforced at compile time.

Tuxosaurus: Same thing with HTMElement

Tuxosaurus: Alright, left a comment in the code about why this happens.
".closest" returns a base Element that doesn't have offsetHeight so we cast it to HTMLElement using the "as" TS syntax

PR #401, opened 2019-09-20
Bbc 693/tabs icons

Tuxosaurus:
Add the possibility to use icons within Tabs.
You can chose to also display the text (label) along with the icon (desktop context). This needed to review the alignment of the different elements withing a tab.

Also added 2 icons that are specific to tabs.
Screen Shot 2019-09-20 at 13 45 05
Screen Shot 2019-09-20 at 13 45 52

### DIFF
--- a/src/icon/index.tsx
+++ b/src/icon/index.tsx
@@ -56,6 +56,8 @@ import ShareIcon from './shareIcon'
 import SmokeIcon from './smokeIcon'
 import SpeakIcon from './speakIcon'
 import StarIcon from './starIcon'
+import BusIcon from './busIcon'
+import CarpoolIcon from './carpoolIcon'
 import TicketIcon from './ticketIcon'
 import TrashIcon from './trashIcon'
 import TwitterIcon from './twitterIcon'
@@ -123,6 +125,8 @@ export {
   SmokeIcon,
   SpeakIcon,
   StarIcon,
+  BusIcon,
+  CarpoolIcon,
   TicketIcon,
   TrashIcon,
   TwitterIcon,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # Unreleased
 
 - **[FIX]** Fix font-size for `MessagingSummaryItem` sublabel.
-- [...]
+- **[UPDATE]** Add possibility to use icons in `Tabs`
+- **[NEW]** Add `BusIcon` and `CarpoolIcon`
 
 # v11.2.4 (16/09/2019)
 
--- a/src/icon/busIcon.tsx
+++ b/src/icon/busIcon.tsx
@@ -1,0 +1,36 @@
+// tslint:disable:max-line-length
+import React from 'react'
+import { color } from '_utils/branding'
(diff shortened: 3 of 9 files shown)


src/icon/index.tsx
```diff
@@ -56,6 +56,8 @@ import ShareIcon from './shareIcon'
 import SmokeIcon from './smokeIcon'
 import SpeakIcon from './speakIcon'
 import StarIcon from './starIcon'
+import TabBusIcon from './tabBusIcon'
```
mcellier: I don't think we should prefix them with `Tab`. The library is open source, so we shouldn't expect that these will always be used inside a tab

src/icon/tabBusIcon.tsx
```diff
@@ -0,0 +1,35 @@
...
+      <path
+        d="M.67 5h20c1 0 1.66 1.66 2.66 1.66h4v4.67a1 1 0 0 1-1 1H4.67L1 11.1a.5.5 0 0 1-.34-.47V5z"
+        stroke={props.iconColor}
+        fill={props.iconColor}
+      />
+      <path d="M4 7h7.33" stroke="#FFF" strokeLinecap="round" strokeLinejoin="round" />
+      <path
+        d="M4.67 5.67V.33M10 5V.33m5.33 6v-6m5.34 6v-6"
+        stroke={props.iconColor}
+        fill="#FFF"
```
mcellier: Remove hard coded color

src/icon/tabBusIcon.tsx
```diff
@@ -0,0 +1,35 @@
...
+        d="M.67 6.33V2C.67.92 1.33.33 2.33.33h21.02C26 .33 27.33 3.01 27.33 5v6.34a1 1 0 0 1-1 1H4.67l-3.32-1.11a1 1 0 0 1-.68-.95V6.33z"
+        stroke={props.iconColor}
+        fill="#FFF"
+      />
+      <path
+        d="M.67 5h20c1 0 1.66 1.66 2.66 1.66h4v4.67a1 1 0 0 1-1 1H4.67L1 11.1a.5.5 0 0 1-.34-.47V5z"
+        stroke={props.iconColor}
+        fill={props.iconColor}
+      />
+      <path d="M4 7h7.33" stroke="#FFF" strokeLinecap="round" strokeLinejoin="round" />
```
mcellier: Remove hard coded color

src/icon/index.tsx
Tuxosaurus: Agreed but I'm reluctant to name it Car and Bus, knowing they are not what we actually use for cars and buses... If you have any other suggestions

src/icon/tabBusIcon.tsx
Tuxosaurus: Done

src/icon/tabBusIcon.tsx
Tuxosaurus: Done

src/icon/index.tsx
Tuxosaurus: Alright I did remove `Tab`

PR #406, opened 2019-10-03
Create TheVoice component

grosbouddha:
Create TheVoice component based on incomplete pixar specs and current implementation in SPA.

Tested=Locally in storybook + a few unit tests

image


src/theVoice/specifications/theVoice.md
```diff
@@ -0,0 +1,32 @@
+
+## The Voice
+
+```js
+import { TheVoice } from '@blablacar/ui-library'
+```
+
+```html
+<TheVoice responsiveMode={ResponsiveMode.SMALL}>
+    This is the Voice!
```
mcellier: Noice!

pith: Voice!

src/_utils/branding.ts
```diff
@@ -167,6 +168,22 @@ export const color: Color = {
   tapHighlight: 'rgba(221, 221, 221, .4)', // gray, 40%
 }
 
+
+export enum ResponsiveMode {
+  SMALL = 'small',
+  LARGE = 'large',
+}
+
+export const computeResponsiveModeClass = (mode:ResponsiveMode) => {
```
pith: Why not having the CSS breakpoint in Kirk to avoid passing responsive flag via JavaScript ?

src/_utils/branding.ts
```diff
@@ -167,6 +168,22 @@ export const color: Color = {
...
+  switch(mode) {
+    case ResponsiveMode.LARGE:
+      return 'rm-large'
```
pith: nit: usually I map `rm` to `remove` so it took me few seconds to understand ;)

src/theVoice/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+const StyledTheVoice = styled(TheVoice)`
+  &.rm-small {
+    font-size: ${font.l.size};
+    margin: ${space.xl} 0;
+  }
+  
+  &.rm-large {
+    text-align: center;
+    font-size: ${font.xl.size};
```
pith: Not the same font-size ? On large screen ?

src/theVoice/index.tsx
```diff
@@ -0,0 +1,19 @@
+import styled from 'styled-components'
+import { space, font } from '_utils/branding'
+
+import TheVoice from './TheVoice'
+
+const StyledTheVoice = styled(TheVoice)`
+  &.rm-small {
```
pith: Is it also suppose to include the Kairos CSS for the Voice component ?

```
.theVoice {
  padding: var(--space-xl) 0;
  font-weight: 500;
}
@media (--large) {
  .theVoice {
    padding: 48px 0;
    text-align: center;
  }
}
```

src/_utils/branding.ts
grosbouddha: should I use full responsive-mode-large/small ?

src/theVoice/index.tsx
grosbouddha: Yep it's part of the specs, it get smaller for small viewports.
We might need to renormalize all our font sizes, none of Dawid's spec values match what we have in our branding config. For now I just use a smaller font.

src/theVoice/index.tsx
grosbouddha: This kairos CSS will go away when I integrate TheVoice in kairos.

src/_utils/branding.ts
grosbouddha: CSS media query breakpoints are related to the viewport, not the width of a particular layout element. To support future multi-column layouts, we need to make sure that the component is responsive to its containing layout column, not the viewport size.
But maybe, we don't care for now ? and I go for viewport based media queries that will break when we introduce multi-column desktop layouts ? That will simplify this PR a bit and the integration points in kairos.

We will have the same discussion about layout paddings/margins in future PRs (e.g next week probably).

pith: So you could have component with the responsive mode large and small on the same page depending on their container ?

src/_utils/branding.ts
pith: or `responsive-large` as you prefer

grosbouddha: +1, will do (If we keep these modes - cf the other discussion)

src/_utils/branding.ts
grosbouddha: yes.
As discussed offline, I will remvoe for now the responsive mode and add a responsive breakpoint in TheVoice. We will iterate on that when we have actual desktop specs.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - **[NEW]** Add `Disclaimer` component
+- **[NEW]** Add `TheVoice` widget.
```
pith: Also add break point

grosbouddha: Added !

PR #412, opened 2019-10-08
Add highlighted option to ItemRadio

mcellier:


src/itemRadio/ItemRadio.tsx
```diff
@@ -24,6 +24,7 @@ export interface ItemRadioProps {
   readonly status?: ItemRadioStatus
   readonly key?: string | number
   readonly chevron?: boolean
+  readonly highlighted?: boolean
```
Tuxosaurus: post-nit: I prefer boolean with verbs like "isHighlighted"

mcellier:
I used the prop that we already have in item.

Le mer. 9 oct. 2019 à 10:03, Cyrille Jouineau <notifications@github.com> a
écrit :

PR #417, opened 2019-10-08
Add missing display modes in Disclaimer

grosbouddha:
- Add deprecated rendering for button icon on the right side (used for SRP)
- Add non-caption mode for body text

### DIFF
--- a/src/disclaimer/Disclaimer.tsx
+++ b/src/disclaimer/Disclaimer.tsx
@@ -1,18 +1,41 @@
 import React from 'react'
 import Item from '_utils/item'
 import InfoIcon from 'icon/infoIcon'
+import QuestionIcon from 'icon/questionIcon'
+import Button, { ButtonStatus } from 'button'
 import { TextDisplayType } from 'text'
+import { color } from '_utils/branding'
+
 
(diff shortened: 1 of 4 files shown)


src/button/story.tsx
```diff
@@ -113,6 +113,7 @@ stories.add('icon + text', () => (
   </Button>
 ))
 
+
```
mcellier: Maybe add a `**[BREAKING CHANGE]**` for this one ;)

grosbouddha: :)

grosbouddha:
Changelog.md updated

src/disclaimer/Disclaimer.tsx
```diff
@@ -1,18 +1,41 @@
 import React from 'react'
 import Item from '_utils/item'
 import InfoIcon from 'icon/infoIcon'
+import QuestionIcon  from 'icon/questionIcon'
```
Tuxosaurus: there are 2 spaces before "from"

grosbouddha: Done

PR #422, opened 2019-10-18
Fix Itinerary/TripCard visual alignment and a11y

grosbouddha:
- Set time cell width to 48px (was 64px) + 8px of right padding (was 0)
- Fix a11y for title (it's now a SubHeader outside the list, it was a nested li/text)
- Fix departure/arrival city alignement with dot
- Fix multiple :before/:after elements being created due to non selective CSS rule (div/span)

Itinerary:
image
Small itinerary:
image

TripCard:
image

### DIFF
--- a/src/itinerary/Itinerary.tsx
+++ b/src/itinerary/Itinerary.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { Fragment } from 'react'
 import cc from 'classcat'
 import isEmpty from 'lodash.isempty'
 import isString from 'lodash.isstring'
@@ -6,6 +6,8 @@ import isString from 'lodash.isstring'
 import { color } from '_utils/branding'
 import Text, { TextDisplayType, TextTagType } from 'text'
 import ChevronIcon from 'icon/chevronIcon'
+import SubHeader from 'subHeader'
+import BlankSeparator from 'blankSeparator'
 
 interface ItineraryProps {
   readonly places: Place[]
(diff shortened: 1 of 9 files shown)


src/itinerary/Itinerary.tsx
```diff
@@ -16,8 +18,19 @@ interface ItineraryProps {
   readonly headline?: string
 }
 
+
 const isNonEmptyString = (str: string) => isString(str) && str.trim().length > 0
 
+const computeKeyFromPlace = (place: any) => {
```
pith: Why not using `Place` interface instead of `any` ?

src/itinerary/index.tsx
```diff
@@ -47,58 +44,64 @@ const StyledItinerary = styled(Itinerary)`
...
     border-radius: 50%;
-    top: calc(4px + ${space.m});
-    left: calc(${componentSizes.timeWidth} - 2px);
+    top: calc(4px + ${space.m} + 4px);
+    left: calc(${gutterPaddingStart} + ${componentSizes.timeWidth} - 2px);
+  }
+  
+  & .kirk-itinerary-location.kirk-itinerary--departure .kirk-itinerary-location-city::after,
+  & .kirk-itinerary-location.kirk-itinerary--arrival .kirk-itinerary-location-city::after {
+    top: calc(4px + ${space.m} + 4px);
```
pith: the 4px corresponds to what ?

grosbouddha: It's just ugly adjustments to align stuff. I could use some magic constants to make it feel it was engineered if you prefer. We will need a much deeper refactoring if we want meaningful constants. For now, I will at least merge the 2 4px into one 8px :)

grosbouddha:
FYI, I added some missing padding in the TripCard to align the avatar name with the itinerary city names. (I updated the screenshot in the desc)

PR #426, opened 2019-11-04
Add textColor prop to Stepper component

mcellier:
In publication, we need to change the text color of the stepper depending on the user's price.

Capture d’écran 2019-11-04 à 12 28 26

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 [...]
+- **[NEW]** Export new colors
 - **[FIX]** Restore focus and click when Modal component unmounts.
 
 # v13.0.0 (22/10/2019)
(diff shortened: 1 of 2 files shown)


PerrineBoissieres:
I'd rather have an enum of the different colors available than allowing any hex code.

src/stepper/Stepper.tsx
```diff
@@ -34,6 +35,7 @@ interface StepperState {
 // Support IE. Same value returned with Number.MAX_SAFE_INTEGER / Number.MIN_SAFE_INTEGER
 const defaultInteger = 2 ** 53 - 1
 const isTouchScreen = isTouchEventsAvailable()
+const cssColorRegex = /^#([0-9a-f]{3}|[0-9a-f]{6})$/i
```
Tuxosaurus: We may need a "sanitize" helper to take care of this kind of verification. For our use case it's fine but maybe others will want to use rgb(a) formats for instance.

mcellier: Removed

PR #430, opened 2019-11-06
Add outline back on buttons

pith:
image

### DIFF
--- a/src/button/index.tsx
+++ b/src/button/index.tsx
@@ -41,9 +41,7 @@ const StyledButton = styled(Button)`
   }
 
   &:hover,
-  &:focus,
   &:active {
-    text-decoration: none;
     outline: 0;
   }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
-[...]
+- **[FIX]** Restore outline on Button, Dropdown, HamburgerButton and ToggleButton
+  [...]
 
 # v13.0.3 (06/11/2019)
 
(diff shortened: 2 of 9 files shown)


src/button/index.tsx
```diff
@@ -41,10 +41,7 @@ const StyledButton = styled(Button)`
   }
 
   &:hover,
-  &:focus,
   &:active {
```
grosbouddha: Rule is empty, remove ?

src/button/index.tsx
```diff
@@ -41,10 +41,7 @@ const StyledButton = styled(Button)`
...
-    text-decoration: none;
```
grosbouddha: I would keep the text-decoration since i'm not sure why it's here in the first place.

src/button/index.tsx
```diff
@@ -41,10 +41,7 @@ const StyledButton = styled(Button)`
   }
 
   &:hover,
-  &:focus,
```
wakooka: Focus is also call onmousdown depending on browsers, which might results in a quick flash of focus state when clicking on button. So maybe it should apply only on keyboard navigation like we did in Kairos using focus-visible. 
https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible
wdyt?

src/button/index.tsx
pith: Oups boulette

src/button/index.tsx
nickoola: Yes same remark. And it's probably to avoid undeline on <a> tag.

src/button/index.tsx
pith: ✅ PTAL

wakooka: This won't work, because it means that we need to pass a `.focus-visible` class to each component.
We have this in Kairos:
```
/* Remove default focus outline when focus is not visible. */
body:not(.focus-visible) :focus {
  outline: none;
}
```
So depending on CSS specificity your previous commit might work sorry 🤦‍♂

PR #431, opened 2019-11-08
Overhaul Message kirk component

grosbouddha:
- Remove a bunch of unused property and not up-to-spec code
- Messages have now proper left/right alignment and can stack properly (author and date were removed)
- Add a new generic annotation field which is added below the Message instead of inside
- Added new story book cases for complex conversation cases
- Added support for very long message or long message without breaks

TESTED=Locally in story book + unit tests for basic structure

image

### DIFF
--- a/src/message/Message.tsx
+++ b/src/message/Message.tsx
@@ -1,40 +1,45 @@
-import React from 'react'
+import React, { Fragment } from 'react'
 import cc from 'classcat'
 
 import prefix from '_utils'
 
-import Caption from 'caption'
+import BlankSeparator, { BlankSeparatorSize } from "blankSeparator"
+import Text, { TextDisplayType } from "text"
 
 export interface MessageProps {
   readonly children: string
-  readonly isoDate?: string
   readonly date?: string
-  readonly secondaryLink?: string
-  readonly secondaryText?: string
   readonly active?: boolean
   readonly author?: string | JSX.Element
   readonly className?: Classcat.Class
+  readonly messageAnnotation?: string
 }
 
 const Message = ({
   active,
-  author,
   date,
-  isoDate,
-  secondaryLink,
-  secondaryText,
   children,
+  messageAnnotation,
   className,
 }: MessageProps) => (
-  <blockquote className={cc(['kirk-message', prefix({ active }), className])}>
-    {author && <cite>{author}</cite>}
-    <div className="kirk-label">
-      <p>{children}</p>
-      <Caption href={secondaryLink} secondaryText={secondaryText} isoDate={isoDate}>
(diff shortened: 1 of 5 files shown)


src/message/story.tsx
```diff
@@ -1,33 +1,72 @@
...
 stories.addDecorator(withKnobs)
 
+const wrapInContainer = (content: JSX.Element) => {
+  return <div style={{width: '600px', border: '1px solid lightgray'}}>
+    {content}
+  </div>
+}
+
+const veryLongMessage = "long message ".repeat(10)
+const veryLongMessageWithoutBreaks = "longmessage_".repeat(10)
```
Tip-Sy: Smart 😄

src/message/index.tsx
```diff
@@ -4,9 +4,35 @@ import { color, space, font } from '_utils/branding'
...
+  }
+  
+  & .kirk-message-content {
+    word-break: break-word;
+  }
+  
+  & .kirk-message-content,
+  & .kirk-message-annotation {
+    display: flex;
+    margin-left: 48px;
```
Tuxosaurus: 48px would be ${space.xxl}

src/message/Message.tsx
```diff
@@ -1,40 +1,45 @@
...
-  secondaryText,
   children,
+  messageAnnotation,
   className,
 }: MessageProps) => (
-  <blockquote className={cc(['kirk-message', prefix({ active }), className])}>
-    {author && <cite>{author}</cite>}
-    <div className="kirk-label">
-      <p>{children}</p>
-      <Caption href={secondaryLink} secondaryText={secondaryText} isoDate={isoDate}>
```
Tip-Sy: So I guess it's on purpose: the annotation can only be a text right?
But in fact we never needed the link did we?

grosbouddha: yes

src/message/index.tsx
grosbouddha: Done and I used the branding border radius too.

src/message/story.tsx
```diff
@@ -1,33 +1,72 @@
...
+stories.add('Conversation', () => (
+    wrapInContainer(
+        <Fragment>
+          <Message messageAnnotation="Delivered - 15:29" active={false}>Msg</Message>
```
pith: I find the `active` props not really clear, I suppose that it means "current user" and that it changes the bubble alignement but that's a guess 😉.

Otherwise the API feels nice to use 👍.

grosbouddha: active is from the previous API and is not really explicit, I agree.

PR #432, opened 2019-11-08
Handle TimePicker focus on mount

PerrineBoissieres:
In a similar fashion to TextField, we can now use the `focus` prop on the TimePicker to focus the select when the component is mounted.

Focused style (copied from TextField component):
Capture d’écran 2019-11-08 à 15 27 25

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-- **[NEW]** `focus` prop on `TimePicker` to focus the select on mount
+- **[NEW]** `focus` prop on `TimePicker`
 
 # v13.0.3 (06/11/2019)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-[...]
+- **[NEW]** `focus` prop on `TimePicker` to focus the select on mount
 
 # v13.0.3 (06/11/2019)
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Unreleased
 
-[...]
+- **[NEW]** `focus` prop on `TimePicker` to focus the select on mount
```
mcellier: Added `focus` prop to `TimePicker` should be sufficient

src/timePicker/TimePicker.tsx
```diff
@@ -2,18 +2,6 @@ import React, { PureComponent } from 'react'
 import cc from 'classcat'
 import prefix from '_utils'
 
-/**
- * Whether or not Date#toLocaleTimeString supports arguments `locales` et `options`.
- */
-const toLocaleTimeStringSupportsLocales = (() => {
```
mcellier: Not used?

src/timePicker/TimePicker.tsx
```diff
@@ -105,6 +96,12 @@ export default class TimePicker extends PureComponent<TimePickerProps, TimePicke
     this.props.onChange({ name, value })
   }
 
+  componentDidMount() {
```
mcellier: Do we have a specific style when focussed?

src/timePicker/TimePicker.tsx
PerrineBoissieres: Seems not, it was a raised warning by the linter.

src/timePicker/TimePicker.tsx
PerrineBoissieres: I'm adding it (I'll update the description)

PR #437, opened 2019-11-12
Stepper font-size auto-adapt magic

mcellier:
Breaking changes on the `Stepper` component:
- Added `display` property to handle 2 two display we have (SMALL & LARGE)
- Removed `buttonSize` property, as  the size depends on the display. 

Previously, the LARGE display was handled in kairos by adding a `text-xxl` class and customizing `buttonSize`.

Also, I added a "dirty" thing to handle large font size on small screens. We won't need that for countries with low numeric values/short currencies, but in Ukraine for instance, we have to display long texts:

stepper


src/_utils/branding.ts
```diff
@@ -176,6 +176,10 @@ export const color: Color = {
   tapHighlight: 'rgba(221, 221, 221, .4)', // gray, 40%
 }
 
+export const pxToInteger = (size:string):number => {
```
PerrineBoissieres: OMG thanks for this <3

src/stepper/Stepper.tsx
```diff
@@ -70,6 +108,23 @@ export default class Stepper extends PureComponent<StepperProps, StepperState> {
     this.props.onChange({ name: this.props.name, value })
   }
 
+  handleFontSize() {
+    if (! this.ref.current) {
```
Tuxosaurus: Remove whitespace after `!`

src/stepper/specifications/stepper.md
```diff
@@ -0,0 +1,46 @@
...
+
+When reaching the highest value the “increase” button is disabled. Same thing when reaching the lowest value, the “decrease” button is disabled.
+
+We can’t display more than 1 stepper on the same line.
+
+When user is coming back from the next step, we remember the value.
+
+> Long texts
+
+It can’t go on more than 1 line. If data is too long we reduce the font size (see edge case).
```
Tuxosaurus: I guess we can remove the `(see edge case)` part

PR #443, opened 2019-11-13
Add a11y attributes for Itinerary component

nickoola:


src/itinerary/story.tsx
```diff
@@ -28,59 +31,80 @@ const places = [
...
 
 stories.add('with proximity', () => {
   const Distances = ['NONE', 'CLOSE', 'MIDDLE', 'FAR']
-  const fromAddonLabel = text('From addon label', 'Lille')
-  const toAddonLabel = text('To addon label', 'Biarritz')
-  const headlineLabel = text('Headline label', 'Mon 11 December')
+  const fromAddon = text('From addon', 'Lille')
+  const fromAddonLabel = text('From addon label', 'Driver departure')
+  const toAddon = text('To addon', 'Biarritz')
+  const toAddonLabel = text('To addon label', 'Dirver arrival')
```
Tuxosaurus: `Dirver` typo

src/itinerary/Itinerary.tsx
```diff
@@ -100,6 +109,7 @@ const Itinerary = ({
               itemProp="location"
               itemScope
               itemType="http://schema.org/Place"
+              aria-label={place.stepLabel}
```
mcellier: What happens if `place.stepLabel` is undefined? 
Also, `stepLabel` should probably be in our Place definition (https://github.com/blablacar/ui-library/blob/master/src/typings/place.d.ts)

src/itinerary/Itinerary.tsx
```diff
@@ -50,9 +56,12 @@ const Itinerary = ({
             <BlankSeparator />
           </Fragment>
       )}
-      <ul className={cc([className, { 'kirk-itinerary--small': isSmall }])}>
+      <ul
+        className={cc([className, { 'kirk-itinerary--small': isSmall }])}
+        aria-labelledby={ariaLabelledBy}
```
mcellier: if null, we have an empty `aria-labelledby`?

src/itinerary/Itinerary.tsx
```diff
@@ -110,7 +120,7 @@ const Itinerary = ({
                     : place.mainLabel
                 }
               />
-              <Component {...hrefProps}>
+              <Component {...hrefProps} aria-label={place.actionLabel}>
```
mcellier: What happens if place.stepLabel is undefined?
Also, actionLabel should probably be in our Place definition (https://github.com/blablacar/ui-library/blob/master/src/typings/place.d.ts)

src/typings/place.d.ts
```diff
@@ -1,4 +1,6 @@
 declare interface Place {
+  stepLabel?: string,
```
mcellier: extra comma

src/itinerary/Itinerary.tsx
Tuxosaurus: As a "spectator" it's weird that this conversation is marked as "resolved" when there was no reply...

I guess the answer is that the attribute is not rendered if it's undefined, but still.

src/typings/place.d.ts
```diff
@@ -1,4 +1,6 @@
 declare interface Place {
+  stepLabel?: string
```
Tuxosaurus: I would name them so that we know what they are used for just by reading the names:
stepAriaLabel
actionAriaLabel

src/itinerary/Itinerary.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { Fragment } from 'react'
+import React, { Fragment, Children } from 'react'
```
Tuxosaurus: Why `Children`?

PR #447, opened 2019-11-19
Fix a11y content issue on SuccessModal

nickoola:


src/successModal/SuccessModal.tsx
```diff
@@ -58,15 +58,16 @@ class SuccessModal extends Component<SuccessModalProps> {
...
-        ariaLabelledBy={ariaLabelledBy}
-        ariaDescribedBy={ariaDescribedBy}
+        ariaLabelledBy={successContentId}
       >
         <img
           className={`${baseClassName}-bodyItem ${baseClassName}-image`}
           src={imageSrc}
           alt={imageText}
+          // This image is always decorative
+          aria-hidden={true}
```
mcellier: Shouldn't we just put an empty `alt` then?

src/successModal/SuccessModal.tsx
```diff
@@ -37,15 +38,14 @@ class SuccessModal extends Component<SuccessModalProps> {
       forwardedRef,
       imageSrc,
       imageText,
-      ariaLabelledBy,
-      ariaDescribedBy,
```
mcellier: Breaking change?

src/successModal/SuccessModal.tsx
nickoola: That's what we do with the default value. I let the possibility to add a description (SEO purpose, image not loaded...).

src/successModal/SuccessModal.tsx
nickoola: No previously those props were not well used and not defined in the render at the end.

PR #448, opened 2019-11-20
Fix random size issue on large stepper

mcellier:
Fix the following issue by preventing flex-grow on LARGE display (it's always displayed at 100% of its container).

image-2019-11-14-17-21-31-451

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **[FIX]** Fix random font-size issue in `Stepper` using `StepperDisplay.LARGE`
 [...]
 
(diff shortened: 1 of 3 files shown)


wakooka:
I don't think the price should get bigger when a button dissapears. Can you check with a designer?

mcellier:
> I don't think the price should get bigger when a button dissapears. Can you check with a designer?
> I have this in the dev specs:
> 
> > When reaching the highest value the “increase” button is disabled. Same thing when reaching the lowest value, the “decrease” button is disabled.
> > Buttons should not dissapear then.

Buttons don't disapear. The GIF is the visual representation of what was wrong :)

PR #451, opened 2019-11-20
Set initialMonth in DatePicker

mcellier:
In our publication flow, when selecting a date in a far future, we end up in the return date step with all next dates disabled. We have to either scroll down or click next months button until arriving to the first available date.

I'm using the `initialMonth` prop from `react-day-picker` to set it to the `initialDate`.

TESTED: in publication flow, desktop & mobile display

### DIFF
--- a/src/datePicker/DatePicker.tsx
+++ b/src/datePicker/DatePicker.tsx
@@ -153,6 +153,7 @@ export class DatePicker extends PureComponent<DatePickerProps, DatePickerState> 
       weekdaysLong,
       months,
       firstDayOfWeek,
+      initialDate,
     } = this.props
     const { date } = this.state
     const layoutClassName = `months-grid-${numberOfMonths}`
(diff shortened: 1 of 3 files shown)


src/datePicker/DatePicker.tsx
```diff
@@ -180,6 +181,7 @@ export class DatePicker extends PureComponent<DatePickerProps, DatePickerState>
           months={months}
           firstDayOfWeek={firstDayOfWeek}
           localeUtils={{ ...DayPicker.LocaleUtils, formatMonthTitle: this.formatMonthTitle }}
+          initialMonth={initialDate}
```
wakooka: unit test?

mcellier: The feature is from the `react-day-picker` lib so it's not our job to test it. Unless you want me to test that we pass the prop each time?

wakooka: Looks like we didn't do that for DatePicker but we normally do check that our props are forwared in other components. I'll approve it and leave it up to you.

PR #456, opened 2019-11-22
Ensure that image in `SuccessModal` doesn't take more than 33vh

mcellier:


Tuxosaurus:
Can you add screenshots so we can better visualize what's the problem?

mcellier:
> Can you add screenshots so we can better visualize what's the problem?

Image from iOS (1)

PR #464, opened 2019-11-26
Fix click behavior on textarea to set focus

grosbouddha:
The textarea has margins that look like the textarea but won't focus the textarea if clicked.
A click handler was added to  make sure we focus the textarea when clicking these margins.

TESTED=Locally in storybook

### DIFF
--- a/src/textarea/Textarea.tsx
+++ b/src/textarea/Textarea.tsx
@@ -105,6 +105,16 @@ export default class Textarea extends PureComponent<TextAreaProps, TextAreaState
     this.maybeAdaptHeightToContent()
   }
 
+  // To ensure a proper scrolling behavior for the textarea when growing with content, we need to
+  // have a wrapper independent from the native textarea. This wrapper is bigger that the
+  // the textarea but still looks like the textarea (same background color). When activating the
+  // wrapper, the focus should set the focus on the native textarea.
+  onWrapperClick = () => {
+    if (this.textareaRef && this.textareaRef.current) {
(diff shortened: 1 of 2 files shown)


src/textarea/index.tsx
```diff
@@ -22,7 +22,7 @@ const StyledTextarea = styled(Textarea)`
     display: flex;
     box-sizing: border-box;
     color: ${color.primaryText};
-    background-color: ${color.inputBackground};
+    background-color: yellow;
```
Tuxosaurus: changed for testing purposes?

grosbouddha: You don't like yellow :) ?
Updated, thanks for catching :)

Tip-Sy: 👏

src/textarea/Textarea.tsx
```diff
@@ -237,6 +247,7 @@ export default class Textarea extends PureComponent<TextAreaProps, TextAreaState
       <div className={cc(['kirk-textarea', prefix({ error: !!error, disabled }), className])}>
         {label && <label htmlFor={id}>{label}</label>}
         <div
+          onClick={this.onWrapperClick}
```
Tip-Sy: Just wondering: is there also a way to pass a `onClick` handler to the component? If not, do we want to add that for future need?

grosbouddha: Let's wait to actually need that before adding it.

PR #468, opened 2019-11-29
Add default Tabs panel transition. Disabled by default.

grosbouddha:
Add a new Tabs panel transition based on the current Tabs panel transition on the search result page.
This new transition is disabled by default.

Note for reviewers: I had to move the classname root to a parent div element (instead of fragment) to be able to target the tab panels in the Styled component. I think this should work with our current search and messaging use cases.

TESTED=Locally in storybook (new story added)

### DIFF
--- a/src/tabs/Tabs.tsx
+++ b/src/tabs/Tabs.tsx
@@ -25,6 +25,7 @@ interface TabsProps {
   readonly className?: Classcat.Class
   readonly tablistWrapperClassName?: Classcat.Class
   readonly isWrapped?: boolean
+  readonly useDefaultPanelTransition?: boolean
 }
 
 interface TabsState {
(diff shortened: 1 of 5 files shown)


src/tabs/index.tsx
```diff
@@ -5,6 +5,9 @@ import Tabs from './Tabs'
 
 const highlightHeight = '2px'
 
+// Some arbitrary delay to build the panel before starting the revealing transition.
```
Tip-Sy: Thanks for the comment

src/tabs/Tabs.tsx
```diff
@@ -176,8 +178,8 @@ export class Tabs extends PureComponent<TabsProps, TabsState> {
     }
 
     return (
-      <Fragment>
-        <div className={cc(['kirk-tabs', className, { 'kirk-tabs-fixed': isFixedTabs }])}>
+      <div className={cc([className])}>
```
Tip-Sy: Just wondering: do you know the impact of changing the Fragment into a div? I remember we broke the trip cards in search with such a change few days ago. I guess here it's fine as it's not a dependancy on some other components, and you will update them with your PR on Kairos.

Tuxosaurus: I fear moving `className` to a `div` wrapping both the tabs and panels is a breaking change. That's what we use to have the tabs sticky (we don't want to stick the panels too)

PR #473, opened 2019-12-09
Allow rendering of non-highlighted ItemAction

grosbouddha:
This is needed when rendering many ItemAction in the same page, they don't need to be all highlighted.

TESTED-Unit tests and locally in storybook

### DIFF
--- a/src/itemAction/index.tsx
+++ b/src/itemAction/index.tsx
@@ -6,6 +6,7 @@ import Item, { ItemStatus } from '_utils/item'
 import prefix from '_utils'
 
 export interface ItemActionProps {
+  readonly highlighted?: boolean
   readonly tag?: JSX.Element
   readonly className?: Classcat.Class
   readonly href?: string | JSX.Element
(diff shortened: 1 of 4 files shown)


src/itemAction/index.tsx
```diff
@@ -29,6 +30,7 @@ class ItemAction extends PureComponent<ItemActionProps> {
 
   render() {
     const {
+      highlighted = true,
```
mcellier: We use something like `static defaultProps: Partial<**> = {}` for default values

src/itemAction/index.unit.tsx
```diff
@@ -77,5 +77,17 @@ describe('ItemAction', () => {
       expect(wrapper.find(Loader).exists()).toBe(true)
       expect(wrapper.find(Loader).prop('done')).toBe(true)
     })
+
+    it('Should render an highlighted ItemAction by default', () => {
```
Tip-Sy: nit: for consistency, use lowercase everywhere "should" (or uppercase below)

grosbouddha: Done

src/itemAction/index.tsx
grosbouddha: Done, thanks !

PR #475, opened 2019-12-10
Add rightTitleStrikeThrough prop to Item and ItemData

nickoola:


nickoola:
Used for discounted price cases
Capture d’écran 2019-12-10 à 09 33 14

src/_utils/item/Item.tsx
```diff
@@ -29,6 +29,7 @@ export interface ItemProps {
   readonly leftAddon?: React.ReactNode
   readonly rightTitle?: string
   readonly rightTitleDisplay?: TextDisplayType
+  readonly rightTitleStrikeThrough?: boolean
```
mcellier: Did you consider to add a new value to `TextDisplayType` instead? That way, it'll be usable in every place and prevent us adding another prop to an already fat component

Tuxosaurus: Hum I see what @nickoola said, adding a new value to `TextDisplayType` would work but mean that the strikethrough only applies to a given size of text.

Though for now I agree just one new displayType with strikethrough and one font size is OK

PR #481, opened 2019-12-13
Fix layout issue on TopTripCardList

wakooka:
2 issues fixed:
- The cards were not equal size on large device but dependent on content.
- The padding right scroll hack were not applying because the container didn't have a `position: relative`

See column issue:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-[...]
+- **[FIX]** Fix layout issue on `TopTripCardList`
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Unreleased
 
-[...]
+- **[FIX]** Fix layout issue on TopTripCardList
```
mcellier: nit: Add backquotes around TopTripCardList

wakooka: Done

PR #482, opened 2019-12-13
add slider keyboard controls to Stepper

Tuxosaurus:
Stepper can now be controlled with keyboard (decrement with left/down, increment with right/up, min with Home, max with End).

It almost looks the same, except for the focus.
It looks like this now in Chrome and Safari:
Screen Shot 2019-12-13 at 19 00 34

Unfortunately, I can't get Firefox to show the focus.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **[UPDATE]** Add keyboard controls to `Stepper`
 - [...]
 
 # v16.2.0 (16/12/2019)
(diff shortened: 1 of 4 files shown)


nickoola:
Tested with VoiceOver, I by using right/left button the increment/decrement feature is not working.
We could try to use the output tag:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_slider_role

nickoola:
This PR enhance the a11y for keyboard navigation (without screen reader).
So let's merge it and create an other task to fix the Screen reader case.

PR #485, opened 2019-12-18
[prototype] Pixar layout normalization.

grosbouddha:
This PR introduces a < Column > wrapper and some utility CSS classes for solid and fluid column items.
UI widgets will be annotated with these classes to decide if they should spread over the column left/right spacing or not (see _utils/layout/index.jsx).
By default these classes are noop sand it's safe to have them on all the widgets. It's only when the widgets are nested inside a < Column > element that the new behaviors are activated. Currently
the < Column > is not used (beside some storybook pages) and will be used to migrate page by page or section by section the new behavior.

I created new pages storybook pages (SRP, messaging, ridedetails) that uses this layout. Ultimately all pages could be in the storybook too.

Note that some widgets are currently broken when the new behavior is activated (the Itinerary for instance). This will be fixed as follow tasks after the initial commit.

image

TESTED=Storybook that widgets are not broken when not wrapped in a <Column>

### DIFF
--- a/pages/rideDetailsStory.tsx
+++ b/pages/rideDetailsStory.tsx
@@ -1,0 +1,77 @@
+import React from 'react'
+
+import { storiesOf } from '@storybook/react'
+import {withKnobs, text, boolean, select} from '@storybook/addon-knobs'
+
+import { Column } from '_utils/layout'
+import ItemInfo from 'itemInfo'
+import ItemData from 'itemData'
+import Divider from 'divider'
+import TheVoice from 'theVoice'
+import BubbleIcon from 'icon/bubbleIcon'
+import SmokeIcon from 'icon/smokeIcon'
+import PetIcon from 'icon/petIcon'
+import { status } from '_utils/icon/status'
+import ItemAction from 'itemAction'
+import SubHeader from 'subHeader'
+import Itinerary from 'itinerary'
+import ItemChoice from 'itemChoice'
+import Avatar from 'avatar'
+
+const stories = storiesOf('Pages|Ride details', module)
+stories.addDecorator(withKnobs)
+
+stories.add('default', () => (
+    <div style={{width: '600px', margin: 'auto', border: 'lightgray 1px solid'}}>
+        <Column>
+            <TheVoice>Ven. 11 octobre</TheVoice>
+            <Itinerary
+                fromAddon='Paris'
+                toAddon='Marseille'
+                places={[
+                    {
+                        mainLabel: 'Paris',
+                        isoDate: '2017-12-11T09:00',
+                        time: '09:00',
+                        href: '#',
+                    },
+                    {
+                        mainLabel: 'Marseille',
+                        isoDate: '2017-12-11T12:00',
+                        time: '17:00',
+                    },
+                ]}
+                small={boolean('small', false)}
+            />
+            <ItemData data='17,50 €' dataInfo='Par place' mainInfo='3 places restantes'/>
+            <Divider />
+            <ItemChoice label="Vince" rightAddon={<Avatar  />} href="#" />
+            <ItemAction action='Contacter Vince' leftAddon={<BubbleIcon />} href="#"/>
+            <Divider />
+            <ul>
+                <ItemInfo
+                    mainInfo="Fumer nest pas autporisé dans la voiture."
+                    icon={<SmokeIcon  status={status.OFF} />}
+                />
+                <ItemInfo
+                    mainInfo="Pas d'animaux dans la voiture."
+                    icon={<PetIcon status={status.OFF} />}
+                />
+            </ul>
+            <Divider />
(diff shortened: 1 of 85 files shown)


pages/rideDetailsStory.tsx
```diff
@@ -0,0 +1,77 @@
...
+                        time: '09:00',
+                        href: '#',
+                    },
+                    {
+                        mainLabel: 'Marseille',
+                        isoDate: '2017-12-11T12:00',
+                        time: '17:00',
+                    },
+                ]}
+                small={boolean('small', false)}
```
Tuxosaurus: Not sure we want to keep this knob. Or maybe name it like "[Itinerary] small" so we know which component is affected

pages/searchResultStory.tsx
```diff
@@ -0,0 +1,164 @@
...
+    <div style={{width: '600px', margin: 'auto', border: 'lightgray 1px solid'}}>
+        <Column>
+            <UneditableTextField href='#' addOn={<SearchIcon />}>
+                <div>
+                    <Text>Top stuff</Text>
+                    <Text>Bottom stuff</Text>
+                </div>
+            </UneditableTextField>
+
+            <TopTripCardList isWrapped={boolean('isWrapped', false)}>
```
Tuxosaurus: Not sure we want to keep this knob. Or maybe name it like "[TopTripCardList] isWrapped" so we know which component is affected

pages/rideDetailsStory.tsx
wakooka: Can we group knobs? (I don't think so, but since we now have pages in ui-library, it would be cool if we could)

pages/rideDetailsStory.tsx
```diff
@@ -0,0 +1,77 @@
...
+            />
+            <ItemData data='17,50 €' dataInfo='Par place' mainInfo='3 places restantes'/>
+            <Divider />
+            <ItemChoice label="Vince" rightAddon={<Avatar  />} href="#" />
+            <ItemAction action='Contacter Vince' leftAddon={<BubbleIcon />} href="#"/>
+            <Divider />
+            <ul>
+                <ItemInfo
+                    mainInfo="Fumer nest pas autporisé dans la voiture."
```
wakooka:
```suggestion
                    mainInfo="Fumer nest pas autorisé dans la voiture."
```

src/_utils/layout/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+        <Fragment>{children}</Fragment>
+    </div>
+export const Column = styled(UnstyledColumn)`
+  & ${FLUID_COLUMN_ITEM} {
+    padding-left: ${columnItemHorizontalSpacing} !important;
+    padding-right: ${columnItemHorizontalSpacing} !important;
+  }
+  
+  & ${SOLID_COLUMN_ITEM} {
+    margin-top: 8px;
```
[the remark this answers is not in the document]
wakooka: Or create a const like you did for `gutterWidth`, and `internalHorizontalPadding` based on `space.m`

src/_utils/layout/index.tsx
```diff
@@ -0,0 +1,54 @@
+import styled from 'styled-components'
+import { space } from '_utils/branding'
+import cc from 'classcat'
+import React, { Fragment } from 'react'
+
+// Root class for a Column element.
+const KIRK_LAYOUT_COLUMN_CLASS = 'kirk-layout-column'
+
+// Class for a fluid column item.
+// The content of a fluid item will try to take all the hortizontal space of the column and will
```
wakooka:
```suggestion
// The content of a fluid item will try to take all the horizontal space of the column and will
```

src/_utils/layout/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+// overlap the left and right spacing of the column.
+// This class will be a noop if not contained inside a column with the class
+// KIRK_LAYOUT_COLUMN_CLASS. We use this particularity to enable on demand the column layout.
+export const KIRK_LAYOUT_FLUID_ITEM_CLASS = 'kirk-layout-fluid-item'
+
+// Class for a solid column item.
+// The content of a solid item has a visible frontier (a border, a shadow or delimiting background
+// color). This content will not overlap the left and right padding of a column container.
```
wakooka: nit: Add an example of component: Textfield, TripCard...

src/itinerary/Itinerary.tsx
```diff
@@ -139,7 +143,7 @@ const Itinerary = ({
               />
               <Component {...hrefProps} aria-label={place.actionAriaLabel}>
                 {!isSmall && (
-                  <time dateTime={place.isoDate}>
+                  <time className="kirk-itinerary-time-cell" dateTime={place.isoDate}>
```
wakooka: thanks :+1:

PR #491, opened 2019-12-31
Added ratings on tripCard

wakooka:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - **[FIX]** Fix `Itinerary` component display with only one stopover
 - **[FIX]** Fix `Itinerary` component display without time
 - [...]
+- **[NEW]** Added rating prop to `TripCard`
 
 # V17.0.0 (24/12/2019)
 
--- a/src/tripCard/TripCard.tsx
+++ b/src/tripCard/TripCard.tsx
@@ -6,6 +6,7 @@ import Avatar from 'avatar'
 import ComfortIcon from 'icon/comfortIcon'
 import LightningIcon from 'icon/lightningIcon'
 import LadyIcon from 'icon/ladyIcon'
+import Star from 'icon/starIcon'
 import Itinerary from 'itinerary'
 import Item from '_utils/item'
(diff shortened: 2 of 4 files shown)


src/tripCard/TripCard.tsx
```diff
@@ -172,7 +174,15 @@ const TripCard = ({
                 <div className="kirk-tripCard-avatar">
                   <Avatar image={driver.avatarUrl} />
                 </div>
-                <Text display={TextDisplayType.TITLE}>{driver.firstName}</Text>
+                <div>
+                  <Text display={TextDisplayType.TITLE}>{driver.firstName}</Text>
+                  {driver.rating && (
+                    <div className="kirk-tripCard-ratingContainer">
+                      <Star fill={1} size={16} />
```
Tuxosaurus: Maybe we can create a new case for the Rating component and use it here?

wakooka: I don't know, this case with only 1 star filled only exists in the component Profile / Search results (in sketch). I think we shouldn't make the Rating component more complex, since it's only used in the TripCard. WDYT?

Tuxosaurus: I have no strong opinion about this (hence why I used "maybe") so that's OK.

Another thing though is that after I looked at the design, I think there should be a little bit more space between the name and the rating

wakooka: Good catch, I'll double check the spacing

wakooka: Fixed & tested on SPA
image

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Unreleased
 
-- [...]
+- **[UPDATE]** Added rating prop to `TripCard`
```
mcellier: NEW ?

Tuxosaurus: Debatable, it's an update to TripCard, although it's updated with new stuff. I'd have used "UPDATE" myself

wakooka: I checked the changelog you're right NEW makes more sense based on on we used it before.

PR #492, opened 2020-01-03
Fix itinerary display with single step

mcellier:
Before:
Capture d’écran 2020-01-03 à 16 12 15

After:
Capture d’écran 2020-01-03 à 16 18 54

### DIFF
--- a/src/itinerary/index.tsx
+++ b/src/itinerary/index.tsx
@@ -107,6 +107,13 @@ const StyledItinerary = styled(Itinerary)`
     height: calc(6px + ${space.m} + ${gutterTopOffset});
   }
 
+  /* Only one step */
+  &
+    .kirk-itinerary-location.kirk-itinerary--departure.kirk-itinerary--arrival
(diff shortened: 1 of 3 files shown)


src/itinerary/index.tsx
```diff
@@ -107,6 +107,13 @@ const StyledItinerary = styled(Itinerary)`
     height: calc(6px + ${space.m} + ${gutterTopOffset});
   }
 
+  /* Only one step */
+  &
+    .kirk-itinerary-location.kirk-itinerary--departure.kirk-itinerary-location.kirk-itinerary--arrival
```
Tuxosaurus: I guess there's one unnecessary `.kirk-itinerary-location`

mcellier: Nice catch!

Tuxosaurus: I catch'em like https://media.giphy.com/media/Sf66mvUdRZru8/giphy.gif

PR #493, opened 2020-01-03
Fix margin in itinerary if we want to display content without time

mcellier:
Before:
Capture d’écran 2020-01-03 à 17 29 01

After:
Capture d’écran 2020-01-03 à 17 29 27

Context: 
6 months ago, we needed to display small itineraries in Your Rides, but the component is rendered through `TripCard`, so we didn't have lot of room to customize the thing. So I naively assumed that we should display the component as small if we don't have time nor subLabel (that's what the small prop was doing).

Except... that if we want to render itineraries with subLabel but without time, it breaks. 
Initial commit: https://github.com/blablacar/ui-library/commit/b5f58202d1e28ced0dd888476d003e132e084d29

### DIFF
--- a/src/itinerary/Itinerary.tsx
+++ b/src/itinerary/Itinerary.tsx
@@ -63,8 +63,9 @@ const Itinerary = ({
   small = false,
   headline = null,
 }: ItineraryProps) => {
-  // Dislay itinerary as small if required or if we don't have time or subLabel for provided places
-  const isSmall = small || places.filter(p => !isEmpty(p.time) || !isEmpty(p.subLabel)).length === 0
+  // Add the small class if we don't have times to prevent empty content
+  const withTime = places.filter(p => !isEmpty(p.time)).length > 0
+
   // Remove aria-labelledby attribute if aria-label already used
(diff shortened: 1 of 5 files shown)


mcellier:
Note: i copied the builded kirk into kairos and checked all majors flows (search, publish, your rides, ride plan drvr) to check. It should be OK :D

src/itinerary/Itinerary.tsx
```diff
@@ -75,7 +76,7 @@ const Itinerary = ({
           <BlankSeparator />
         </Fragment>
       )}
-      <ul className={cc([{ 'kirk-itinerary--small': isSmall }])}>
+      <ul className={cc([{ 'kirk-itinerary--small': small || !withTime }])}>
```
Tuxosaurus: Definitely "nit" since I don't have any other suggestion but that'd be cool to find a different name for `kirk-itinerary--small` since it's not just when it's "small" (and well even so, it doesn't make it smaller, more simplier)

mcellier: `kirk-itinerary--no-sidebar`?

PerrineBoissieres: `kirk-itinerary--noTime`?

PR #497, opened 2020-01-08
Fix visible input range track on Safari

PerrineBoissieres:
### Before:
IMG-1028

### After:
Image from iOS (2)


src/stepper/index.tsx
```diff
@@ -55,10 +55,15 @@ const StyledStepper = styled(Stepper)`
     -webkit-appearance: none; /* Hides the slider so that custom slider can be made */
     width: 100%; /* Specific width is required for Firefox. */
     background: transparent; /* Otherwise white in Chrome */
+    border: transparent;
```
mcellier: Unnecessary?

PR #505, opened 2020-01-20
Refactor itinerary

Tuxosaurus:
Refactoring:
- clean CSS
- remove pseudo-elements
- create "Bullet" component

Only visual change is the "Bullet" size is bigger, as intended in the design


src/bullet/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+    box-sizing: content-box;
+    width: 6px;
+    height: 6px;
+    background-color: ${color.white};
+    border: 2px solid ${color.primaryText};
+    border-radius: 50%;
+  }
+
+  &.kirk-bullet--addon {
+    border-color: ${color.border};
```
pith: Not sure we want to use the border semantic here.

src/bullet/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  &.kirk-bullet--small {
+    width: 4px;
+    height: 4px;
+  }
+
+  &.kirk-bullet--map {
```
pith: I'm not sure map is really clear here. As it will not be used in the map but in the itinerary component

pith: It's more a highlight WDYT? 🤷‍♂

Tuxosaurus: @mcellier actually requested to have a separate component so he could use it on maps too

src/bullet/index.tsx
Tuxosaurus: Yes, I only moved what was already there. Would `color.fadedText` be OK? It's actually used along with an actual faded text

src/bullet/index.tsx
mcellier: Yes, I did!

src/bullet/Bullet.tsx
```diff
@@ -0,0 +1,24 @@
...
+  ADDON = 'addon',
+  MAP = 'map',
+}
+
+export interface BulletProps {
+  readonly className?: Classcat.Class
+  readonly type?: BulletTypes
+}
+
+const Bullet = ({ className, type }: BulletProps) => {
```
mcellier: Use defaultProps instead of isEmpty(type)

PR #506, opened 2020-01-21
Added amenity icons

wakooka:
Added plugUsbSocketIcon, priorityServiceIcon, quietIcon, qrCode, recliningSeat, seatBelt, seatWithTable, snack, airConditioning, audio, bikeArea, blanket, child, comfortSeat, crew, drink, genericAmenity, handLuggage, holdLuggage, legRest, legRoom, light, magazine, meal, seatSliding, sleep, standardSeat, videoSystem, vipLounge, wc, wifi icons


src/icon/comfortSeat.tsx
```diff
@@ -0,0 +1,21 @@
+// tslint:disable:max-line-length
+import React from 'react'
+import StatusIcon from '_utils/icon/status'
+
+export const ComfortSeat = (props: Icon) => (
+  <StatusIcon {...props}>
+    <g fill="#708C91" fillRule="nonzero">
```
mcellier: Hardcoded color?

src/icon/genericAmenity.tsx
```diff
@@ -0,0 +1,17 @@
+// tslint:disable:max-line-length
+import React from 'react'
+import StatusIcon from '_utils/icon/status'
+
+export const GenericAmenity = (props: Icon) => (
+  <StatusIcon {...props}>
+    <path
+      d="M17.7 3a.5.5 0 01.39.187l3.8 4.738c.14.174.147.419.017.6l-9.501 13.267a.5.5 0 01-.812 0L2.093 8.525a.496.496 0 01.016-.6l3.8-4.738A.5.5 0 016.3 3zm-.24.995H6.539L3.124 8.25 12 20.646l8.875-12.395-3.415-4.256zm-.71 3.743a.499.499 0 01.09.99l-.09.008h-9.5a.499.499 0 01-.09-.99l.09-.008h9.5z"
+      fill="#708C91"
```
mcellier: Hardcoded color?

src/icon/index.tsx
```diff
@@ -1,111 +1,160 @@
+import AirConditioningIcon from './airConditioning'
```
mcellier: Don't you want to add the `Icon` suffix for all to be consistent with the existing ones?

src/icon/legRest.tsx
```diff
@@ -0,0 +1,17 @@
+// tslint:disable:max-line-length
+import React from 'react'
+import StatusIcon from '_utils/icon/status'
+
+export const LegRest = (props: Icon) => (
```
mcellier: Are you sure you need them to extends `StatusIcon`, not the `BaseIcon`?

src/icon/magazine.tsx
```diff
@@ -0,0 +1,30 @@
+// tslint:disable:max-line-length
+import React from 'react'
+import StatusIcon from '_utils/icon/status'
+
+export const Magazine = (props: Icon) => (
+  <StatusIcon {...props}>
+    <g fill="none" fillRule="evenodd">
+      <path
+        d="M5.913 12.37a.5.5 0 01.09.992l-.09.008-2.979-.001v6.457c0 .638.492 1.169 1.114 1.233l.126.006a.5.5 0 110 1c-1.18 0-2.151-.92-2.234-2.08l-.005-.159V12.87a.5.5 0 01.41-.492l.09-.008h3.478z"
+        fill="#708C91"
```
mcellier: Hardcoded color

src/icon/magazine.tsx
```diff
@@ -0,0 +1,30 @@
...
+        fillRule="nonzero"
+      />
+      <path
+        d="M16.348 6.783h1.739-1.74zm0 3.478h1.739-1.74zM9.39 13.739h8.696-8.696zm0 3.478h8.696-8.696zm0-10.434h3.479v3.478H9.39V6.783z"
+        stroke="#708C91"
```
mcellier: Hardcoded color

src/icon/index.tsx
wakooka: It' done on purpose some existing ones are missing the icon suffix, so I didn't introduce a new way, it was already hybrid. Also we don't need this in the filename as long as they are exporting as `AirConditioningIcon`.
I'd rather remove icon suffix from all files in their filename. WDYT?

src/icon/legRest.tsx
wakooka: Yes all of them need to be on or off, so that we know if a amenity is available or not.

src/icon/index.tsx
mcellier: I'm OK as long as we're consistent. So: suffix in exported components, no suffix in directories?

wakooka: Yep, but that would mean breaking the paths already used in the lib and the SPA. Let's think about it.

Tuxosaurus: I'd say we choose to remove them as a rule, but the change should take place in another PR

PR #511, opened 2020-01-24
Removed z-index from itinerary bullets and use source order instead

wakooka:
No impact on Itinerary
image


src/itinerary/Itinerary.tsx
```diff
@@ -160,8 +160,8 @@ const Itinerary = ({
                   </time>
                 )}
                 <div className="kirk-itinerary-location-city">
-                  <Bullet />
                   <div className="kirk-itinerary-road" aria-hidden="true" />
+                  <Bullet />
```
Tuxosaurus: There's another "Bullet above Road" earlier in the fromAddon

wakooka: That one had no impact because it was already in the correct order, but I'll update it for consistency :+1:

PR #515, opened 2020-01-29
Update MeetingPointIcon (new active state, optional shadow)

mcellier:
mp icon

### DIFF
--- a/src/_utils/icon/index.tsx
+++ b/src/_utils/icon/index.tsx
@@ -1,11 +1,17 @@
 import styled from 'styled-components'
 
+import { shadow } from '_utils/branding'
 import BaseIcon from './BaseIcon'
 
 const StyledBaseIcon = styled(BaseIcon)`
   &.kirk-icon-wrapper {
     display: inline-block;
     position: relative;
+  }
+
+  &.kirk-icon-shadowed {
+    border-radius: 50%;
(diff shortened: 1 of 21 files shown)


src/_utils/icon/index.tsx
```diff
@@ -8,6 +9,11 @@ const StyledBaseIcon = styled(BaseIcon)`
     position: relative;
   }
 
+  &.kirk-icon-shadowed {
```
mcellier: I wanted to add the style only for `MeetingPointIcon`, but we break our stories doing so (styled HOC prevents access to defaultProps)

src/icon/meetingPoint.tsx
```diff
@@ -1,51 +1,44 @@
 // tslint:disable:max-line-length
 import React from 'react'
+import cc from 'classcat'
 import BaseIcon from '_utils/icon'
 import { BaseIconDefaultProps } from '_utils/icon/BaseIcon'
 import { color } from '_utils/branding'
 
 export interface MeetingPointIconProps extends Icon {
   readonly active?: boolean
+  readonly shadowed?: boolean
```
mcellier: I initially wanted to put this globally available for all icons, but most of them renders poorly with a round shadow around them.

PR #516, opened 2020-01-29
Add option to toggle road highlighting in Itinerary

mcellier:
:'( 

highlight road

This case is used in publication flow


Tuxosaurus:
Too bad we didn't notice that before the upcoming refacto but well

PR #519, opened 2020-01-31
Add a way to desactivate the hover background color on ItemAction

grosbouddha:


src/itemAction/story.tsx
```diff
@@ -30,6 +30,11 @@ stories.add(
           subLabel={isSubLabel ? text('Secondary info', 'Info') : null}
           href="#"
         />
+        <ItemAction
+          action="Action 3 - A action without hover color."
```
Tip-Sy: nit: "An action"

grosbouddha: Done

PR #528, opened 2020-02-06
Added rule on itinerary collapsible for displaying stopover directly

wakooka:
Kapture 2020-02-06 at 17 02 19

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-- [...]
+- **[UPDATE]** Itinerary is no longer collapsed when we have only one stopover
 
 # v19.1.0 (07/02/2020)
 
--- a/src/itinerary/Itinerary.tsx
+++ b/src/itinerary/Itinerary.tsx
@@ -50,6 +50,8 @@ const computeRootA11yProps = (ariaLabel?: string, ariaLabelledBy?: string): Root
 // Get places between departure and arrival
 const getIntermediatePlaces = (places: Place[]) => places.slice(1, -1)
 
+const STOPOVER_COUNT_TO_COLLAPSE_FROM = 1
+
 const renderLocation = (
   places: Place[],
   isArrival: boolean,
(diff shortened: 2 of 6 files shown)


src/itinerary/story.tsx
```diff
@@ -71,7 +98,15 @@ stories.add('default', () => {
   const toAddon = text('To addon', 'Biarritz')
   const toAddonLabel = text('To addon label', 'Driver arrival')
   const headline = text('Headline', 'Mon 11 December')
-  const isStopover = boolean('Stopover', false)
+  const placesSelect = select(
```
Tuxosaurus: 👍

src/_utils/itineraryCollapsible/ItineraryCollapsible.tsx
```diff
@@ -26,16 +27,23 @@ export default class ItineraryCollapsible extends PureComponent<ItineraryCollaps
...
+    collapsed: this.props.collapsed,
   }
 
   componentDidMount() {
     this.addListeners()
   }
 
+  componentDidUpdate(prevProps: ItineraryCollapsibleProps) {
+    if (prevProps.collapsed !== this.props.collapsed) {
+      this.setState({ collapsed: this.props.collapsed })
```
mcellier: In kairos we're blocked by eslint as it's considered a bad practice to update the state in a `componentDidUpdate` (https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-did-update-set-state.md)

Should you consider using `getDerivatedStateFromProps`?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 - **[FIX]** Fix background for `HighlightSection`
 - **[UPDATE]** Rename left nav widget section from 'Others' to 'Widgets'
+- **[UPDATE]** `ItineraryCollapsible` extended by default when 1 stopover
```
mcellier: `- **[UPDATE]** `Itinerary` is no more collapsed when we have only one stopover`

src/itinerary/Itinerary.tsx
```diff
@@ -50,6 +50,8 @@ const computeRootA11yProps = (ariaLabel?: string, ariaLabelledBy?: string): Root
 // Get places between departure and arrival
 const getIntermediatePlaces = (places: Place[]) => places.slice(1, -1)
 
+const stopoverCountToCollapseFrom = 2
```
mcellier: Consistency: 
- Put your const in UPPER_CAMEL_CASE
- More understandable to let it at 1 since the behavior is "no collapse if only one"

You'll need to change the code below:
`intermediatePlaces.length > stopoverCountToCollapseFrom`

PR #542, opened 2020-02-11
Improve HighlightSection visual

grosbouddha:
image

### DIFF
--- a/src/layout/section/highlightSection/highlightSection.tsx
+++ b/src/layout/section/highlightSection/highlightSection.tsx
@@ -1,23 +1,27 @@
 import React from 'react'
 import BaseSection, { SectionContentSize } from 'layout/section/baseSection'
+import SubHeader from 'subHeader'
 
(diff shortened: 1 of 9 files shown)


src/layout/section/highlightSection/highlightSection.tsx
```diff
@@ -1,23 +1,27 @@
...
-  const { className, children, contentClassName } = props
+  const { className, children, contentClassName, title } = props
   return (
     <BaseSection
+      tagName="article"
       className={className}
       contentClassName={contentClassName}
       contentSize={SectionContentSize.LARGE}
     >
+      {title !== null && <SubHeader>{title}</SubHeader>}
```
wakooka: What if title is `undefined`, use this instead:
`{title && <SubHeader>{title}</SubHeader>}`

grosbouddha: Done

PR #545, opened 2020-02-11
New MediaContentSection section and Columns/Column components

grosbouddha:
- Create new MediaContentSection (marketing content + media) that match specs on home and home bus (both responsive modes).
- Introduce private Columns and Column helper to create equally sized columns next to each other.

Large:
image

Small:
image


src/layout/section/mediaContentSection/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+    & .kirk-media-content-title {
+      /* Remove padding top from subheader to allow proper vertical centering. */
+      padding-top: 0;
+    }
+  }
+
+  @media (${responsiveBreakpoints.isMediaSmall}) {
+    & .kirk-media-content-img {
+      height: 136px;
+      min-height: 0;
```
wakooka: `min-height: 0` ?

grosbouddha: It's needed to override the default min-height: 320px; from above.

PR #547, opened 2020-02-12
Fixes to mediaContentSection

grosbouddha:
- Add missing border radius
- Use <article> html tag
- Properly center image in small mode
- Left align title in large mode


grosbouddha:
Added more fixes, PTAL

PR #554, opened 2020-02-14
Tabs and TabsSection layout overhaul

grosbouddha:
The core issue was: Some responsive layout was both in Tabs, TabsSection and in Kairos. With this PR, the Tabs is just a full width component, no padding, no wrapped. All layout code (reponsive, paddings) is inside TabsSection. The TabsSection is now divided into 2 full witdh subsections (one for the tabs, one for the panels) which are full width (and does not use a BaseSection anymore) and it sets layout contrains on nested elements inside these sub sections. Having this full width subsection for the tab list gives a perfect container to set the full width separator border too: We don't need a double border in Tabs and TabsSection like before.

Complete changelog:
- create two new virtual sections inside the Tabs => one for the tab lists, one for the panels. Each section is full width but can have constrained elements inside depending on viewport
- remove isWrapped props and replace it with media queries on kirk side
- add margin bottom below tabs (to remove margin bottom from kairos)
- remove unused tablistWrapperClassName
- add wrapper around tabs and panels
- create new tabsClassName (previously className)
- Remove paddings on panels in stories for Tabs and TabsSection to ensure proper vertical alignment between tabs/content
- no more double border (one from TabsSection overlapping one from Tabs). Only one defined on Tabs.
- proper fullwidth layout to remove the negative margins from Kairos

TESTED=Locally in storybook for tabs and tabssection. Locally in kairos (both in small/large/sticky modes on SRP) with this integration branch: BOYSCOUT-integrate-tabs-overhaul

### DIFF
--- a/src/tabs/Tabs.tsx
+++ b/src/tabs/Tabs.tsx
@@ -1,4 +1,4 @@
-import React, { createRef, PureComponent, RefObject, Fragment } from 'react'
+import React, { createRef, PureComponent, RefObject } from 'react'
 import cc from 'classcat'
 import Badge from 'badge'
 
@@ -23,8 +23,7 @@ export interface TabsProps {
   readonly onChange?: Function
   readonly status?: TabStatus
   readonly className?: Classcat.Class
-  readonly tablistWrapperClassName?: Classcat.Class
-  readonly isWrapped?: boolean
+  readonly tabsClassName?: Classcat.Class
 }
 
 interface TabsState {
@@ -78,8 +77,7 @@ export class Tabs extends PureComponent<TabsProps, TabsState> {
     onChange() {},
     status: TabStatus.SCROLLABLE,
     className: '',
-    tablistWrapperClassName: '',
-    isWrapped: false,
+    tabsClassName: '',
   }
 
   state: TabsState = {
@@ -88,6 +86,7 @@ export class Tabs extends PureComponent<TabsProps, TabsState> {
   }
 
   highlightRef: RefObject<HTMLDivElement> = React.createRef()
+  tabsGroupRef: RefObject<HTMLDivElement> = React.createRef()
 
   static STATUS = TabStatus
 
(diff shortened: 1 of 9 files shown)


src/tabs/Tabs.tsx
```diff
@@ -227,22 +228,24 @@ export class Tabs extends PureComponent<TabsProps, TabsState> {
...
-        })}
-      </Fragment>
+        <div className="kirk-panels-section" role="presentation">
+          {tabs.map(tab => {
+            const isSelected = selectedTab.id === tab.id
+            return (
+              <div
+                role="tabpanel"
+                className="kirk-tab-panel"
+                id={`${generateTabPanelId(tab)}`}
```
wakooka: nit
```suggestion
id={generateTabPanelId(tab)}
```

src/tabs/index.tsx
```diff
@@ -10,7 +10,8 @@ const StyledTabs = styled(Tabs)`
     position: relative;
   }
 
-  & .kirk-tablist-wrapper {
+  & .kirk-tabs {
+    margin-bottom: 16px;
```
Tuxosaurus: ${space.l}

src/tabs/Tabs.tsx
```diff
@@ -227,22 +228,24 @@ export class Tabs extends PureComponent<TabsProps, TabsState> {
...
-              key={tab.id}
-              aria-labelledby={tab.id}
-              hidden={!isSelected}
-            >
-              {isSelected ? tab.panelContent : null}
-            </div>
-          )
-        })}
-      </Fragment>
+        <div className="kirk-panels-section" role="presentation">
```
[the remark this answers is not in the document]
grosbouddha: It might for some screenreaders. I will add a chapter item to speak about that topic.

src/tabs/Tabs.tsx
grosbouddha: Done, thanks!

src/tabs/index.tsx
grosbouddha: Thanks!

PR #556, opened 2020-02-14
Bbc 6764 create lock icon

nickoola:


nickoola:
Capture d’écran 2020-02-14 à 18 43 23

pith:
Here is what we expect in the specs 
image

nickoola:
Here is the render after using the proper icon:
Capture d’écran 2020-02-17 à 14 46 11

src/datePicker/__snapshots__/DatePicker.unit.tsx.snap
```diff
@@ -40,7 +40,7 @@ exports[`DatePicker renderNavbar Should render the previous/next buttons in hori
   display: -webkit-inline-flex;
   display: -ms-inline-flexbox;
   display: inline-flex;
-  padding: 16px 24px;
+  padding: 0 24px;
```
gabrielagabriel: **[no magic numbers]**  Setting paddings according to brand spaces, eg: `space.x` to inforce typography grid

Tuxosaurus: This is in the snapshot, it's fine in index.tsx :)

src/button/index.tsx
```diff
@@ -19,21 +19,26 @@ const StyledButton = styled(Button)`
...
   }
 
   & svg,
   & img {
     flex-shrink: 0;
   }
 
+  &:not(.kirk-button-bubble) svg {
+    margin-right: ${space.s};
+  }
```
wakooka: I don't think we should do this :/ having exceptions inside a button component based on what icon is inside is defeating the purpose of normalizing styles.

Bidjii:
@loick for review plz, he is the one concerned 😬

loick:
Can you stop doing typo on my name please?

nickoola:
Here is an update ()visually validated by @wakooka  using our brand spacer (Pixar specs are not precise for now in this case):
Capture d’écran 2020-02-26 à 13 22 59

src/button/index.tsx
nickoola: I inverted the css logic ;-)

PR #559, opened 2020-02-17
[BBC-5337] Upgrade `Modal` with `TheVoice` component

gabrielagabriel:
## Description
Update **Modal** component to receive `TheVoice`

## What has been done

- `TheVoice` component now receive a prop `isInverted` to define its color `title` or `textWithBackground`. It receives a modifier class `.kirk-title--inverted`.
- `ConfirmationModal` and `SuccessModal` now use `TheVoice` component
- Move around **Modals stories** (SuccessModal, ConfirmationModal) to improve discoverability/heuristics. _We could also consider to move folders_

## Warnings  / Things to pay attention to
- I've declare `children` as `string` since we had only text as child... **I don't feel its a good practice.. What do you think?!**

## Last considerations
It's my first PR, so please be kind! 🗡

### DIFF
--- a/src/confirmationModal/ConfirmationModal.tsx
+++ b/src/confirmationModal/ConfirmationModal.tsx
@@ -5,6 +5,7 @@ import { color } from '_utils/branding'
 import Button, { ButtonStatus } from 'button'
 import Modal, { ModalSize } from 'modal'
 import { ModalProps } from 'modal/Modal'
+import TheVoice from 'theVoice'
 import CrossIcon from 'icon/crossIcon'
 import WarningIcon from 'icon/warningIcon'
 import InfoIcon from 'icon/infoIcon'
@@ -26,6 +27,7 @@ export interface ConfirmationModalProps extends ModalProps {
   readonly confirmLabel?: string
   readonly confirmIsLoading?: boolean
   readonly size?: ModalSize
+  readonly children?: string
 }
 
 class ConfirmationModal extends Component<ConfirmationModalProps> {
@@ -98,7 +100,9 @@ class ConfirmationModal extends Component<ConfirmationModalProps> {
       >
         <div className={`${baseClassName}-dialog`}>
           {getIcon()}
-          <div className={`${baseClassName}-body`}>{children}</div>
+          <div className={`${baseClassName}-body`}>
+            <TheVoice isInverted>{children}</TheVoice>
+          </div>
           <footer className={`${baseClassName}-footer`}>
             {isWarning && (
               <Button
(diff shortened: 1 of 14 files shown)


src/theVoice/specifications/theVoice.md
```diff
@@ -16,17 +15,23 @@ import { TheVoice } from '@blablacar/ui-library'
...
 When screen width is less than 800px:
- - the font size moves down to 22px.
- - the text becomes left-aligned instead of centered
+
+- the font size moves down to 22px.
+- the text becomes left-aligned instead of centered
+
+**COLORS**
+
+It can switch colors when applyed to darker backgrounds.
```
Bidjii: `applyed` > `applied`

Bidjii:
The CHANGELOG update is missing :)

src/confirmationModal/ConfirmationModal.tsx
```diff
@@ -26,6 +27,7 @@ export interface ConfirmationModalProps extends ModalProps {
   readonly confirmLabel?: string
   readonly confirmIsLoading?: boolean
   readonly size?: ModalSize
+  readonly children?: string
```
mcellier: I agree with you, it feels a bit weird to have children as string. You may use another prop instead, i.e. `label`

grosbouddha: What about just 'title' ?

src/successModal/SuccessModal.tsx
```diff
@@ -3,15 +3,15 @@ import React, { Component } from 'react'
...
-import { color } from '_utils/branding'
+import TheVoice from 'theVoice'
 import { assertModalSizes } from '_utils/assert'
 import uuidv4 from 'uuid/v4'
 
 export interface SuccessModalProps extends ModalProps {
   readonly confirmLabel?: string
   readonly imageSrc: string
   readonly imageText?: string
+  readonly children?: string
```
grosbouddha: Same here, title should be more semantically correct.

src/confirmationModal/ConfirmationModal.tsx
gabrielagabriel: @mcellier @grosbouddha To avoid break changes on a "good first issue", I would rather keep the `TheVoice` API and only change `children` type to `ReactNode`

Check it out:
https://medium.com/@martin_hotell/react-children-composition-patterns-with-typescript-56dfc8923c64

It sounds a good approach since `TheVoice` could also receive `<br />`  for exemple.

I've also updated `Title` component `children` type, since it has been declared as `string` before.

PR #563, opened 2020-02-19
New MainContent component to handle layout vertical composition

PerrineBoissieres:
### New components: `MainContent`, `Content` and `BottomContent`

These are used for handling full screen pages with bottom-sticky content. `height: stretch` takes care of the first render, and then Javascript takes place by forcing layout recalculations when necessary (on screen resize, phone rotation or keyboard opening/closing on mobile).

Capture d’écran 2020-02-19 à 09 50 43

@grosbouddha I've used some sections to recreate the login page as a demo, but a few things are off (no wrapper for buttons yet, and inputs margins as well). Are those expected, or did I miss something with the new sections?


src/layout/content/content.tsx
```diff
@@ -0,0 +1,67 @@
+import React, { Component } from 'react'
+
+export interface MainContentProps {
+  readonly children: React.ReactNode
+  readonly topBarSelector?: string
+  readonly topBarHeight?: string
```
grosbouddha: Can we derive the height from the topBarSelector above ?
If we need to keep this prop, can we document what expected in the string param (I guess something like '24px')

PerrineBoissieres: We need this prop for when extracting the CSS at build (for SSR for example) as the topbar isn't mounted at this moment.
I'll had a comment :)

grosbouddha:
> @grosbouddha I've used some sections to recreate the login page as a demo, but a few things are off (no wrapper for buttons yet, and inputs margins as well). Are those expected, or did I miss something with the new sections?

The vertical spacing is not managed by the section system. For vertical spacing, each component should have intrinsic margins/paddings (to be added) and can use BlankSeparator and Divider widgets (inside parent components).
Thanks for adding the login pages that will help me to normalize the vertical spacing :pray:!

PerrineBoissieres:
> > @grosbouddha I've used some sections to recreate the login page as a demo, but a few things are off (no wrapper for buttons yet, and inputs margins as well). Are those expected, or did I miss something with the new sections?
> 
> The vertical spacing is not managed by the section system. For vertical spacing, each component should have intrinsic margins/paddings (to be added) and can use BlankSeparator and Divider widgets (inside parent components).
> Thanks for adding the login pages that will help me to normalize the vertical spacing 🙏!

Ok ^^
A small remark on this: if you can, use paddings instead of margins. For example, the external margin on TheVoice can create some layout bugs (margin collapse) if it's the first component in the page, and there are no paddings on its container :)

src/layout/content/content.tsx
```diff
@@ -0,0 +1,67 @@
...
+    const contentElement = element
+    if (!contentElement) {
+      return null
+    }
+    const topBar: HTMLElement = document.querySelector(this.props.topBarSelector)
+    const height = topBar
+      ? Math.floor(window.innerHeight - topBar.offsetHeight)
+      : window.innerHeight
+
+    contentElement.style.minHeight = `${height}px`
```
gabrielagabriel: Does `px` will be transformed to another accessible unit `rem, em...` after that?

PerrineBoissieres: As much as I love relative sizing for content and spacing, here I need to have the container being at minimum the height of the device's screen, so pixel values are needed for precision :)

PR #566, opened 2020-02-20
Add yourrides, ride history and SRP pages in kirk

grosbouddha:
image
image
image


Tuxosaurus:
Is it expected that there's no margin between cards?

grosbouddha:
> Is it expected that there's no margin between cards?

Yes, they are not normalized yet. If I add margin here, I will break margins in kairos.

PR #569, opened 2020-02-21
[BOYSCOUT] Update Storybook dependencies

gabrielagabriel:
## Description
Updating **Storybook** to the latest version `5.3.13`

## What has been done
Removed `addon-options` since it has been deprecated on version `5.0`
Reorder tabs  💅


package.json
```diff
@@ -28,7 +28,7 @@
     "storybook-build": "export NODE_PATH=src && build-storybook -c .storybook -o storybook-build",
     "storybook-deploy": "npm run storybook-build && gh-pages -d storybook-build"
   },
-  "author": "Blablacar front-end team",
+  "author": "BlaBlaCar front-end team",
```
mcellier: 🥇

package.json
```diff
@@ -73,6 +73,7 @@
     "babel-loader": "8.0.6",
     "babel-plugin-module-resolver": "3.2.0",
     "babel-plugin-typescript-to-proptypes": "0.17.1",
+    "babel-preset-react-app": "9.1.1",
```
mcellier: Why this change?

gabrielagabriel: Oh! My mistake. I will remove it from this PR since it will be needed for adding `addons-doc` with typescript instead.

https://www.npmjs.com/package/@storybook/addon-docs

Good catch! 🧐

PR #570, opened 2020-02-21
add SearchRecap + update UneditableTextField classNames

Tuxosaurus:
Added 'kirk-' prefix to classNames in `UneditableTextField`
Added `SearchRecap`, made it compliant with Pixar rules, and added `large` prop to deal with larger devices (desktop experience wasn't designed, I had this checked with designers).

kap_search_recap


src/searchRecap/story.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import { storiesOf } from '@storybook/react'
+import { withKnobs, boolean, text } from '@storybook/addon-knobs'
+import SearchRecap from './index'
+
+const stories = storiesOf('Widgets|SearchRecap', module)
+stories.addDecorator(withKnobs)
+
+stories.add('default', () => (
+  <SearchRecap
```
grosbouddha: Wrap in a <Section /> ?

Tuxosaurus: Done, also included in the SearchResults page (Tabs story)

Tuxosaurus:
> Thanks for this 👍 We should remove the `large` prop and handle it directly from the component itself, but not blocker can be done later.


> Thanks for this 👍 We should remove the `large` prop and handle it directly from the component itself, but not blocker can be done later.

Right, I've made the change

PR #571, opened 2020-02-25
[BBC-1272] Define Typography Components on top of `Text`

gabrielagabriel:
## Description

Create *TextComponents* on top a common component to **avoid redefinitions** on the application

## What has been done
- Create components by overriding a `TextComponent` definition with **styled components**
-- `TextDisplay2`
-- `TextDisplay1`
-- `TextSubHeader`
-- `TextSubHeaderStrong`
-- `TextTitle`
-- `TextTitleStrong`
-- `TextBody`
-- `TextBodyStrong`
-- `TextCaption`

- General documentation with **addon-readme**  and follow pattern `/specification/componentName`

## Considerations / Things to pay attention to
**- This PR is a conversation about how to use these new components.** 
IMHO given _semantics_ to its parent its gonna make things harder to fix spaces, for _titles_ for instance. We can notice that on the story itself

**- isCTA Rule**
I've not worked on isCTA rule since is mostly applied to `Buttons` and I'd like to live it there

## TODO
[x] **Edge Cases**
~~On SketchCloud File, PIXAR has an edge case for small types... I don't know if is only applied to `Body`, I'm going to check to  keep going~~ => Edge Cases will be reviewed by Design Team later on.


grosbouddha:
Can we squash to ease the review ?

src/typography/story.tsx
```diff
@@ -0,0 +1,146 @@
...
+      <p>
+        <TextButton>
+          Medium 16px <br /> Height 20px
+        </TextButton>
+      </p> */}
+    </Section>
+  </Fragment>
+))
+
+// stories.add('Display2', () => (
```
grosbouddha: remove ?

src/typography/display2/TextDisplay2.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { ReactNode } from 'react'
+
+import Text, { TextDisplayType, TextTagType } from 'text'
+
+export interface TextBodyProps {
+  readonly className?: Classcat.Class
+  readonly children: string | ReactNode
+  readonly isInverted?: boolean
+  readonly role?: string
```
grosbouddha: If we consider these as TextNodes, they should have no role or aria labels IMO

src/typography/display2/TextDisplay2.tsx
```diff
@@ -0,0 +1,22 @@
...
+  readonly ariaLabel?: string
+}
+
+const TextDiplay2 = ({ children, className, isInverted = false, ...props }: TextBodyProps) => {
+  const classes = isInverted ? `kirk-text--inverse ${className}` : className
+  return (
+    <Text display={TextDisplayType.DISPLAY2} tag={TextTagType.SPAN} className={classes} {...props}>
```
grosbouddha: Why do we use Text here a simple <span> with the good styles should be enough ?
I think Ideally we should get rid of Text.

grosbouddha:
Having both:
- Create TextDisplay2 component.
- Started documentation with addon-readme and follow pattern /specification/componentName
is too much IMO for one PR.
I think we should keep PRs focused on one clear goal. 

In particular, I believe the second change deserve a chapter discussion to get team approval.

grosbouddha:
Having both:
- Create TextDisplay2 component.
- Started documentation with addon-readme and follow pattern /specification/componentName
is too much IMO for one PR.
I think we should keep PRs focused on one clear goal. 

In particular, I believe the second change deserve a chapter discussion to get team approval.

.storybook/config.js
```diff
@@ -20,6 +20,7 @@ addParameters({
   },
   readme: {
     sidebar: props,
+    codeTheme: 'xonokai',
```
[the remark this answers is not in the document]
gabrielagabriel: Configuration on the `ReadMe` pluging to highlighting code (just changing colors)

src/typography/story.tsx
gabrielagabriel: This is a WIP, so no "dead" code yet

src/typography/display2/TextDisplay2.tsx
gabrielagabriel: To do not introduce a hard break change, we could update that "later" on each component

gabrielagabriel:
@grosbouddha 

> Started documentation with addon-readme and follow pattern /specification/componentName
> is too much IMO for one PR.
> I think we should keep PRs focused on one clear goal.
> In particular, I believe the second change deserve a chapter discussion to get team approval.

I'm not introducing _"anything new"_,  I'm just _documenting_ the New component with the same pluggin that were already available on Kirk. I've followed the plugins documentation to do so... https://github.com/tuchk4/storybook-readme


Yes, I want to move from 'addon-readme' to 'addon-docs'  (official from Storybook - https://github.com/storybookjs/storybook/tree/master/addons/docs) but this is going to be on another PR, 'cos it brings A LOT of changes that will be discussed in the chapter and its really not the point here.

src/typography/display2/TextDisplay2.tsx
wakooka: You don't need to delete the `Text` component, however not using it here will save you time when we'll remove it

src/typography/display2/specifications/display2.md
```diff
@@ -0,0 +1,34 @@
...
+<h1>
+  <TextDisplay2>The quick brown fox jumps over the lazy dog</TextDisplay2>
+</h1>
+```
+
+## Props
+
+| **Props**    | **Value**       | **Description**                  |
+| ------------ | --------------- | -------------------------------- |
+| `isInverted` | `true`\|`false` | When applied to dark backgrounds |
```
wakooka: Thanks! This is nice, I'm ok to go with that for now, but we should probably think on having a documentation template so that all the documentation section are in the same order. Also I'm not a big fan of having `Props` section since we're using typescript it's easy enough to use it, and if it's not we should look into a way to generate it automatically in the docs otherwise it will get out of date really quick.

pith: A generated list could be nice, but we really need these descriptions for each props. Thanks @gabrielagabriel for adding this documentation.

src/typography/story.tsx
grosbouddha: WIP code should be removed.

PR #573, opened 2020-02-27
New IllustratedSection component

PerrineBoissieres:
New section to handle responsive illustrated content.

### Mobile
Capture d’écran 2020-02-27 à 10 58 29

### Desktop
Capture d’écran 2020-02-27 à 10 58 51


src/layout/section/illustratedSection/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+import IllustratedSection from './illustratedSection'
+
+const StyledIllustratedSection = styled(IllustratedSection)`
+  & .section-content {
+    display: flex;
+    flex-direction: column;
+  }
+
+  & .kirk-illustratedSection-illustration {
+    margin-left: -${space.xl};
```
grosbouddha: Why do we need these negative margins ?
Does it mean the illustration is touching the viewport edges in mobile mode ?

I would probably remove these to make it simpler (and the missing padding probably does not really matter)

pith:
I'm sure the illustration is properly centered on desktop see what we do in modals

image

mcellier:
> I'm not sure the illustration is properly centered on desktop see what we do in modals
> 
> image

The thing is that we're not centering the content vertically. We may double check with Pantea but these designs were validated by the Pixar team

src/layout/section/illustratedSection/illustratedSection.tsx
```diff
@@ -0,0 +1,26 @@
...
+  readonly className?: Classcat.Class
+  readonly illustrationUrl: string
+  readonly illustrationAlt?: string
+}
+
+/**
+ * A specialized section which show some content with an illustration.
+ */
+const IllustratedSection: React.SFC<IllustratedSectionProps> = (props: IllustratedSectionProps) => {
+  const { className, illustrationUrl, illustrationAlt = '' } = props
```
mcellier: `className` has no default but `illustrationAlt` has?

src/layout/section/illustratedSection/index.unit.tsx
```diff
@@ -0,0 +1,16 @@
...
+import renderer from 'react-test-renderer'
+
+import IllustratedSection from './index'
+
+const defaultProps = {
+  illustrationUrl: 'http://blablacar.com/pic.jpg',
+}
+
+describe('IllustratedSection', () => {
+  it('should render default MediaContentSection section', () => {
```
mcellier: copy/paste ? `MediaContentSection`

src/layout/section/illustratedSection/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+    & .kirk-illustratedSection-illustration {
+      margin-left: 0;
+      margin-right: 0;
+      width: 33.33%;
+    }
+
+    & .kirk-illustratedSection-content {
+      margin-left: ${space.xl};
+      width: 66.66%;
```
grosbouddha: Do we want a max-width here ? Otherwise on a 30inch monitor, the itemchoices will look quite weird (text of the right and the chevron will be 50 cm away).
The same for the illustration, we probably don't want to grow indefinitely ?
This can be done iteratively in a future PR if needed.

src/layout/section/illustratedSection/illustratedSection.tsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <BaseSection tagName="article" className={className} contentSize={SectionContentSize.LARGE}>
+      <div className="kirk-illustratedSection-illustration">
+        <img src={illustrationUrl} alt={illustrationAlt} />
```
mcellier: Warning: in the Profile picture step, we're not displaying an illustration, but an `Avatar`. How do you think we can handle this?

src/layout/section/illustratedSection/index.tsx
grosbouddha: Actually you have a wrapping BaseSection so that should work, diimiss my comment :pray:

src/layout/section/illustratedSection/illustratedSection.tsx
grosbouddha: Maybe juste have a `illustrationMedia=string | JSX.Element ` and do different rendering based on type ? And pray to not have more rendering cases :)

PerrineBoissieres: Good point for the Profile Picture case. I'll add support for the Avatar component :)

src/layout/section/illustratedSection/index.tsx
PerrineBoissieres: The illustration touches the borders of the screen on mobile, yes.

PerrineBoissieres:
> I'm not sure the illustration is properly centered on desktop see what we do in modals
> 
> image

Success screen has a different vertical alignment than form steps. For those cases, content is top-aligned, same for illustration.

PR #574, opened 2020-02-27
Datepicker overlay & fullscreen

wakooka:
First step for the new search form layer
image
image

### DIFF
--- a/src/datePicker/index.tsx
+++ b/src/datePicker/index.tsx
@@ -90,10 +90,12 @@ const StyledDatePicker = styled(DatePicker)`
     color: ${color.fadedText};
     cursor: default;
   }
-  & .DayPicker-Day--selected span {
+  & .DayPicker-Day--selected span,
+  & .DayPicker-Day:focus > span {
     background-color: ${color.primary};
     color: ${color.white};
   }
+
   & .DayPicker-Day span {
     display: inline-flex;
     justify-content: center;
--- a/src/icon/calendarIcon.tsx
+++ b/src/icon/calendarIcon.tsx
@@ -1,0 +1,32 @@
+// tslint:disable:max-line-length
+import React from 'react'
+import BaseIcon from '_utils/icon'
+import { BaseIconDefaultProps } from '_utils/icon/BaseIcon'
+
+export const CalendarIcon = (props: Icon) => (
+  <BaseIcon {...props}>
+    <g fill="none" fillRule="evenodd">
+      <path
+        d="M18.087 4.174h3.478v17.391H2.435V4.174h3.478m1.74 0h8.695"
+        stroke={props.iconColor}
+        strokeLinecap="round"
+        strokeLinejoin="round"
+      />
+      <path
+        d="M5.913 2.435h1.74v3.478h-1.74V2.435zm10.435 0h1.739v3.478h-1.74V2.435zM2.435 9.39h19.13-19.13zM5.913 12.87h1.74v.87h-1.74v-.87zm5.217 0h1.74v.87h-1.74v-.87zm-5.217 4.347h1.74v.87h-1.74v-.87zm5.217 0h1.74v.87h-1.74v-.87zm5.218-4.347h1.739v.87h-1.74v-.87z"
+        stroke={props.iconColor}
+        strokeLinecap="round"
+        strokeLinejoin="round"
+      />
+      <path
+        d="M5.913 12.87h1.74v.87h-1.74v-.87zm5.217 0h1.74v.87h-1.74v-.87zm-5.217 4.347h1.74v.87h-1.74v-.87zm5.217 0h1.74v.87h-1.74v-.87zm5.218-4.347h1.739v.87h-1.74v-.87z"
+        fill={props.iconColor}
+        fillRule="nonzero"
+      />
+    </g>
+  </BaseIcon>
+)
+
+CalendarIcon.defaultProps = BaseIconDefaultProps
+
+export default React.memo(CalendarIcon)
--- a/src/searchForm/datePicker/datePickerFullScreen/DatePickerFullScreen.tsx
+++ b/src/searchForm/datePicker/datePickerFullScreen/DatePickerFullScreen.tsx
@@ -1,0 +1,30 @@
+import React from 'react'
+import Datepicker, { DatePickerProps, DatePickerOrientation } from 'datePicker'
+import Divider from 'divider'
+import Section from 'layout/section/baseSection'
+import ChevronIcon from 'icon/chevronIcon'
+import Item from '_utils/item'
+
+export interface DatePickerFullScreenProps extends DatePickerProps {
+  readonly title: string
+  readonly onClick?: (event: React.MouseEvent<HTMLElement>) => void
+}
+
+export const DatePickerFullScreen = (props: DatePickerFullScreenProps) => (
+  <Section className={props.className}>
+    <Item
+      leftAddon={<ChevronIcon left />}
+      leftTitle={props.title}
+      tag={<button type="button" />}
+      onClick={props.onClick}
+    />
+    <Divider />
+    <Datepicker
+      {...props}
+      className="kirk-datePickerFullScreen-datepicker"
+      orientation={DatePickerOrientation.VERTICAL}
+    />
+  </Section>
+)
+
+export default DatePickerFullScreen
--- a/src/searchForm/datePicker/datePickerFullScreen/index.tsx
+++ b/src/searchForm/datePicker/datePickerFullScreen/index.tsx
@@ -1,0 +1,11 @@
+import styled from 'styled-components'
+
+import DatePickerFullScreen from './DatePickerFullScreen'
+
+const StyledDatePickerFullScreen = styled(DatePickerFullScreen)`
+  & {
+    box-sizing: border-box;
+  }
+`
+
+export default StyledDatePickerFullScreen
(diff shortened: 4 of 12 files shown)


src/datePicker/index.tsx
```diff
@@ -90,10 +90,12 @@ const StyledDatePicker = styled(DatePicker)`
     color: ${color.fadedText};
     cursor: default;
   }
-  & .DayPicker-Day--selected span {
+  & .DayPicker-Day--selected span,
+  & .DayPicker-Day:focus span {
```
wakooka: Making the focus state visible

src/searchForm/datePicker/datePickerTopBar/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  & .kirk-datePickerTopBar-text {
+    color: ${color.primaryText};
+    padding-left: ${space.m};
+  }
+
+  & .kirk-datePickerTopBar-button {
+    width: auto;
+    height: auto;
```
wakooka: Will sync with Jonas about this, because we introduced a new type of button which is basically inline and unstyled

stories/index.tsx
```diff
@@ -59,6 +62,7 @@ import '../src/proximity/story'
 import '../src/pushInfo/story'
 import '../src/rating/story'
 import '../src/searchRecap/story'
+import '../src/searchForm/datePicker/story'
```
mcellier: Double import

src/searchForm/datePicker/datePickerFullScreen/index.tsx
```diff
@@ -0,0 +1,11 @@
+import styled from 'styled-components'
+
+import DatePickerFullScreen from './DatePickerFullScreen'
+
+const StyledDatePickerFullScreen = styled(DatePickerFullScreen)`
+  & {
+    box-sizing: border-box;
```
mcellier: @grosbouddha Didn't you want to add this globally?

src/searchForm/datePicker/datePickerOverlay/DatePickerOverlay.tsx
```diff
@@ -0,0 +1,26 @@
...
+import cc from 'classcat'
+import Datepicker, { DatePickerProps, DatePickerOrientation } from 'datePicker'
+import Divider from 'divider'
+import DatePickerTopBar from 'searchForm/datePicker/datePickerTopBar'
+
+export interface DatePickerOverlayProps extends DatePickerProps {
+  readonly title: string
+}
+
+const numberOfMonths = 1
```
mcellier: UPPER_SNAKE_CASE?

src/searchForm/datePicker/datePickerOverlay/DatePickerOverlay.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export const DatePickerOverlay = (props: DatePickerOverlayProps) => (
```
mcellier: This component looks A LOT like the fullscreen one. Can't we fuse both of them by having two sets of default props (vertical, number of months) and a className for css differencies?

src/searchForm/datePicker/datePickerOverlay/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+    width: ${datePickerOverlayWidth};
+    border: 1px solid ${color.border};
+    border-radius: ${radius.l};
+    padding: ${space.xl};
+    /* Same as TripCard, we might need to have box-shadow specs so that we can move it to branding file */
+    box-shadow: 0 1pt 4pt rgba(0, 0, 0, 0.16), 0 2pt 8pt rgba(0, 0, 0, 0.08);
+  }
+
+  & .kirk-datePickerOverlay-datepicker {
+    margin-top: 16px;
```
mcellier: Use var

src/searchForm/datePicker/datePickerFullScreen/index.tsx
wakooka: There's no way to add it globally in the library, however this file is actually useless since I don't set any size, so maybe I should just remove it.

src/icon/calendarIcon.tsx
```diff
@@ -0,0 +1,33 @@
+// tslint:disable:max-line-length
+import React from 'react'
+import BaseIcon from '_utils/icon'
+import { BaseIconDefaultProps } from '_utils/icon/BaseIcon'
+
+export const CalendarIcon = (props: Icon) => (
+  <BaseIcon {...props}>
+    <g fill="none" fillRule="evenodd">
+      <path d="M0 0h24v24H0z" />
```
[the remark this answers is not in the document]
wakooka: Removed

src/searchForm/datePicker/story.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+
+import { storiesOf } from '@storybook/react'
+import { action } from '@storybook/addon-actions'
+import { select, text } from '@storybook/addon-knobs'
+import { testLocales } from 'datePicker/story'
+import DatePickerOverlay from './datePickerOverlay'
+import DatePickerFullScreen from './datePickerFullScreen'
+
+const stories = storiesOf('SearchForm|DatePicker', module)
```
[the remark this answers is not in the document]
wakooka: I removed the story, you're right we shouldn't have stories on internal components. And those should only be used from the Search Form component (to be created)

stories/index.tsx
wakooka: Removed both

src/searchForm/datePicker/datePickerOverlay/DatePickerOverlay.tsx
wakooka: Since they're internal components, the switch would be done on the Search Form component level. We talked with @grosbouddha about it and we thought it's easier to understand to have 2 distincts components, but this is probably challengeable

src/datePicker/story.tsx
```diff
@@ -10,21 +10,21 @@ import readme from 'datePicker/specifications/datepicker.md'
 const stories = storiesOf('Widgets|DatePicker', module)
 stories.addDecorator(withKnobs)
 
-const weekdaysShort = (locale: string): string[] => {
+export const weekdaysShort = (locale: string): string[] => {
```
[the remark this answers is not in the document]
wakooka: Because I needed it for the stories of DatePickerOverlay. But I actually remove the stories since they're internal components.

src/searchForm/datePicker/datePickerOverlay/DatePickerOverlay.tsx
grosbouddha: It might useful as we add more differences in the future (like the datepicker might need JS based positioning). For now, I agree, it looks a bit weird to have 2 widgets.

src/searchForm/datePicker/datePickerTopBar/DatePickerTopBar.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import CalendarIcon from 'icon/calendarIcon'
+import Text from 'text/Text'
+import Button, { ButtonStatus } from 'button'
+import cc from 'classcat'
+import ChevronIcon from 'icon/chevronIcon'
+
+export interface DatePickerTopBarProps {
```
grosbouddha: I don't see the value of having this widget, we have two completely different behavior/rendering merged into one place. I would just dispatch each top bars in overlay and fullscreen tsx

src/searchForm/datePicker/datePickerTopBar/DatePickerTopBar.tsx
```diff
@@ -0,0 +1,27 @@
...
+  readonly title: string
+  readonly className?: Classcat.Class
+  readonly isClickable?: boolean
+}
+
+export const DatePickerTopBar = (props: DatePickerTopBarProps) => (
+  <div className={cc(['kirk-datePickerTopBar-row', props.className])}>
```
grosbouddha: why is it not an kirk item with a left addon  ?

grosbouddha:
Not a big fan of the **searchForm** namespace, it's just datepicker overlay/fullscreen, nothing hardocded for a search flow.

src/datePicker/index.tsx
wakooka: Hmm not really I don't control this part, we use react-day-picker

src/searchForm/datePicker/datePickerFullScreen/DatePickerFullScreen.tsx
```diff
@@ -0,0 +1,23 @@
...
+import Datepicker, { DatePickerProps, DatePickerOrientation } from 'datePicker'
+import Divider from 'divider'
+import DatePickerTopBar from 'searchForm/datePicker/datePickerTopBar'
+
+export interface DatePickerFullScreenProps extends DatePickerProps {
+  readonly title: string
+}
+
+export const DatePickerFullScreen = (props: DatePickerFullScreenProps) => (
+  <div className={cc(['kirk-datePickerFullScreen', props.className])}>
```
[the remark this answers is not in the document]
grosbouddha: Actually, we have deeper problem here, the name *FullScreen is not supported by an implementation. IIUC, the fullscreen behavior will depend on the call site, meaning in kairos which makes me sad.

src/datePicker/index.tsx
grosbouddha: let's use at least the > selector to make the rule a bit less lax.

src/searchForm/datePicker/datePickerTopBar/DatePickerTopBar.tsx
wakooka: It's an Item now

wakooka:
> Not a big fan of the **searchForm** namespace, it's just datepicker overlay/fullscreen, nothing hardocded for a search flow.

As discussed those components will only be used from the upcoming SearchForm component (for now). If later on, we want to reuse them from somewhere we can move them out of this namespace.

@grosbouddha PTAL

PR #578, opened 2020-02-28
[BBC-6919] Fix Datepicker on small screen.

Sevrahk:
Previously if we selected date in few months and opened the datepicker again. The first available month was the one that have been selected and it was impossible to scroll up to the current month.

### DIFF
--- a/src/datePicker/DatePicker.tsx
+++ b/src/datePicker/DatePicker.tsx
@@ -30,6 +30,7 @@ export interface DatePickerProps {
   readonly months?: string[]
   readonly onChange?: (obj: OnChangeParameters) => void
   readonly initialDate?: Date
+  readonly initialMonth?: Date
   readonly className?: Classcat.Class
   readonly numberOfMonths?: number
   readonly orientation?: DatePickerOrientation
@@ -51,6 +52,7 @@ export class DatePicker extends PureComponent<DatePickerProps, DatePickerState> 
     months: defaultMonths,
     numberOfMonths: 2,
     initialDate: null,
+    initialMonth: new Date(),
     isOutsideRange: day => DayPicker.DateUtils.isDayBefore(day, new Date()),
     orientation: DatePickerOrientation.HORIZONTAL,
     fromMonth: new Date(),
@@ -61,6 +63,8 @@ export class DatePicker extends PureComponent<DatePickerProps, DatePickerState> 
   state = {
     date: this.props.initialDate,
   }
+
+  dayPickerContainer = React.createRef<HTMLDivElement>()
 
   formatMonthTitle = (date: Date): string => {
     const currentYear = `${new Date().getFullYear()}`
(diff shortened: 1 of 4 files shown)


src/datePicker/DatePicker.tsx
```diff
@@ -181,7 +182,7 @@ export class DatePicker extends PureComponent<DatePickerProps, DatePickerState>
           months={months}
           firstDayOfWeek={firstDayOfWeek}
           localeUtils={{ ...DayPicker.LocaleUtils, formatMonthTitle: this.formatMonthTitle }}
-          initialMonth={initialDate}
+          initialMonth={initialMonth}
```
mcellier: Is the view scrolled to the selected date?

src/datePicker/DatePicker.tsx
```diff
@@ -157,6 +157,7 @@ export class DatePicker extends PureComponent<DatePickerProps, DatePickerState>
     } = this.props
     const { date } = this.state
     const layoutClassName = `months-grid-${numberOfMonths}`
+    const initialMonth = orientation === DatePickerOrientation.VERTICAL ? new Date() : initialDate
```
mcellier: Add unit test

wakooka: Could we just pass initialMonth as a nullable prop instead?
I think this logic shoud be handled in SPA, I'm not a fan of having a `new Date()`
Also how does that work with the `fromMonth` prop, can they conflict?

src/datePicker/DatePicker.tsx
Sevrahk: The view is not scrolled

src/datePicker/DatePicker.tsx
Sevrahk: Ok for me, I will directly handle this logic in SPA.

`fromMonth` refers to the first allowed month that users are allowed to navigate or interact. 
[fromMonth documentation](http://react-day-picker.js.org/api/DayPicker/#fromMonth)

`initialMonth` refers to the selected month that will be displayed in the calendar at first render.
[initialMonth documentation](http://react-day-picker.js.org/api/DayPicker/#initialMonth)

src/datePicker/DatePicker.tsx
Sevrahk: The scroll that displays the selected month we be added in the next commit

src/datePicker/DatePicker.unit.tsx
```diff
@@ -64,19 +64,45 @@ describe('DatePicker', () => {
...
+          name="datepicker"
+          initialDate={initialDate}
+          initialMonth={initialDate}
+          numberOfMonths={12}
+        />,
+      )
+
+      expect(wrapper.find(DayPicker).prop('numberOfMonths')).toEqual(12)
+      expect(wrapper.find(DayPicker).prop('selectedDays')).toEqual(initialDate)
+      expect(wrapper.find(DayPicker).prop('initialMonth')).toEqual(initialDate)
```
wakooka: Can you add a test on window.scrollTo. You would need to mock:
```
global.scrollTo = jest.fn()
expect(global.scrollTo).toHaveBeenCalled()
```
If you want you can mock the offsetTop so that you can even have a test like this:
```
expect(global.scrollTo).toHaveBeenCalledWith(0, topMockValue)
```

src/datePicker/DatePicker.unit.tsx
```diff
@@ -64,19 +64,59 @@ describe('DatePicker', () => {
...
+    })
+
+    it('Should forward only selectedDays and keep default initialMonth props to the DayPicker component', () => {
+      const offsetTop = 1000
+      const stickyPositionTop = 50
+      window.scrollTo = jest.fn()
```
wakooka: Make sure to reset this mock.

PR #581, opened 2020-03-05
Add aria label to ItemInfo, and aria-hidden to Item

florianeyzat:
Capture d’écran 2020-03-05 à 15 16 59

### DIFF
--- a/src/itemInfo/index.tsx
+++ b/src/itemInfo/index.tsx
@@ -8,15 +8,17 @@ export interface ItemInfoProps {
   readonly icon?: React.ReactNode
   readonly mainTitle?: string
   readonly tag?: JSX.Element
+  readonly ariaLabel?: string
 }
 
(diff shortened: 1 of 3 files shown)


src/itemInfo/index.tsx
```diff
@@ -8,15 +8,18 @@ export interface ItemInfoProps {
...
-const ItemInfo = ({ mainInfo, className, mainTitle, icon, tag }: ItemInfoProps) => (
+const ItemInfo = ({ mainInfo, className, mainTitle, icon, tag, ariaLabel }: ItemInfoProps) => (
   <Item
     className={className}
     leftTitle={mainTitle}
     leftBody={mainInfo}
     leftAddon={icon}
     tag={tag}
+    ariaLabel={ariaLabel}
+    ariaHidden={!!ariaLabel}
```
nickoola: We should also be able to have an `aria-hidden="false"` with an `aria-label`.
If the `aria-label` is complementary and not a summary.

PR #582, opened 2020-03-05
improve tripCard layout and how text truncates

Tuxosaurus:
Allow long text in Itinerary to be truncated, and make price always visible.
Snaps the driver/passenger/flags info at the bottom of the card to align with adjacent cards.

The bottom part has also been refactored to remove cases of empty tags and also truncate the driver name if needed. I also made the name use all the space to the right in case there's no flag to display. If for some reason we have no driver/passenger to display, flags, if any, will stick to the bottom right position.

TESTED = locally on storybook and kairos

Screen Shot 2020-03-05 at 16 14 34

Screen Shot 2020-03-05 at 16 16 21

Screen Shot 2020-03-06 at 15 30 03

### DIFF
--- a/src/_utils/itineraryLocation/index.tsx
+++ b/src/_utils/itineraryLocation/index.tsx
@@ -17,7 +17,6 @@ const StyledItineraryLocation = styled(ItineraryLocation)`
   & .kirk-itineraryLocation-wrapper {
     display: flex;
     padding: ${space.m} 0;
-    width: 100%;
   }
 
   & a.kirk-itineraryLocation-wrapper {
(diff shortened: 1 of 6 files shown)


src/_utils/itineraryLocation/index.tsx
```diff
@@ -61,6 +60,12 @@ const StyledItineraryLocation = styled(ItineraryLocation)`
 
   & .kirk-itineraryLocation-label {
     flex: 1;
+    min-width: 0; /* reset flex default to allow shrinking below content width */
+  }
+
+  & .kirk-itineraryLocation-label > p {
```
wakooka: Can we add a class to target the element?

Tuxosaurus: Done

Tuxosaurus:
NEW:
The bottom part as also been refactored to remove cases of empty tags and also truncate the driver name if needed. I also made the name use all the space to the right in case there's no flag to display. If for some reason we have no driver/passenger to display, flags, if any, will stick to the bottom right position.

See last screenshot in description, showcasing almost all changes

src/layout/section/tripsSection/index.tsx
```diff
@@ -24,6 +24,8 @@ const StyledTripsSection = styled(TripsSection)`
 
   & .kirk-tripsSection .kirk-tripCard {
     scroll-snap-align: start;
+    max-width: 50%;
```
Tuxosaurus: Need to remove or restrict this rule. We can have only one TopTripCard (cheapest & closest) that takes 100%

PR #584, opened 2020-03-06
Improve a11y on stepper for keyboard navigation and screen reader

wakooka:
Tested with Voiceover.

### DIFF
--- a/src/stepper/Stepper.tsx
+++ b/src/stepper/Stepper.tsx
@@ -160,29 +160,12 @@ export default class Stepper extends PureComponent<StepperProps, StepperState> {
   createButtonListeners(callback: () => void) {
     return isTouchScreen
       ? { onTouchStart: this.handleButtonDown(callback), onTouchEnd: this.handleButtonUp(callback) }
-      : { onMouseDown: this.handleButtonDown(callback), onMouseUp: this.handleButtonUp(callback) }
-  }
-
-  // Keyboard events used on <input type="range">
-  // Cannot be used on buttons since they can be "disabled" and thus unresponsive to events
-  // Note: ideally this would trigger a visually different active state on controls for feedback
-  handleKeyDown = (e: React.KeyboardEvent<HTMLButtonElement>) => {
-    switch (e.key) {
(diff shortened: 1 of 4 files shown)


src/stepper/Stepper.tsx
```diff
@@ -203,60 +185,39 @@ export default class Stepper extends PureComponent<StepperProps, StepperState> {
...
           disabled={isMin}
           isBubble
           {...this.createButtonListeners(this.decrement)}
         >
-          <MinusIcon
-            title={decreaseLabel}
-            iconColor={isMin ? color.disabled : color.primary}
-            size={buttonSize}
-          />
+          <MinusIcon iconColor={isMin ? color.disabled : color.primary} size={buttonSize} />
```
nickoola: Does it change the screen reader output if `decreaseLabel ` is passed to the icon ?

src/stepper/Stepper.tsx
```diff
@@ -203,60 +185,39 @@ export default class Stepper extends PureComponent<StepperProps, StepperState> {
...
-          defaultValue={`${this.state.value}`}
-          step={this.props.step}
-          aria-valuemin={this.props.min}
-          aria-valuemax={this.props.max}
-          aria-valuenow={this.state.value}
-          onKeyDown={this.handleKeyDown}
-        />
+      <div
+        className={cc(['kirk-stepper', `kirk-stepper-${display}`, className])}
+        aria-label={children}
```
grosbouddha: Does this aria label on this div ever get announced ?

src/stepper/Stepper.tsx
wakooka: Not much change, only a minor one. When you have the focus on the button in browsing mode you could still go down to the icon to read the title. Now it says "No title to interact with" but it's reading the button label correctly. Also in code, I feel it makes more sense to put the label on the Button since this is element which contains the action logic.

src/stepper/Stepper.tsx
wakooka: Yes in browsing mode, it focuses the div and read the label.

PR #593, opened 2020-03-10
React context for media size

wakooka:
POC on react context to handle media size

### DIFF
--- a/src/_utils/mediaSizeProvider/MediaSizeProvider.tsx
+++ b/src/_utils/mediaSizeProvider/MediaSizeProvider.tsx
@@ -1,0 +1,35 @@
+import React, { useEffect, useState, ReactNode } from 'react'
+import debounce from 'lodash.debounce'
+import { responsiveBreakpoints } from '../branding'
+
+export enum MediaSize {
+  SMALL = 'small',
+  LARGE = 'large',
+}
+
+interface MediaSizeProviderProps {
+  children: ReactNode
+}
+
+export const MediaSizeContext = React.createContext(MediaSize.SMALL)
+
+const DEBOUNCE_VALUE = 500
+
+export const MediaSizeProvider = ({ children }: MediaSizeProviderProps) => {
+  const [mediaSize, setMediaSize] = useState(MediaSize.SMALL)
+
+  const handleResize = () => {
+    const isSmall = window.innerWidth < parseInt(responsiveBreakpoints.small, 10)
+    setMediaSize(isSmall ? MediaSize.SMALL : MediaSize.LARGE)
+  }
+
+  useEffect(() => {
+    handleResize()
+    window.addEventListener('resize', debounce(handleResize, DEBOUNCE_VALUE))
+    return () => window.removeEventListener('resize', handleResize)
+  }, [])
+
+  return <MediaSizeContext.Provider value={mediaSize}>{children}</MediaSizeContext.Provider>
+}
+
+export default MediaSizeProvider
--- a/src/_utils/mediaSizeProvider/MediaSizeProvider.unit.tsx
+++ b/src/_utils/mediaSizeProvider/MediaSizeProvider.unit.tsx
@@ -1,0 +1,42 @@
+import React, { useContext } from 'react'
+import { mount } from 'enzyme'
+import MediaSizeProvider, { MediaSizeContext } from './MediaSizeProvider'
+
+const mockWindowInnerWidth = value => {
+  Object.defineProperty(window, 'innerWidth', {
+    writable: true,
+    configurable: true,
+    value,
+  })
+}
+
+let ChildComponent
+describe('MediaSizeProvider', () => {
+  beforeEach(() => {
+    ChildComponent = () => {
+      const value = useContext(MediaSizeContext)
+      return <div>{value}</div>
+    }
+  })
+  it('Should provide "small" value when screen is under `responsiveBreakpoints.small`', () => {
+    mockWindowInnerWidth(300)
+    const wrapper = mount(
+      <MediaSizeProvider>
+        <ChildComponent />
+      </MediaSizeProvider>,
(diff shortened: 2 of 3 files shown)


grosbouddha:
👍

src/successModal/SuccessModal.tsx
```diff
@@ -46,11 +47,15 @@ class SuccessModal extends Component<SuccessModalProps> {
     const baseClassName = 'kirk-successModal'
     const successContentId = `${baseClassName}-bodyItem-${uuidv4()}`
 
+    // Without hooks it's not that great because you can only have one contextType
```
pith: Ah good to know

src/badge/Badge.tsx
```diff
@@ -12,6 +13,8 @@ const Badge = ({ className, children, ariaLabel }: BadgeProps) => {
   if (isEmpty(children)) {
     return null
   }
+  const value = useContext(MediaSizeContext)
+  console.log(value) // ouput correctly large / small
```
gabrielagabriel: maybe `console.info` instead

wakooka: Yes I'm cleaning the PR I'll be removing it, it was just for me :)

src/successModal/SuccessModal.tsx
```diff
@@ -46,11 +47,15 @@ class SuccessModal extends Component<SuccessModalProps> {
...
+    console.log(this.context)
+    const successModalSize = this.context === MediaSize.SMALL ? ModalSize.SMALL : ModalSize.LARGE
```
gabrielagabriel: It's there possible to export MediaSize as a "value"  to map directly to the size desired

something like:
```javascript
const successModalSize = ModalSize[MediaSize]
```

wakooka: Good point, I think it's possible. I removed this for now because I just want to merge the Provider with no components using it. Then we can think of an approach to remove the `size` prop from some component and use context instead

src/_utils/mediaSizeProvider/MediaSizeProvider.tsx
```diff
@@ -0,0 +1,33 @@
...
+import debounce from 'lodash.debounce'
+import { responsiveBreakpoints } from '../branding'
+
+export enum MediaSize {
+  SMALL = 'small',
+  LARGE = 'large',
+}
+
+interface MediaSizeProviderProps {
+  children: JSX.Element
```
grosbouddha: ReactNode

src/_utils/mediaSizeProvider/MediaSizeProvider.tsx
```diff
@@ -0,0 +1,33 @@
...
+    const isSmall = window.innerWidth < parseInt(responsiveBreakpoints.small, 10)
+    setMediaSize(isSmall ? MediaSize.SMALL : MediaSize.LARGE)
+  }
+
+  useEffect(() => {
+    handleResize()
+    window.addEventListener('resize', debounce(handleResize, 500))
+    return () => window.removeEventListener('resize', handleResize)
+  }, [])
+  // console.log(mediaSize)
```
grosbouddha: remove

src/_utils/mediaSizeProvider/MediaSizeProvider.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const [mediaSize, setMediaSize] = useState(MediaSize.SMALL)
+
+  const handleResize = () => {
+    const isSmall = window.innerWidth < parseInt(responsiveBreakpoints.small, 10)
+    setMediaSize(isSmall ? MediaSize.SMALL : MediaSize.LARGE)
+  }
+
+  useEffect(() => {
+    handleResize()
+    window.addEventListener('resize', debounce(handleResize, 500))
```
grosbouddha: Extract 500 in a module private const at the top for readability ?

src/_utils/mediaSizeProvider/MediaSizeProvider.unit.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
+import { shallow } from 'enzyme'
+import MediaSizeProvider, { MediaSizeContext } from './MediaSizeProvider'
+
+const mockWindowInnerWidth = value => {
+  Object.defineProperty(window, 'innerWidth', {
```
pith: nit: did you try with `jest.spyOn(window, 'innerWidth').mockReturnValue({ ... })`

Then it's easy to reset

```
  let spy
  beforeEach(() => {
    spy = jest.spyOn(...
  })

  afterEach(() => {
    if (spy) {
      spy.mockReset()
      spy.mockRestore()
    }
  })
```

src/_utils/mediaSizeProvider/MediaSizeProvider.unit.tsx
```diff
@@ -0,0 +1,35 @@
...
+    const wrapper = shallow(
+      <MediaSizeProvider>
+        <div />
+      </MediaSizeProvider>,
+    )
+
+    expect(wrapper.find(MediaSizeContext.Provider).prop('value')).toEqual('small')
+  })
+
+  it('Should provide "large" value when screen is above `responsiveBreakpoints.small`', () => {
```
pith: Don't you want to test that the child component gets the value?

wakooka: Yes I updated the test just now

src/_utils/mediaSizeProvider/MediaSizeProvider.unit.tsx
wakooka: Ah that was nice I tried it, but spyOn only accept a method as the second parameter, here `innerWidth` is only a readonly property of `window`

PR #595, opened 2020-03-11
Spacing fixes on Paragraph and ColumnedContentSection

grosbouddha:


src/layout/section/columnedContentSection/index.tsx
```diff
@@ -9,7 +9,7 @@ const StyledColumnedContentSection = styled(ColumnedContentSection)`
   }
 
   & .kirk-columned-content-section-title {
-    margin: ${space.l} 0;
+    margin: ${space.xl} 0;
```
wakooka: Don't we want to use padding instead? Since margin are merged if the elements after or before also have margins, it's probably working in this case though so not blocker

grosbouddha: yep, swapped to padz.

PR #596, opened 2020-03-11
remove 'higlighted' prop from TripCard

Tuxosaurus:


src/tripCard/index.tsx
```diff
@@ -39,7 +39,7 @@ const StyledTripCard = styled(TripCard)`
     padding-top: 0;
   }
 
-  &.kirk-item--highlighted .kirk-text-body {
+  & .kirk-item--highlighted .kirk-text-body {
```
wakooka: is it intended?

Tuxosaurus: Yes, I messed up before and mistook this as the main "highlighted" class while it's a modifier for an item inside the card

PR #606, opened 2020-03-13
BBC 7140 handle autocomplete loading

Sevrahk:
Display `Loader` component when `AutoComplete` loads its autocomplete list
Backward compatibility is preserved and the loader is only visible within search form

a

### DIFF
--- a/src/autoComplete/AutoComplete.tsx
+++ b/src/autoComplete/AutoComplete.tsx
@@ -10,6 +10,7 @@ import ItemInfo from 'itemInfo'
 import TextField, { inputTypes } from 'textField'
 import AutoCompleteList from './AutoCompleteListStyle'
 import Divider from 'divider'
+import Loader from 'loader'
 
 type query = string | number | boolean
 export interface AutoCompleteProps {
@@ -51,7 +52,7 @@ export interface AutoCompleteProps {
   readonly required?: boolean
   readonly error?: string | JSX.Element
   readonly selectedItemStatus?: ItemStatus
-  readonly shouldDisplayDivider?: boolean
+  readonly embeddedInSearchForm?: boolean
 }
 
 interface AutoCompleteState {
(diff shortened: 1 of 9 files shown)


src/autoComplete/AutoComplete.tsx
```diff
@@ -51,7 +51,8 @@ export interface AutoCompleteProps {
   readonly required?: boolean
   readonly error?: string | JSX.Element
   readonly selectedItemStatus?: ItemStatus
-  readonly shouldDisplayDivider?: boolean
+  readonly embeddedInSearchForm?: boolean
+  readonly loader?: JSX.Element
```
mcellier: Not sure we should expose the `loader` prop. We might end with random elements. Maybe just put a boolean like `shouldDisplayLoader`

mcellier: Well, I just saw the `embeddedInSearchForm` prop. You can just remove the loader prop as we'll always use the `Loader` component. Especially with your css targeting

src/searchForm/autoComplete/overlay/index.tsx
```diff
@@ -39,6 +39,13 @@ const StyledAutoComplete = styled(autoComplete)`
     border-radius: 0;
   }
 
+  & .kirk-textField-wrapper .kirk-loader {
+    position: absolute;
+    top: 12px;
+    bottom: 0;
```
mcellier: Why the bottom?

Sevrahk: Used it for a test, it is not needed anymore
Thanks :)

src/autoComplete/AutoComplete.tsx
Sevrahk: I agree with you.
Actually, I hesitated between keeping the loader outside of AutoComplete to have the lower impact or creating it in the AutoComplete and having only `embeddedInSearchForm`

PR #610, opened 2020-03-18
[BBC-7161] Create small version of Timepicker for filters

Sevrahk:
Capture d’écran 2020-03-18 à 11 32 40

### DIFF
--- a/src/timePicker/TimePicker.tsx
+++ b/src/timePicker/TimePicker.tsx
@@ -36,6 +36,7 @@ export interface TimePickerProps {
   readonly onChange?: (obj: OnChangeParameters) => void
   readonly timeStart?: string
   readonly focus?: boolean
+  readonly small?: boolean
 }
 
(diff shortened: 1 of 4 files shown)


src/timePicker/TimePicker.tsx
```diff
@@ -36,6 +36,7 @@ export interface TimePickerProps {
   readonly onChange?: (obj: OnChangeParameters) => void
   readonly timeStart?: string
   readonly focus?: boolean
+  readonly filterEmbedded?: boolean
```
mcellier: Rename to `small: boolean` to make it more generic

Sevrahk: I thought `small` was too generic so I renamed it just before my commit :)
If it's ok for you I'll change it, thanks

Sevrahk:
Oops forgot this one, I'll update it right away

PR #611, opened 2020-03-19
Fix focus visual indicator for keyboard navigation on Item type Button

nickoola:


nickoola:
I apply this change to keep the focus visual indicator on Buttons for keyboard navigation without displaying it on mouse/pointer click.
`:focus-visible` will be the perfect solution but actually not well supported by browsers. 
In a second time we could use a polyfill to apply `:focus-visible` to differentiate focus on click and focus with keyboard navigation.

PR #615, opened 2020-03-20
Fix MediaSection display when inside a modal

mcellier:
When inside a `Modal`, `MediaSection` didn't take the whole width.

Capture d’écran 2020-03-20 à 14 39 35


gabrielagabriel:
**IMHO We shouldn't mix `Modal` and `Media-Section` components**
Those are distinguished components and should be kept a part. If you wanna join both of them it becomes a "Template" if we follow Atomic Design Presets and then we have "another" component (technical speaking) that emerges.

src/modal/index.tsx
```diff
@@ -81,6 +81,14 @@ const StyledModal = styled(Modal)`
     left: 8px;
     background-color: transparent;
   }
+
+  /* MediaSection is fullwith on small media. Counter the default padding with this */
```
grosbouddha: Can you add a line:
"Ideally, the Modal should not define any interal paddings and let the contained sections take care of the horizontal layout."

grosbouddha:
> **IMHO We shouldn't mix `Modal` and `Media-Section` components**
> Those are distinguished components and should be kept a part. If you wanna join both of them it becomes a "Template" if we follow Atomic Design Presets and then we have "another" component (technical speaking) that emerges.
> 
> In addition **width** is a matter of **"grid" and context**. It should be tread by a _layout grid_ instead of modal component. Or we create a **Modal Grid**.

I see your point and I think it's debatable:
In particular, we have fullscreen modals that need the Section layouts and you might need to compose multiple sections inside the modal.
I would be careful with creating a new grrid system (I believe a good design system should not need a grid but it's subjective and debatable)

Could you add an item to next chapter to discuss this ? @gabrielagabriel

PR #616, opened 2020-03-20
Fix MediaContent height when used inside a modal

mcellier:
Before:
Capture d’écran 2020-03-20 à 15 36 43

After:
Capture d’écran 2020-03-20 à 15 36 05


src/_utils/closest.tsx
```diff
@@ -0,0 +1,16 @@
...
+  let el = this
+
+  do {
+    if (el.matches(s)) return el
+    el = el.parentElement || el.parentNode
+  } while (el !== null && el.nodeType === 1)
+  return null
+}
+
+if (!Element.prototype.closest) {
```
mcellier: Element is not defined server-side. Adding something to handle this

src/layout/content/content.tsx
```diff
@@ -17,11 +18,17 @@ export class MainContent extends Component<MainContentProps> {
...
     }
+
     const topBar: HTMLElement = document.querySelector(this.props.topBarSelector)
-    const height = topBar
-      ? Math.floor(window.innerHeight - topBar.offsetHeight)
-      : window.innerHeight
+    let minus = 0
+
+    if (this.mainContentRef.current.closest('.kirk-modal--hasCloseButton')) {
+      minus = 70 // padding-top of modals with close button
```
gabrielagabriel: **No magic numbers**: is that possible to create a variable for that?

src/layout/content/content.tsx
```diff
@@ -17,11 +18,17 @@ export class MainContent extends Component<MainContentProps> {
...
+    } else if (topBar) {
+      minus = topBar.offsetHeight
+    }
```
gabrielagabriel: else... it will receive `0`?

mcellier: Yes. I've let the previous behavior in place (no topbar).

gabrielagabriel:
IMHO these kind of height issue are fixed with **Flex-Box** does the _bottom height_ is a dynamic variable?

mcellier:
> IMHO these kind of height issue are fixed with **Flex-Box** does the _bottom height_ is a dynamic variable?

I'll let you talk with @PerrineBoissieres who implemented this at first. She'll be better than me at explaining the context

src/layout/content/content.tsx
mcellier: Added the size into `_utils/banding`'s `componentSizes`

PerrineBoissieres:
@gabrielagabriel `MainContent`, `Content` and `BottomContent` use flexbox positionning to have bottom-aligned elements in a page, and it handles mobile browser weird behaviors as well as window resizing :)
`MainContent` has a prop to take in account a specific component (like a topbar) for height calculations, but it doesn't work for modals as they use a top padding. This PR implements a workaround so the `MainContent` can ignore the padding for the close button.

PR #617, opened 2020-03-20
Improve & rename TripSection as CardSection to be able to handle Trip Card and QrCard elements

florianeyzat:
Contains a breaking change (TripSection -> CardSection)

Capture d’écran 2020-03-20 à 16 14 28


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **[UPDATE]** Improve & rename `TripSection` as `CardSection` to be able to handle `TripCard` and `QrCard` elements
```
wakooka: `**[BREAKING CHANGE]**`

src/layout/section/cardsSection/CardsSection.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import cc from 'classcat'
+
+import { TripCardProps } from 'tripCard'
+import { QrCardProps } from 'qrCard'
+
+type CardsProps = TripCardProps | QrCardProps
```
pith: Why using the props and not the component?

```
type CardType = TripCard | QrCard
 ```

Then 

```
readonly children: CardType[] | CardType
```

File: src/layout/section/cardsSection/CardsSection.tsx [hunk 3]
```diff
@@ -0,0 +1,21 @@
...
+
+export interface CardsSectionProps {
+  readonly children: React.ReactElement<CardsProps>[] | React.ReactElement<CardsProps>
+  readonly className?: Classcat.Class
+  readonly width?: string
+}
+
+const CardsSection = ({ children, className = '' }: CardsSectionProps) => (
+  <div className={cc(['kirk-cardsSection-wrapper', className])}>
```
wakooka: Do we need a className prop here?

CHANGELOG.md
pith: This is not an update, this is a breaking change 😉. 

If you want to avoid the breaking change, just add the new CardsSection component without removing the legacy `TripSection`. Then remove it in a future version when `CardsSection` will be used everywhere.

NB: you missed the **s** for CardsSection

src/layout/section/cardsSection/CardsSection.tsx
florianeyzat: It seems not permitted

`'TripCard' refers to a value, but is being used as a type here.ts(2749)
Property 'children' of exported interface has or is using private name 'TripCard'.ts(4033)`

src/layout/section/cardsSection/index.tsx
```diff
@@ -1,52 +1,56 @@
...
-    overflow: auto; /* Make TripCards scrollable horizontally */
+    overflow: auto; /* Make Cards scrollable horizontally */
     -ms-overflow-style: none; /* Remove scrollbar visually IE 10+ */
     scrollbar-width: none; /* Remove scrollbar visually Firefox */
     scroll-snap-type: x mandatory;
     scroll-padding: ${space.xl};
   }
 
-  & .kirk-tripsSection .kirk-tripCard {
+  & .kirk-cardsSection .kirk-tripCard,
```
grosbouddha: Can we have a reused util class between trip and qr cards to avoid all these duplicated rules ?

florianeyzat: Welle technically they both use a `<Card  />` as root so I could use the `.kirk-card` class, I'm  gonna take a look  you're right

PR #618, opened 2020-03-20
New Hint component (and a11y changes)

PerrineBoissieres:
### New Hint component
Supports above and below displays, and basic hiding. Cannot be shown again after hiding.

Capture d’écran 2020-03-20 à 14 40 31
Capture d’écran 2020-03-20 à 14 41 48

### New paradigm for a11y attributes
Since I needed to be able to add `aria-describedby` to other components, I've added a more robust way to add support for aria attributes and others to most our components, and without having to update everything to support one more attribute.

The new interface `A11yProps` whitelists generic DOM attributes that can be used when combining components for accessibility. If you have more to support, you can add them in this interface and all components using it will automatically support them (you'll also need to update the list of keys in the `pickA11yProps<T>()` util).


pith:
Do you think we can challenge the desktop design? It feels really empty with short content. What do you think @wakooka?

src/_utils/interfaces.ts
```diff
@@ -7,3 +7,30 @@ export interface CommonFieldsProps {
...
+}
+
+type A11yKeys = keyof A11yProps
+
+export function pickA11yProps<T extends A11yProps>(source: T): Pick<T, A11yKeys> {
+  const returnValue = {} as Pick<T, A11yKeys>
+  const keys: A11yKeys[] = [
+    'id',
+    'role',
+    'aria-label',
```
mcellier: React uses camelCase for these. Don't you have warnings?

src/_utils/item/Item.tsx
```diff
@@ -47,40 +48,42 @@ export interface ItemProps {
...
+    leftAddon,
+    rightTitle,
+    rightTitleDisplay = TextDisplayType.TITLE,
+    rightTitleColor,
+    rightBody,
+    rightBodyDisplay = TextDisplayType.BODY,
+    rightBodyColor,
+    rightAddon,
+    tag = <div />,
+    ariaLabel,
```
mcellier: This prop duplicates one in the A11yProps

src/hint/Hint.tsx
```diff
@@ -0,0 +1,78 @@
+import React, { Component } from 'react'
+import cc from 'classcat'
+import uniqueid from 'lodash.uniqueid'
+
+import { A11yProps } from '_utils/interfaces'
+import HintBubble, { HintBubblePosition } from './HintBubble'
+
+export interface HintProps extends A11yProps {
+  children: (a11yAttrs: A11yProps) => React.ReactNode
```
mcellier: Why is children a function? It'll be more consistant to just have a react node

src/hint/HintBubble.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const a11yProps = pickA11yProps<HintBubbleProps>(props)
+  const { title, description, closeButtonTitle, position, className, onClose } = props
+  return (
+    <aside className={cc([className, `bubble-arrow--${position}`])} role="tooltip" {...a11yProps}>
+      <p>
+        <strong>{title}</strong>
+        {description}
+      </p>
+      <Button onClick={onClose} title={closeButtonTitle} status={ButtonStatus.UNSTYLED}>
+        <CrossIcon size="18" iconColor={color.textWithBackground} />
```
mcellier: Size should be a number (pros linting doesn't work it seems)

src/itemChoice/index.tsx
```diff
@@ -32,6 +33,7 @@ export interface ItemChoiceProps {
   readonly onMouseDown?: (event: React.MouseEvent<HTMLElement>) => void
   readonly onDoneAnimationEnd?: () => void
   readonly ariaLabel?: string
+  readonly ariaDescribedby?: string
```
mcellier: You already added the prop in A11yProps

src/_utils/item/Item.tsx
PerrineBoissieres: I'll do another ticket for this, I don't want to create breaking changes with this new component.

src/hint/Hint.tsx
PerrineBoissieres: I need to pass some new props to the child, so it's cleaner this way to keep the typing and better for performances than React.clone(). Also it's more flexible so you can attach the a11y attribute to the right element.

src/itemChoice/index.tsx
PerrineBoissieres: Oh yeah, forgot to clean it. Thanks for spotting it.

src/hint/Hint.tsx
```diff
@@ -0,0 +1,78 @@
...
+      position,
+      children,
+      closeButtonTitle,
+      onClose,
+      hidden,
+      ...props
+    } = this.props
+    return (
+      <div className={cc([className, { 'hidden-bubble': this.state.hiddenBubble }])} {...props}>
+        {children(this.id)}
```
grosbouddha: Visually, the bubble appears first (above the hiint target). We should have it first in source order too.

PerrineBoissieres: The aria attribute will make the link between the two components for screen readers. When reading, I think it's better to have first the main content, and then we have the additional informations on it. And I don't want the visual position of the bubble to impact the DOM reading flow.
Maybe @nickoola can confirm or infirm this.

src/_utils/interfaces.ts
PerrineBoissieres: Nope, no warning whatsoever 🤔

wakooka: React doesn't use camelCase for aria-* props

mcellier: My bad, i may have mixed with something else

src/hint/HintBubble.tsx
```diff
@@ -0,0 +1,99 @@
...
+  closeButtonTitle?: string
+  description?: string
+  position?: HintBubblePosition
+}
+
+export const HintBubble = (props: HintBubbleProps) => {
+  const a11yProps = pickA11yProps<HintBubbleProps>(props)
+  const { title, description, closeButtonTitle, position, className, onClose } = props
+  return (
+    <aside className={cc([className, `bubble-arrow--${position}`])} role="tooltip" {...a11yProps}>
```
wakooka: Not sure about the role="tooltip" looks like the specs is not even finished, so I'm wondering how screenreader would read this.

src/hint/Hint.tsx
```diff
@@ -0,0 +1,78 @@
...
+      description,
+      position,
+      children,
+      closeButtonTitle,
+      onClose,
+      hidden,
+      ...props
+    } = this.props
+    return (
+      <div className={cc([className, { 'hidden-bubble': this.state.hiddenBubble }])} {...props}>
```
wakooka: Maybe adding a `aria-live="polite"` would be good so when the component is rendered the screen reader would start reading the content of the Hint. To be tested with a screenreader.

src/hint/Hint.tsx
```diff
@@ -0,0 +1,78 @@
...
+  className?: Classcat.Class
+  onClose?: () => void
+  hidden?: boolean
+}
+
+export interface HintState {
+  hiddenBubble: boolean
+}
+
+class Hint extends Component<HintProps, HintState> {
```
wakooka: Convert to func component using hooks

PerrineBoissieres:
@pith I made a proposal for desktop, but it raises other issues with the bubble needing to be better attached to the main element, ex (text left align, centered bubble, a lot of white background and no clear link between both components):

Capture d’écran 2020-03-23 à 11 28 21

So for now, we keep it as it is, until we work more on global desktop specs.

src/hint/HintBubble.tsx
PerrineBoissieres: Not sure about the tooltip role as well, I'll take other suggestions on how to treat this component :)

nickoola:
After a quick tab navigation test, the focus order is not the good one:
We first **focus on the wrapper**, then focus on the **next focusable elements** in the page and finally go back to **focus the Hint close button**.

src/hint/Hint.tsx
nickoola: In terms on reading order, we should have the Hint content before. Here it's like a `tips label/title`. More over the tabbing order is probably broken because of that (Bubble >> Target >> Close) instead of (Bubble >> Close >> Target).

src/hint/HintBubble.tsx
nickoola: The role="tooltip" doesn't correspond to our component behavior. You can remove it.
It's more coherent to add `aria-live="polite"` to precise there is an important dynamic content (can be removed).

src/hint/HintBubble.tsx
```diff
@@ -0,0 +1,99 @@
...
+      <p>
+        <strong>{title}</strong>
+        {description}
+      </p>
+      <Button onClick={onClose} title={closeButtonTitle} status={ButtonStatus.UNSTYLED}>
```
nickoola: You should add `aria-controls={id-HintBubble}` to link the button to the Hint.

src/hint/HintBubble.tsx
nickoola: Ref for aria-live and aria-controls: https://developer.mozilla.org/fr/docs/Accessibilit%C3%A9/ARIA/Zones_live_ARIA

src/hint/Hint.tsx
nickoola: After screen reader + keyboard navigation tests, I confirm that moving `{children(this.id)}` after the `HintBubble ` will fix the reading order. To resume the appropriate order should be **Hint content >> Hint Close >> The target**

src/hint/HintBubble.tsx
nickoola: And perfect for the `<aside>` tag, the screen reader explicit that it's a complementary content :-)

PR #619, opened 2020-03-23
Boyscout/css nesting

Sevrahk:


src/searchForm/autoComplete/overlay/index.tsx
```diff
@@ -14,47 +14,47 @@ const StyledAutoComplete = styled(autoComplete)`
...
+        position: absolute;
+        top: 12px;
+        right: 12px;
+      }
+    }
+
+    .kirk-items-list {
+      padding: 0;
+
+      &.kirk-autoComplete-list .kirk-items-list-item::before {
```
gabrielagabriel: `.kirk-items-list-item` is repeated on line **54** do we need this selector weight

grosbouddha: let's keep and be safe (since we don't know for sure).

src/searchForm/autoComplete/section/index.tsx
```diff
@@ -6,37 +6,37 @@ import autoComplete from './AutoCompleteSection'
...
+    }
+
+    .kirk-items-list {
+      padding: 0;
+
+      &.kirk-autoComplete-list .kirk-items-list-item::before {
+        content: none;
+      }
+
+      .kirk-items-list-item .kirk-item-choice {
```
gabrielagabriel: `.kirk-items-list-item` has ben already declared.  On line **32**, do you need this strong selector or can you merge both of them?

src/searchForm/autoComplete/section/index.tsx
```diff
@@ -6,37 +6,37 @@ import autoComplete from './AutoCompleteSection'
...
+      input {
+        font-size: ${font.m.size}
+        line-height: ${font.m.lineHeight}
+        background: transparent;
+        border-radius: 0;
+      }
+
+      .kirk-loader {
+        position: absolute;
+        top: 12px;
```
gabrielagabriel: please, use **spaces** defined on utils/branding

src/searchForm/autoComplete/section/index.tsx
```diff
@@ -6,37 +6,37 @@ import autoComplete from './AutoCompleteSection'
...
+        right: 12px;
```
gabrielagabriel: same here

grosbouddha:
Add an entry to the changelog.md

src/searchForm/autoComplete/section/index.tsx
grosbouddha: 12 px does not exist in the branding, you will end up adding constants to create a new const that has no semantic value.
If the PR is about adding nesting, I would keep it focused on just that.

src/searchForm/autoComplete/overlay/index.tsx
```diff
@@ -14,47 +14,47 @@ const StyledAutoComplete = styled(autoComplete)`
...
-    padding: ${space.m} ${space.l};
+    .kirk-textField-wrapper {
+      background: transparent;
+      border-radius: 0;
+      border: 0;
+
+      .kirk-autoComplete-icon {
+        padding: 0;
+        line-height: 0;
+        width: 48px;
```
gabrielagabriel: no magic numbers: could we create variables for icon size?

src/searchForm/autoComplete/section/index.tsx
gabrielagabriel: I mean.. if its a boyscout we should care about the overal style. 
Indeed `12px` doesn't exists, and that is the point: remove **magical numbers**.

Sevrahk: I agree with @grosbouddha, since the PR is about CSS nesting we should only focus on the PR topic and not change anything else but we can plan another PR to fix this issue

wakooka: Even if it's a boyscout PR, it's good to have only one goal per PR, here the goal was to add nesting, not fixing magic numbers.

src/searchForm/autoComplete/overlay/index.tsx
wakooka: I agree. But to fix in another PR

src/searchForm/autoComplete/overlay/index.tsx
Sevrahk: @gabrielagabriel I hesitated but I finally decided to split in 2 classes to improve readability. But I think you are right, regrouping everything inside one `kirk-items-list-item` selector will help understand nesting of HTML elements
I'll make the change thanks :)

src/searchForm/autoComplete/section/index.tsx
Sevrahk: Same comment as above, I'll make the change, thanks ;)

Sevrahk:
> IMHO selectors weight are way too high 3 levels + combination of classes (`class.class`).
> I would recommend reduce levels by not following the DOM

I agree with you.
Actually I don't follow the DOM tag by tag but sometimes increasing CSS weight is required.
For example the selector `.kirk-items-list.kirk-autoComplete-list` seems useless but I needed this in order to increase the CSS weight of nested classes to apply their properties

PerrineBoissieres:
Not a big fan of nesting. I'd rather use the possibilities that styled-component offers, but this would be a bigger refacto, so... okay for now as this is just minor changes, I just wouldn't keep it for everything in the long term.

Sevrahk:
> Not a big fan of nesting. I'd rather use the possibilities that styled-component offers, but this would be a bigger refacto, so... okay for now as this is just minor changes, I just wouldn't keep it for everything in the long term.

For my part I'm a big fan of nesting but based on this PR, for this project I would recommend to use it only for simple nesting like level 1 nesting.
I agree with you, styled-component capabilities are great and I think it's a good idea to use them :)

PR #625, opened 2020-03-25
Search Form Interactions for large devices

wakooka:
This PR takes care of:
- fixing some wrong positions on large devices (not as per specs)
- opening overlays when clicking on each button
- keyboard navigation

To be taken care of in other PRs
- Create Overlay component to be reused in all overlay components (autocomplete, stepper & datepicker)
- Add a `format` method prop to format form values e.g. value for stepper is `2` but it should be rendered as `2 seats`
- Interaction on small devices

Kapture 2020-03-25 at 16 48 06

### DIFF
--- a/src/datePicker/index.tsx
+++ b/src/datePicker/index.tsx
@@ -90,7 +90,8 @@ const StyledDatePicker = styled(DatePicker)`
     color: ${color.fadedText};
     cursor: default;
   }
-  & .DayPicker-Day--selected span {
+  & .DayPicker-Day--selected span,
+  & .DayPicker-Day:focus span {
     background-color: ${color.primary};
     color: ${color.white};
   }
--- a/src/searchForm/SearchForm.tsx
+++ b/src/searchForm/SearchForm.tsx
@@ -1,4 +1,4 @@
-import React, { useContext } from 'react'
+import React, { useContext, useState, useEffect, useRef } from 'react'
 import cc from 'classcat'
 
 import { color } from '_utils/branding'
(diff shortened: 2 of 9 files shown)


src/datePicker/index.tsx
```diff
@@ -90,7 +90,8 @@ const StyledDatePicker = styled(DatePicker)`
     color: ${color.fadedText};
     cursor: default;
   }
-  & .DayPicker-Day--selected span {
+  & .DayPicker-Day--selected span,
+  & .DayPicker-Day:focus span {
```
wakooka: Adding focus indicator on datepicker (see gif)

src/searchForm/autoComplete/overlay/index.tsx
```diff
@@ -12,27 +12,30 @@ const StyledAutoComplete = styled(autoComplete)`
     border: 1px solid ${color.border};
     border-radius: ${radius.l};
     padding: ${space.xl};
+    padding: ${space.xl};
```
Sevrahk: Duplicate with previous line ;)

src/searchForm/index.tsx
```diff
@@ -71,7 +75,6 @@ const StyledSearchForm = styled(SearchForm)`
     font-family: inherit;
     color: ${color.secondaryText};
     background: none;
-    display: flex;
     height: ${buttonHeight};
```
Tuxosaurus: This is messing with the vertical alignment in Firefox, having set to 100% fixes it (in desktop).
Screen Shot 2020-03-26 at 13 31 03

src/searchForm/SearchForm.tsx
```diff
@@ -10,19 +10,79 @@ import { StandardSeat } from 'icon/standardSeat'
...
+}: SearchFormProps) => {
   const mediaSize = useContext(MediaSizeContext)
+
+  const [elementOpened, setElementOpened] = useState('')
+  const [formValues, setFormValues] = useState<FormValues>({})
+
+  const container = useRef(null)
+
+  useEffect(() => {
+    function hideAllOverlay(e: Event) {
```
grosbouddha: hideAllOverlays ? :)

src/searchForm/autoComplete/overlay/index.tsx
```diff
@@ -12,27 +12,30 @@ const StyledAutoComplete = styled(autoComplete)`
...
+    padding-top: ${space.s};
     /* Same as TripCard, we might need to have box-shadow specs so that we can move it to branding file */
     box-shadow: 0 1pt 4pt rgba(0, 0, 0, 0.16), 0 2pt 8pt rgba(0, 0, 0, 0.08);
+    background: ${color.defaultBackground};
 
     .kirk-textField-wrapper {
```
grosbouddha: Loud thinking: I found this much harder to review since we used the CSS nesting :/ (I need to expand all the above code and can't scroll quickly)

src/searchForm/index.tsx
wakooka: Ouch!

src/searchForm/index.tsx
```diff
@@ -17,6 +17,9 @@ const primaryFieldsWidth = '240px'
 const secondaryFieldsWidth = '160px'
 const buttonHeight = '48px'
 const submitButtonWidth = '88px'
+const positionAutocompleteFrom = '-8px'
+const positionAutocompleteTo = '268px'
```
grosbouddha: What are these magic values ?

grosbouddha: We should either:
 - compute them from some other CSS values (to understand tier value)
 - or compute them using some JS
 - or heavily document what's going on

src/searchForm/autoComplete/overlay/index.tsx
Sevrahk: Good point! I didn't thought about this kind of issues.

wakooka: I agree :/

src/searchForm/index.tsx
wakooka: Documented in sketch, according to the grid they use for the form (which I think it's the same as @gabrielagabriel showed us. So maybe there's a need to have a grid, where we could also retrieve the left position of each column.

src/searchForm/index.tsx
wakooka: Fixed by adding `align-items: center` so that the height is still correct for small devices

Tuxosaurus:
There's a "minor issue" when the component appears but that can be investigated later:
kap_search_form_reposition

PR #632, opened 2020-03-27
[BOYSCOUT] Factorize common styles

Sevrahk:


src/searchForm/baseStyles.tsx
```diff
@@ -0,0 +1,50 @@
+import { space, font, color, radius } from '_utils/branding'
+
+const overlayWidth = '375px'
```
Tuxosaurus: nit: not sure if this should be placed in `componentSizes` in `branding`

src/searchForm/baseStyles.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+export const overlayBaseStyle = `
+  box-sizing: border-box;
+  width: ${overlayWidth};
+  border: 1px solid ${color.border};
+  border-radius: ${radius.l};
+  padding: ${space.xl};
+  padding-top: ${space.s};
+  /* Same as TripCard, we might need to have box-shadow specs so that we can move it to branding file */
+  box-shadow: 0 1pt 4pt rgba(0, 0, 0, 0.16), 0 2pt 8pt rgba(0, 0, 0, 0.08);
```
wakooka: Maybe just move this to branding file.

src/searchForm/stepper/overlay/index.tsx
```diff
@@ -1,21 +1,11 @@
...
-    box-sizing: border-box;
-    width: ${overlayWidth};
-    border: 1px solid ${color.border};
-    border-radius: ${radius.l};
-    padding: ${space.xl};
-    padding-top: ${space.s};
-    /* Same as TripCard, we might need to have box-shadow specs so that we can move it to branding file */
-    box-shadow: 0 1pt 4pt rgba(0, 0, 0, 0.16), 0 2pt 8pt rgba(0, 0, 0, 0.08);
-    background: ${color.defaultBackground};
+    ${overlayBaseStyle}
```
gabrielagabriel: question: if you are already overriding the component SepperOverlay why do you need to use overlayBaseStyleHere?

Sevrahk: I don't override components I just factorized common elements and I call them in components ;)

gabrielagabriel: I meant ```const StyledComponent = styled(OtherComponent)``` as an override.
The question is that you could already apply this style to that component directly.. I guess not.

src/searchForm/baseStyles.tsx
Sevrahk: I directly used `shadow.searchForm` since we already defined it :)

PR #634, opened 2020-03-30
[BBC-6997] Responsive`ModalSuccess`

gabrielagabriel:
## Description
**Same same but different**

https://www.loom.com/share/4dd212b8d87b45eeae1fe77f19da08d1

Follow new specs to layout `ModalSuccess`. 

## What has been done

- Override `Modal`
- Use `styled-components` to override/style components
- Since using `styled-components`, no more classes, so converting `classes` to `data-test`
- **chore**: Install a plugin babel to help on debugging

## Things to consider
- Converting classes to data-tests

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -12,6 +12,12 @@
     "@babel/preset-react"
   ],
   "plugins": [
+    [
+      "babel-plugin-styled-components",
+      {
+        "pure": true
+      }
+    ],
     "@babel/plugin-transform-react-jsx",
     "@babel/plugin-proposal-class-properties",
     [
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **[BREAKING CHANGE]** Resposive `SuccessModal` (no more size)
 - **[UPDATE]** Update `jsx-a11y` linter autofocus rule
 - [...]
 
@@ -21,14 +22,25 @@
 
 # v26.0.0 (06/04/2020)
 
+<<<<<<< HEAD
+
 - **[FIX]** Remove DOM Warning on `MainContent` component
 - **[FIX]** Fix markup `Item` tag label by replacing `div` children with `span`
+- # **[BREAKING CHANGE]** Import `icons` as default instead of named export for a better code splitting
 - **[BREAKING CHANGE]** Import `icons` as default instead of named export for a better code splitting
+- **[FIX]** Remove DOM Warning on `MainContent` component
+- **[FIX]** Fix markup `Item` tag label by replacing `div` children with `span`
+  > > > > > > > [BBC-6997] Responsive ModalSuccess
 - **[UPDATE]** Make `Review` more semantic with proper HTML and Review/Rating microdata
 - **[FIX]** Fix clicking oustide of `SearchForm` closes overlays
 - **[UPDATE]** Use `Divider` in `ItemList` instead of custom CSS
 - **[UPDATE]** Remove legacy style from Autocomplete
 
+<<<<<<< HEAD
+
+=======
+
+> > > > > > > [BBC-6997] Responsive ModalSuccess
 
 # v25.0.0 (02/04/2020)
 
--- a/src/successModal/SuccessModal.tsx
+++ b/src/successModal/SuccessModal.tsx
@@ -1,11 +1,12 @@
-import React, { Component } from 'react'
-
-import Modal, { ModalSize } from 'modal'
-import { ModalProps } from 'modal/Modal'
-import Button, { ButtonStatus } from 'button'
-import TheVoice from 'theVoice'
-import { assertModalSizes } from '_utils/assert'
+import React from 'react'
 import uuidv4 from 'uuid/v4'
+
+import { ModalProps } from 'modal'
+import { ButtonStatus } from 'button'
+
+import TextDisplay1 from 'typography/display1'
+
+import { SuccessModalStyle } from './index'
 
 export interface SuccessModalProps extends ModalProps {
   readonly confirmLabel?: string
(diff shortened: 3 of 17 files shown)


src/successModal/SuccessModal.unit.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import { shallow, mount } from 'enzyme'
 
-import SuccessModal from 'successModal'
+import SuccessModal from './successModal'
```
pith: No idea what it wrong with that, but that the only thing I see that could output this error:

```
FAIL src/successModal/SuccessModal.unit.tsx
  ● Test suite failed to run
    Cannot find module './successModal' from 'SuccessModal.unit.tsx'
      1 | import React from 'react'
      2 | import { shallow, mount } from 'enzyme'
    > 3 | 
        | ^
      4 | import SuccessModal from './successModal'
      5 | 
      6 | const defaultProps = {
      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:259:17)
      at Object.<anonymous> (src/successModal/SuccessModal.unit.tsx:3:1)
```

src/successModal/SuccessModal.unit.tsx
```diff
@@ -27,35 +27,23 @@ describe('<SuccessModal>', () => {
   })
 
   it('Should be not visible if isOpen is set to false', () => {
-    expect(wrapper.find('.kirk-modal-dialog').exists()).toBe(false)
+    expect(wrapper.find('[data-test="success-modal"]').exists()).toBe(false)
```
pith: I find this pattern interesting, however it would be better to discuss this with the team before introducing it into the code.

src/successModal/index.tsx
```diff
@@ -1,100 +1,84 @@
...
-  }
+const SuccessAction = styled.div`
+  padding: ${space.xl};
+`
 
-  & .kirk-successModal-confirmButton {
-    vertical-align: bottom; /* vertical alignment to be fixed in Button */
-    border: 0; /* content alignment to be fixed in Button */
-  }
+// @todo: This should be applyed on Button component since it is part of its specification
```
pith: This should be part of our normalisation task (see with @grosbouddha).

src/successModal/story.tsx
```diff
@@ -48,14 +49,16 @@ stories.add(
     <SuccessModalOpener
       isOpen={false}
       onClose={action('onClose')}
-      imageSrc={text('imageSrc', 'https://svgshare.com/i/AGz.svg')}
-      imageText={text('imageText', 'Illustation description')}
-      confirmLabel={text('confirmLabel', 'Got it!')}
-      size={select('size', SuccessModalSize, SuccessModalSize.SMALL)}
```
pith: That's really cool to be able to remove that field :)

src/successModal/index.tsx
```diff
@@ -1,100 +1,84 @@
...
-  & .kirk-modal--large.kirk-successModal .kirk-modal-body {
-    position: relative;
+  .kirk-modal-body {
     display: flex;
-  }
-
-  & .kirk-modal--large .kirk-successModal-bodyItem {
-    width: 50%;
-  }
+    min-height: 100vh;
```
pith: Where you able to test it on different mobile devices? We had a lot of issue in the past with `100vh` as it's not always computed the same way on Android of iOS (for instance when the browser address bar disappear).

src/successModal/SuccessModal.tsx
```diff
@@ -1,86 +1,68 @@
...
+      displayCloseButton={false}
+      displayDimmer={false}
+      forwardedRef={forwardedRef}
+      className={className}
+      ariaLabelledBy={successContentId}
+      data-test="success-modal"
+    >
+      <StyledIllustratedSection illustrationUrl={imageSrc} illustrationAlt={imageText}>
+        <SuccessTitle isInverted data-test="success-title">
+          {' '}
```
pith: Why this space here?

src/successModal/index.tsx
gabrielagabriel: Thanks for the feedback, I'm going test it

.babelrc
```diff
@@ -12,6 +12,12 @@
     "@babel/preset-react"
   ],
   "plugins": [
+    [
+      "babel-plugin-styled-components",
+      {
+        "pure": true
```
grosbouddha, line 18: What is the effect of this ?

src/successModal/index.tsx
```diff
@@ -1,100 +1,84 @@
...
+      flex-wrap: wrap;
+    }
   }
+`
 
-  & .kirk-modal--small .kirk-successModal-image {
-    max-height: 33vh;
-    object-fit: contain;
-  }
+const StyledIllustratedSection = styled(IllustratedSection)`
```
grosbouddha, line 40: I don't feel confortable overriding and coupling illustrated section inside the SuccessModal styles.
Components should be as orthogonal as possible and here we are introducing a very deep coupling between their CSS (which is the worst coupling to maintain and test)

src/successModal/index.tsx
grosbouddha: Button and ButtonLayout is a big piece of incoming work for normalization. I only work on high level layout normalization so far.
About this, If it's really a generic layout property of the button, this margin should be in the Button CSS not here in the success modal.

src/successModal/index.tsx
```diff
@@ -1,100 +1,84 @@
...
-    margin: 0 auto;
+  .kirk-illustratedSection-content {
+    display: flex;
+    flex-direction: column;
+    flex: 1;
   }
+`
+const SuccessTitle = styled(TextDisplay1)`
+  padding: ${space.xl};
+  flex: 1;
```
grosbouddha, line 62: Why the flex here ?
Why the specific title here ? could we reuse either a TheVoice or SubHeader ?

The global question is: why the content of the success modal is not a list of existing Kirk widgets and why do we need all these CSS overrides ?

src/successModal/index.tsx
grosbouddha: Ideally, the modal should just be a particular case of a kairos page e.g.
<MyModal>
  <TheVoice>
  <IllustratedSection />
  <ButtonSection />
</MyModal>

We could have some generic CSS for all modal to manage the layout but not a bunch of CSS that couple a modal type to components inside (like you have between the succcessmodal and illustrated section).

Once more, I did not touch the layout part for modals and modal components. But the fact that we write a bunch more of styles (either it's CSS or Styled component) for only one case is a sign that's it's not generic enough and maybe not a good direction.

src/successModal/index.tsx
gabrielagabriel: This brings a lot of impact and rather create another ticket to solve the issue.. and keep it here.

.babelrc
gabrielagabriel: Help to debug. Styled Component create classes with component names instead of only encryption.

`typography__StyledText-sc-19f9nb1-0 display1__TextDisplay1-sc-1dbfelc-0 successModal__SuccessTitle-sc-17o2h4e-2 kjwuIY`

src/successModal/index.tsx
gabrielagabriel: 🤷‍♀️ Well `IllustratedSection` does not support centred layouts (And that is why I've been proposing grid components pour manage layout instead of overriding components layout style with other styles).

You can check my other draft PR to see the solution without `IllustratedSection`. I've only applied what we have discussed in chapter to avoid creating technical debt by creating the same component twice.

src/successModal/index.tsx
gabrielagabriel: The CSS overrides were already there. I need _flex_ to control, well, _layout_. And sadly, kirk doest have no grid components.. 

`TheVoice` has its behavior on his own and I rather use Typography components to manage that to have a **pure** state. _Otherwise, it will be more and more overrides_

Should I replace to `TheVoice` then?

src/successModal/index.tsx
gabrielagabriel: actually I've found that `min-height: -webkit-fill-available;` would resolve this issue

CHANGELOG.md
```diff
@@ -9,9 +10,9 @@
 
 # v26.0.0 (06/04/2020)
 
+- **[BREAKING CHANGE]** Import `icons` as default instead of named export for a better code splitting
```
grosbouddha, line 13: Remove ?

src/successModal/index.tsx
grosbouddha: I don't understand your last comment; min-height: 100vh is still used in the last snapshot.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **[BREAKING CHANGE]** Remove `size` from `SuccessModal` api
```
grosbouddha, line 3: I would split the breaking change bit in a separate and as-small-as-possible changeset.

PR #636, opened 2020-03-30
[BBC-7350] Update icons import

gabrielagabriel:
## Description

Make sure that no components are using named exports

## What has been done

**From** 
```
import { Icon } from 'icon'
```

**To**
```
import Icon  from 'icon/icon_name'
```


mcellier:
> We might as well remove the main icon file to prevent import errors. Breaking change, but that'll be nicer

As discussed, let's do it in another PR

PR #637, opened 2020-03-31
[BBC-7351] Make sure to import named icons

gabrielagabriel:
## Description
Make sure that we can't name import an icon

## What has been done

- Move index.tsx with icons export to icons story folder
- Separate icons spec from components (create a folder called a story)
- Move design tokens around


src/icon/story/story.tsx
```diff
@@ -3,8 +3,9 @@ import { storiesOf } from '@storybook/react'
 import { withKnobs, text, select, number, boolean } from '@storybook/addon-knobs'
```
mcellier, line 3: Rename this file `index.tsx` to have nicer import

src/index.unit.tsx
```diff
@@ -1,7 +1,6 @@
 import fs from 'fs'
 
 import * as kirk from './index'
-import * as icon from './icon/index'
```
mcellier, line 4: Why are you removing single icon stories?
That's the only way we have to use knobs to test modifing their props.

gabrielagabriel: It's the test file.
Knobs still up and running

PR #640, opened 2020-04-01
Update a11y linter settings to recommended

nickoola:


src/menu/index.tsx
```diff
@@ -10,6 +10,8 @@ export interface MenuProps {
 }
 
 const Menu = ({ className, children, withSeparators = false }: MenuProps) => (
+  /* TODO: BBC-7416 fix a11y issue */
+  // eslint-disable-next-line jsx-a11y/no-noninteractive-element-to-interactive-role
   <nav className={cc(className)} role="menu">
```
PerrineBoissieres, line 15: This one is weird, no?

nickoola: Following this spec:
https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html
the `role="menu"` should be applied on an `ul` element.

grosbouddha: We overuse the <nav> in general (I'm trying to remove them gradually from kairos).
In short, a <nav> is non focusable and only to describe content while the menu role is attached to an element that will receive focus and keys (to open/close/up/down in the menu) hence the valid warning.

src/text/Text.unit.tsx
```diff
@@ -16,7 +16,7 @@ describe('Text', () => {
   it('should have a role and classname', () => {
     const text = renderer
       .create(
-        <Text className="custom" role="role">
+        <Text className="custom" role="complementary">
```
grosbouddha, line 19: I know it's just a test but having this complementary  role on a text element is quite wrong and might only lead to people copy/pasting that bad exemple.
I would remove the role completely (why would you want to do that in the first place ?)

wakooka: I'm guessing it's to check that the prop `role` is passed, but it could be a unit test instead of a snapshot

nickoola: Yep and the snapshot to test custom class is useless too.
I'll check props another way

PR #642, opened 2020-04-01
Web Search Form: small devices interactions

wakooka:
Kapture 2020-04-01 at 18 26 03

### DIFF
--- a/src/searchForm/SearchForm.tsx
+++ b/src/searchForm/SearchForm.tsx
@@ -1,5 +1,7 @@
-import React, { useContext, useState, useEffect, useRef } from 'react'
+import React, { useState, useEffect, useRef } from 'react'
+import { createPortal } from 'react-dom'
 import cc from 'classcat'
+import { canUseDOM } from 'exenv'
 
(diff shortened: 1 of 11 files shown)


src/searchForm/SearchForm.tsx
```diff
@@ -59,7 +65,9 @@ const SearchForm = ({
   datepickerProps,
   stepperProps,
 }: SearchFormProps) => {
-  const mediaSize = useContext(MediaSizeContext)
+  // Use React.useContext syntaxt so we can mock it
+  // https://github.com/enzymejs/enzyme/issues/2176#issuecomment-533582429
+  const mediaSize = React.useContext(MediaSizeContext)
```
Tuxosaurus, line 70: Any reason why you removed `useContext` from the import to use it directly here?

wakooka: It's explained in the comment above. A line break before the comment will help to group them together maybe?

Tuxosaurus: Oh sorry I didn't see it, perfect then!

PR #643, opened 2020-04-02
Fix ItemRadio markup

nickoola:


src/itemRadio/ItemRadio.tsx
```diff
@@ -105,7 +105,10 @@ class ItemRadio extends Component<ItemRadioProps> {
           rightTitle={data}
           rightTitleDisplay={TextDisplayType.SUBHEADERSTRONG}
           rightBody={dataInfo}
-          /* TODO: BBC-7415 fix a11y issue */
+          /* No a11y issue here
+            - The input is well wrapped with the label
+            - The linter can't access the complex components implementation
+          */
```
nickoola, line 111: Tested with keyboard nav + VoiceOver. Working well on my side.

src/_utils/item/Item.tsx
```diff
@@ -116,14 +116,14 @@ const Item = (props: ItemProps) => {
       {...a11yAttrs}
       aria-label={ariaLabel}
     >
```
nickoola, line 118: https://stackoverflow.com/questions/18609554/is-div-inside-label-block-correct

src/_utils/item/index.tsx
```diff
@@ -73,6 +73,10 @@ const StyledItem = styled(Item)`
     position: relative;
   }
 
+  & .kirk-text-title {
```
nickoola, line 76: Exept this element, no visual impact on my side.

src/_utils/item/index.tsx
```diff
@@ -90,6 +94,7 @@ const StyledItem = styled(Item)`
   }
 
   /* Highlighted state */
+
```
wakooka, line 97: Remove?

src/itemRadio/ItemRadio.tsx
wakooka: I think you should add a unit test to make sure that itemRadio always ouput a `<label />` with a very clear test description so that we don't break it.

nickoola: Done!

PR #650, opened 2020-04-06
Use Divider component in ItemListDivider

wakooka:
Tested: visually with all the stories

### DIFF
--- a/src/itemsList/ItemsList.tsx
+++ b/src/itemsList/ItemsList.tsx
@@ -4,6 +4,7 @@ import cc from 'classcat'
 import { ItemChoiceProps } from 'itemChoice'
 import { ItemRadioProps } from 'itemRadio/ItemRadio'
 import { ItemCheckboxProps } from 'itemCheckbox/ItemCheckbox'
+import Divider from 'divider'
 
 export const ItemsListDivider: FunctionComponent = () => null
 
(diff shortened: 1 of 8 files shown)


src/itemsList/ItemsList.tsx
```diff
@@ -31,33 +32,23 @@ class ItemsList extends Component<ItemsListProps> {
...
-                ])}
-                key={keyGenerator(index)}
-              >
+              <li className={cc(['kirk-items-list-item'])} key={keyGenerator(index)}>
                 {item}
+                {hasSeparator && <Divider />}
               </li>
             )
-            separator = false
+            hasSeparator = false
```
grosbouddha, line 51: This hasSeparator is never used (it's reset by the next "let hasSeparator =") in the next loop iteration.
Am I missing something or something is missing or can we just remove that line safely ?

src/itemsList/ItemsList.tsx
```diff
@@ -31,33 +32,23 @@ class ItemsList extends Component<ItemsListProps> {
...
-        className={cc([
-          'kirk-items-list',
-          className,
-          { 'kirk-items-list--withSeparators': withSeparators },
-        ])}
-        {...otherProps}
-      >
+      <ul className={cc(['kirk-items-list', className])} {...otherProps}>
         {children.map((item, index) => {
           let child = null
```
grosbouddha, line 39: I would simplify the logic with an early return e.g.
```
if (item.type === ItemsListDivider || item.type === undefined) {
  return null
}

let hasSeparator = ...
return (<li>..</li>)

```

wakooka:
PTAL

PR #651, opened 2020-04-06
SelectField a11y autofocus issue

nickoola:


nickoola:
As the focus can be setted manualy if needed on components (even on page load like do OneAction) and `autoFocus` attribute could be confusing for assistive tools.
I simply suggest to remove it.
>> https://developer.mozilla.org/fr/docs/Web/HTML/Element/Input
>> https://medium.com/@gavyn/til-autofocus-inputs-are-an-accessibility-problem-32ced60c3109
I'm not sure there is an easy way for user to disable `autoFocus` with their browser.
And I agree that forcing the focus on page load (without autoFocus) will occur the same a11y issue.

wakooka:
> I'm not sure there is an easy way for user to disable autoFocus with their browser.
you're right, it doesn't look like there is an easy way to disable this in a browser

PR #654, opened 2020-04-07
[BBC-7444] Calendar is limited to 1 year

Sevrahk:
On large screen we can navigate through calendar with arrow buttons and we can display dates in far future.
The calendar will now display dates until 1 year by default

a

### DIFF
--- a/src/datePicker/DatePicker.tsx
+++ b/src/datePicker/DatePicker.tsx
@@ -36,6 +36,7 @@ export interface DatePickerProps {
   readonly orientation?: DatePickerOrientation
   readonly isOutsideRange?: (day: Date) => boolean
   readonly fromMonth?: Date
+  readonly toMonth?: Date
   readonly firstDayOfWeek?: number
   readonly stickyPositionTop?: number
 }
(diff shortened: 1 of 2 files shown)


mcellier:
Add changelog update

src/datePicker/DatePicker.tsx
```diff
@@ -56,6 +57,7 @@ export class DatePicker extends PureComponent<DatePickerProps, DatePickerState>
     isOutsideRange: day => DayPicker.DateUtils.isDayBefore(day, new Date()),
     orientation: DatePickerOrientation.HORIZONTAL,
     fromMonth: new Date(),
+    toMonth: new Date(new Date().getFullYear() + 1, new Date().getMonth(), new Date().getDate()),
```
wakooka, line 60: is it possible to update the prop `isOutsideRange` instead? Like that we would have one way to handle the bounds.
something like this:
```
const dateInOneYear = new Date(new Date().getFullYear() + 1, new Date().getMonth(), new Date().getDate())
...
isOutsideRange: day => DayPicker.DateUtils.isDayBefore(day, new Date()) || DayPicker.DateUtils.isDayAfter(day, dateInOneYear)
```
I'm just not sure what the results will be in the UI.

Sevrahk: I already tried to use `isOutsideRange ` and the result is not what we expected

`toMonth` is responsible of the min and max month we can navigate through (visibility of navigation arrow buttons)
`isOutsideRange` is responsible of the date we can select but even if we could select only the dates in 1 year range we still would be able to navigate in year far away.

In our case `isOutsideRange` is used to prevent selecting past days in current month

PR #657, opened 2020-04-09
Renamed mediaSizeProvider filename into index.tsx

wakooka:


CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # Unreleased
 
 - **[UPDATE]** Add `toMonth` parameter to `DatePicker` and set it to 1 year by default
-- [...]
+- **[BREAKING CHANGE]** `MediaSizeProvider` renamed
```
mcellier: Was it exported outside kirk? If not, that's not a breaking change

wakooka: Everything is exported actually even though it's in _utils :/

PR #659, opened 2020-04-10
Implement focus visible provider

nickoola:


src/_utils/focusVisible/FocusVisibleProvider.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  /**
+   * When the polfyill first loads, assume the user is in keyboard modality.
+   * If any event is received from a pointing device (e.g. mouse, pointer,
+   * touch), turn off keyboard modality.
+   * This accounts for situations where focus enters the page from the URL bar.
+   * @param {PointerEvent} e
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const onInitialPointerMove = (e: any) => {
```
nickoola, line 29: Don't find a better Typescript event definition for for now because of `target.nodeName`

src/_utils/focusVisible/FocusVisibleProvider.tsx
```diff
@@ -0,0 +1,131 @@
...
+    return () => {
+      document.removeEventListener('keydown', onKeyDown, true)
+      document.removeEventListener('mousedown', onPointerDown, true)
+      document.removeEventListener('pointerdown', onPointerDown, true)
+      document.removeEventListener('touchstart', onPointerDown, true)
+      document.removeEventListener('visibilitychange', onVisibilityChange, true)
+
+      removeInitialPointerMoveListeners()
+    }
+  }, [setHadKeyboardEvent])
```
nickoola, line 122: Hook effect only used if `setHadKeyboardEvent ` value change.

src/_utils/focusVisible/FocusVisibleProvider.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  return (
+    <FocusVisibleContext.Provider value={{ hadKeyboardEvent, isInitialized: true }}>
```
nickoola, line 125: `hadKeyboardEvent ` updated from this component state so might have no desired re-render effect.
It will change when user switch form keyboard navigation to pointer navigation.

nickoola:
Intersting discussion about context re-render:
https://github.com/facebook/react/issues/15156

src/_utils/focusVisible/FocusVisibleConsumer.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  let focusVisible: boolean
+  if (isInitialized) {
+    focusVisible = hadKeyboardEvent && isFocused
+  } else {
+    // Fallback to focused when the `FocusVisibleProvider` is not used.
+    focusVisible = isFocused
+  }
+
+  return React.useMemo(
```
nickoola, line 24: Prevent re-render

src/_utils/focusVisible/FocusVisibleConsumer.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+import { FocusVisibleContext } from '_utils/focusVisible/FocusVisibleProvider'
+
+export const FocusVisibleConsumer = () => {
```
wakooka, line 4: This is a custom hook, it should be named  `useFocusVisible`.

src/_utils/focusVisible/FocusVisibleConsumer.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const [isFocused, setIsFocused] = React.useState(false)
+  const { hadKeyboardEvent, isInitialized } = React.useContext(FocusVisibleContext)
+
+  const onFocus = React.useCallback(() => {
+    if (!isFocused) setIsFocused(true)
+  }, [isFocused])
+
+  const onBlur = React.useCallback(() => {
+    if (isFocused) setIsFocused(false)
+  }, [isFocused])
```
wakooka, line 14: I don't think you need a `if` here:
```
React.useCallback(() => {
  setIsFocused(isFocused)
}, [isFocused])
```

src/_utils/focusVisible/FocusVisibleProvider.tsx
wakooka: Try `MouseEvent`

src/_utils/focusVisible/FocusVisibleConsumer.tsx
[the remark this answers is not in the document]
nickoola: Yep go for `useFocusVisible ` :-)

src/_utils/focusVisible/FocusVisibleConsumer.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+import { FocusVisibleContext } from '_utils/focusVisible/FocusVisibleProvider'
+
+export const FocusVisibleConsumer = () => {
+  const [isFocused, setIsFocused] = React.useState(false)
```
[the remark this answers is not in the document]
nickoola: I think the provider should only manage if the user is using keyboard or pointer.
For me the focus state is related to the consumer, not the context.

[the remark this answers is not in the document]
simonrelet: Oh my bad, I miss read the code 😅

simonrelet:
I'm having difficulties seeing how this will be used.

For example the `Button` component is structured as following:

```tsx
// Button.tsx
class Button extends PureComponent<P, S> {
  // ...
}

// index.tsx
const StyledButton = styled(Button)`
  /* [...] */
`
```

So in order to use the hook we need to transform the `Button` to a functional component:

```tsx
// Button.tsx
function Button({ className }) {
  const { focusVisible, onFocus, onBlur } = useFocusVisible()
  // ...

  // Simplified for the example
  return (
    <button
      onFocus={onFocus}
      onBlur={onBlur}
      className={cc([className, prefix({ focusVisible })])}
    />
  )
}

// index.tsx
const StyledButton = styled(Button)`
  /* ... */

  &.kirk-focusVisible {
    /* ... */
  }
`
```

Is that right?

src/_utils/focusVisible/FocusVisibleProvider.tsx
nickoola: Doesn't work either 😕

src/_utils/focusVisible/FocusVisibleProvider.tsx
```diff
@@ -0,0 +1,129 @@
+import React, { createContext, useEffect, useState, useMemo, ReactNode } from 'react'
+
+// A React hook based on: https://github.com/WICG/focus-visible
+
+interface FocusVisibleProviderProps {
+  children: ReactNode
+}
+
+export const FocusVisibleContext = createContext({
+  hadKeyboardEvent: false,
```
nickoola, line 10: Setted to `false` as initial value cause not the common behavior.

src/_utils/focusVisible/FocusVisibleProvider.unit.tsx
```diff
@@ -0,0 +1,35 @@
...
+      <FocusVisibleProvider>
+        <ChildComponent />
+      </FocusVisibleProvider>,
+    )
+
+    const button = wrapper.find('button')
+
+    expect(button.hasClass('initialized')).toBe(true)
+    expect(button.hasClass('keyboardEvent')).toBe(false)
+  })
```
nickoola, line 34: Don't find a way to simulate key events on `document` with `element.simulate('keydown', { key: 'Tab' }`.

src/why/Why.tsx
```diff
@@ -10,11 +11,27 @@ export interface WhyProps {
...
 }
 
-const Why = ({ className, children, title, onClick }: WhyProps) => (
-  <button type="button" className={cc(['kirk-why', className])} title={title} onClick={onClick}>
-    <QuestionIcon />
-    <span>{children}</span>
-  </button>
-)
+const Why = ({ className, children, title, onClick }: WhyProps) => {
+  const { focusVisible, onFocus, onBlur } = useFocusVisible()
```
nickoola, line 15: @simonrelet to answer your question about the `useFocusVisible` implementation, here is this example with the `Why` component.

src/why/index.tsx
```diff
@@ -20,6 +20,14 @@ const StyledWhy = styled(Why)`
     background-color: ${color.lightBackground};
   }
 
+  :focus:not(.focus-visible) {
+    outline: none;
+  }
+
+  .focus-visible {
+    outline: 1;
+  }
```
nickoola, line 29: @simonrelet by using the provider value, if the user is focused on this component by using keyboard he will see this outline style.

nickoola:
> I'm having difficulties seeing how this will be used.
> 
> For example the `Button` component is structured as following:
> 
> ```tsx
> // Button.tsx
> class Button extends PureComponent<P, S> {
>   // ...
> }
> 
> // index.tsx
> const StyledButton = styled(Button)`
>   /* [...] */
> `
> ```
> 
> So in order to use the hook we need to transform the `Button` to a functional component:
> 
> ```tsx
> // Button.tsx
> function Button({ className }) {
>   const { focusVisible, onFocus, onBlur } = useFocusVisible()
>   // ...
> 
>   // Simplified for the example
>   return (
>     <button
>       onFocus={onFocus}
>       onBlur={onBlur}
>       className={cc([className, prefix({ focusVisible })])}
>     />
>   )
> }
> 
> // index.tsx
> const StyledButton = styled(Button)`
>   /* ... */
> 
>   &.kirk-focusVisible {
>     /* ... */
>   }
> `
> ```
> 
> Is that right?

Yes it'll be used this way + having the <FocusVisibleProvider> as wrapper like <MediaSizeProvider>

src/why/index.tsx
simonrelet: Ok I see, thanks.

simonrelet: It looks like the selector `.focus-visible` (line 27) can be removed as everything is handled by the selector `:focus:not(.focus-visible)` (line 23)

nickoola: Ah yes you right, it's not required now.

src/_utils/focusVisible/FocusVisibleProvider.unit.tsx
nickoola: Tried this idea with no success:
https://bytes.babbel.com/en/articles/2019-01-22-testing-global-event-listener-within-react-component.html

src/_utils/focusVisible/FocusVisibleProvider.tsx
```diff
@@ -0,0 +1,130 @@
...
+      document.removeEventListener('mousedown', onPointerDown, true)
+      document.removeEventListener('pointerdown', onPointerDown, true)
+      document.removeEventListener('touchstart', onPointerDown, true)
+      document.removeEventListener('visibilitychange', onVisibilityChange, true)
+
+      removeInitialPointerMoveListeners()
+    }
+  }, [setHadKeyboardEvent])
+
+  const value = useMemo(() => ({ hadKeyboardEvent, isInitialized: true }), [hadKeyboardEvent])
```
simonrelet, line 125: As `isInitialized` is always `true` we can maybe remove it and simplify the `value` prop to `value={hadKeyboardEvent}` (remove the memoization entirely).

src/_utils/focusVisible/FocusVisibleProvider.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const onPointerDown = () => {
+    setHadKeyboardEvent(false)
+  }
+
+  /**
+   * When the polfyill first loads, assume the user is in keyboard modality.
+   * If any event is received from a pointing device (e.g. mouse, pointer,
+   * touch), turn off keyboard modality.
+   * This accounts for situations where focus enters the page from the URL bar.
+   * @param {MouseEvent} e
```
grosbouddha, line 26: You can drop the param doc with TS.

src/_utils/focusVisible/FocusVisibleProvider.tsx
```diff
@@ -0,0 +1,130 @@
...
+  /**
+   * If the most recent user interaction was via the keyboard;
+   * and the key press did not include a meta, alt/option, or control key;
+   * then the modality is keyboard. Otherwise, the modality is not keyboard.
+   * Apply `focus-visible` to any current active element and keep track
+   * of our keyboard modality state with `hadKeyboardEvent`.
+   * @param {KeyboardEvent} e
+   */
+  const onKeyDown = (e: KeyboardEvent) => {
+    if (e.metaKey || e.altKey || e.ctrlKey) {
```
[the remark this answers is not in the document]
grosbouddha: I would say at least Tab, arrow keys, ESC, space, return/enter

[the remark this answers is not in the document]
nickoola: This condition come from the original polyfill here: https://github.com/WICG/focus-visible
But your right it's better to have a keys whitelist 👍

src/_utils/focusVisible/FocusVisibleProvider.tsx
```diff
@@ -1,5 +1,13 @@
 import React, { createContext, useEffect, useState, useMemo, ReactNode } from 'react'
 
+const KEYCODE_TRIGGERING_FOCUS_VISIBLE = [
+  37, // Arrow left
+  38, // Arrow up
+  39, // Arrow right
+  30, // Arrow down
+  9, // Tab
+]
+
```
wakooka, line 10: There's a keycode file here, maybe it's good to add it there: https://github.com/blablacar/ui-library/blob/master/src/_utils/keycodes.ts

src/_utils/focusVisible/FocusVisibleProvider.unit.tsx
wakooka: I did something similar for the search form look at
https://github.com/blablacar/ui-library/blob/master/src/searchForm/SearchForm.unit.tsx#L83

src/_utils/focusVisible/FocusVisibleProvider.tsx
```diff
@@ -0,0 +1,97 @@
+import React, { createContext, useEffect, useState, useMemo, ReactNode } from 'react'
```
wakooka, line 1: Can you rename this file? `src/_utils/focusVisibleProvider/index.tsx`

src/_utils/focusVisible/FocusVisibleProvider.tsx
```diff
@@ -0,0 +1,110 @@
...
+      lastClientX = e.clientX
+      lastClientY = e.clientY
+    }
+
+    const addInitialPointerMoveListeners = () => {
+      document.addEventListener('mousemove', (event: MouseEvent) => {
+        getLastMouseMove(event)
+        onInitialPointerMove
+      })
+      document.addEventListener('mousedown', onInitialPointerMove)
```
gabrielagabriel, line 54: isnt possible to attach events to document dynamically? (on a `for` for instance)

src/_utils/focusVisible/useFocusVisible.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const onFocus = useCallback(() => {
+    setIsFocused(true)
+  }, [isFocused])
+
+  const onBlur = useCallback(() => {
+    setIsFocused(false)
+  }, [isFocused])
+
+  let focusVisible: boolean
+  if (isInitialized) {
```
gabrielagabriel, line 17:
```suggestion
  focusVisible  = isInitialized ? hadKeyboardEvent && isFocused : isFocused
```

src/_utils/focusVisible/FocusVisibleProvider.tsx
```diff
@@ -0,0 +1,130 @@
...
+   * When the polfyill first loads, assume the user is in keyboard modality.
+   * If any event is received from a pointing device (e.g. mouse, pointer,
+   * touch), turn off keyboard modality.
+   * This accounts for situations where focus enters the page from the URL bar.
+   * @param {MouseEvent} e
+   */
+
+  const onInitialPointerMove = (e: MouseEvent) => {
+    // Work around a Safari quirk that fires a mousemove on <html> whenever the
+    // window blurs, even if you're tabbing out of the page. ¯\_(ツ)_/¯
```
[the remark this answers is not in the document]
nickoola: The check for the Safari quirk on the HTML element come from the library which create the polyfill. I do not reproduce either to I removed it.
And thanks to spot the Safari weird mousemove event.
I added the fix to the mousemove event listener only as it seems related only to this one.

src/_utils/focusVisible/FocusVisibleProvider.tsx
nickoola: Done

src/_utils/focusVisible/FocusVisibleProvider.tsx
nickoola: Done

src/_utils/focusVisible/useFocusVisible.tsx
nickoola: `isInitialized ` is removed now

src/_utils/focusVisible/FocusVisibleProvider.tsx
nickoola: Done

src/_utils/focusVisible/FocusVisibleProvider.tsx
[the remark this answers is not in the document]
simonrelet: Yes, it look like this is expected because inputs should always have a visible focus, that's their way of ensuring it.

src/_utils/focusVisible/FocusVisibleProvider.tsx
simonrelet: [`KeyboardEvent .keyCode` is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode), can't we switch to [`KeyboardEvent .key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key)?

It is already [well supported](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key#Browser_compatibility) and React's [SyntheticEvent](https://reactjs.org/docs/events.html) [set it for us](https://github.com/facebook/react/blob/b87aabdfe1b7461e7331abb3601d9e6bb27544bc/packages/react-dom/src/events/getEventKey.js#L79) when the browser doesn't support it.

src/_utils/focusVisible/FocusVisibleProvider.unit.tsx
nickoola: Yes it works in the end.
I just don't find a way to check the provider value on specific tag element.
I mean for example, fire a keyboard event on a Textarea with typing key (Spacebar/Enter)

src/_utils/focusVisible/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+  'touchmove',
+  'touchstart',
+  'touchend',
+]
+
+const pointerDownEventList = ['mousedown', 'pointerdown', 'touchstart']
+
+export const FocusVisibleContext = createContext(false)
+
+export const FocusVisibleProvider = (props: PropsWithChildren<{}>) => {
```
nickoola, line 25: @simonrelet I use `PropsWithChildren ` as type for children instead of interface like:
```
interface FocusVisibleProviderProps {
  children: ReactNode
}
```
Not typescript expert so just want to know if it's the right way.

src/_utils/keycodes.ts
```diff
@@ -4,4 +4,25 @@ export const KEYCODES = {
...
+
+export const KEYS_TRIGGERING_KEYBOARD_NAVIGATION = [
+  KEY.TAB,
+  KEY.ARROWDOWN,
+  KEY.ARROWLEFT,
+  KEY.ARROWRIGHT,
+  KEY.ARROWUP,
+]
+
+export const KEYS_TRIGGERING_KEYBOARD_NAVIGATION_COMPLEMENTARY = [KEY.ENTER, KEY.SPACEBAR]
```
wakooka, line 26: I don't understand why this is a separate variable and not included in `KEYS_TRIGGERING_KEYBOARD_NAVIGATION`

src/_utils/focusVisible/index.tsx
```diff
@@ -0,0 +1,107 @@
+import React, { PropsWithChildren, createContext, useEffect, useState } from 'react'
```
wakooka, line 1: Can you name the folder focusVisibleProvider so we know that it's a provider and to keep it consistent with mediaSizeProvider

src/_utils/keycodes.ts
nickoola: I do it this way because ENTER and SPACEBAR are related to keyboard navigation if not on `textarea` or `input[type=text]`. That's why I'm checking the event.target.tagName in the provider (`onKeyDown` function)

src/_utils/focusVisible/index.tsx
nickoola: Yep done

src/_utils/keycodes.ts
nickoola: And I finally moved this logic into the `onKeyDown` function by filtering the array.

src/_utils/focusVisible/index.tsx
simonrelet: I have rarely seen `PropsWithChildren` before, but I don't feel comfortable having a component with props but no exported prop types.

If we really want to avoid the `children?: React.ReactNode` part maybe we can write:

```ts
export type FocusVisibleProviderProps = PropsWithChildren<{}>
```

But the empty type (`<{}>`) still feels weird.

Having the extended type declaration (`children?: React.ReactNode`) wouldn't bother me much.

Not sure my comment helps here 😅.

src/_utils/focusVisibleProvider/index.unit.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const hadKeyboardEvent = useContext(FocusVisibleContext)
+  return <input type="button" className={hadKeyboardEvent.toString()} value="Test" />
+}
+
+describe('FocusVisibleProvider', () => {
+  let parentNode
+  let wrapper
+  beforeEach(() => {
+    parentNode = document.createElement('div')
+    document.body.appendChild(parentNode)
```
simonrelet, line 18: Isn't this already done by enzyme?

src/_utils/focusVisibleProvider/index.unit.tsx
```diff
@@ -0,0 +1,58 @@
+import React, { useContext } from 'react'
+import { act } from 'react-dom/test-utils'
+import { mount } from 'enzyme'
+
+import FocusVisibleProvider, { FocusVisibleContext } from '.'
+import { KEY } from '_utils/keycodes'
+
+const ChildComponent = () => {
+  const hadKeyboardEvent = useContext(FocusVisibleContext)
+  return <input type="button" className={hadKeyboardEvent.toString()} value="Test" />
```
simonrelet, line 10: It look like the input is only here to have a `className` prop which is then only used to check the value of the context. maybe we can simplify it like this:

```tsx
describe('FocusVisibleProvider', () => {
  // ...
  let focusVisibleContext = null

  const ChildComponent = () => {
    focusVisibleContext = useContext(FocusVisibleContext)
    return null
  }

  // ...

  afterEach(() => {
    // ...
    focusVisibleContext = null
  })

  it('Should provide "false" for hadKeyboardEvent value on mounting', () => {
    expect(focusVisibleContext).toEqual(false)
  })

  // ...
}
```

What do you think?

nickoola: Yes you right it'll be clearer to only check the context value without using a DOM element.

src/_utils/focusVisibleProvider/index.unit.tsx
nickoola: I do it this way not to refer directly to `document.body` when I attach it to the mount() function. I followed this example:
https://fr.reactjs.org/docs/test-utils.html
And I need to refer to a global element to fire events.

src/_utils/focusVisibleProvider/index.unit.tsx
nickoola: Done

src/_utils/focusVisible/index.tsx
nickoola: I don't really have strong opinion for this point so I'll fallback with:
```
type FocusVisibleProviderProps = {
  children: ReactNode
}
```

src/_utils/focusVisibleProvider/index.unit.tsx
simonrelet: I see, I think they use this in because they rely on `ReactDOM.render()` in their tests.
Here we currently use enzyme to handle the rendering and looking at their doc we don't need it.

For the event, doesn't `document.body.dispatchEvent()` work?

nickoola: Oh you right, no need to attach anything to the mount() function.
I previously try:
```
wrapper = mount(
      <FocusVisibleProvider>
        <ChildComponent />
      </FocusVisibleProvider>,
      { attachTo: document.body },
    )
```
Witch works but trigger some warnings.
Anyway your solution is cleaner :-)

gabrielagabriel:
Could you "illustrate" this PR? Thanks

PR #660, opened 2020-04-10
Search form with autocomplete

wakooka:
- Added `autocompleteFrom` & `autocompleteTo` props
- Factorised `AutocompleteExample` for storybook
- Removed duplicated tests

### DIFF
--- a/src/searchForm/SearchForm.tsx
+++ b/src/searchForm/SearchForm.tsx
@@ -6,8 +6,7 @@ import { canUseDOM } from 'exenv'
 import { color } from '_utils/branding'
 import SearchIcon from 'icon/searchIcon'
 import { CalendarIcon } from 'icon/calendarIcon'
-import { Bullet } from 'index'
-import { BulletTypes } from 'bullet'
+import Bullet, { BulletTypes } from 'bullet'
 import { StandardSeat } from 'icon/standardSeat'
 import { DoubleArrowIcon } from 'icon/doubleArrowIcon'
 import Divider from 'divider'
@@ -23,8 +22,8 @@ import AutoCompleteSection from './autoComplete/section'
 export interface SearchFormProps {
   className?: string
   onSubmit: Function
-  autocompleteFromProps: AutocompleteProps
-  autocompleteToProps: AutocompleteProps
+  autocompleteFrom: React.ReactElement
+  autocompleteTo: React.ReactElement
   datepickerProps: FieldProps
   stepperProps: StepperProps
 }
@@ -32,10 +31,6 @@ export interface SearchFormProps {
 interface FieldProps {
   defaultValue?: string | number
   format?: (value: string | number) => string
-}
-
-interface AutocompleteProps extends FieldProps {
-  placeholder: string
 }
 
 interface StepperProps extends FieldProps {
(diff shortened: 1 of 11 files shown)


src/searchForm/SearchForm.tsx
```diff
@@ -144,9 +143,11 @@ const SearchForm = ({
             className="kirk-search-button"
             onClick={() => setElementOpened(Elements.AUTOCOMPLETE_FROM)}
           >
-            <Bullet type={BulletTypes.SEARCH} className="kirk-bullet--searchForm" />{' '}
+            <span className="kirk-bullet--searchForm">
+              <Bullet type={BulletTypes.SEARCH} />
+            </span>
             <TextBody>
-              {formValues[Elements.AUTOCOMPLETE_FROM] || autocompleteFromProps.placeholder}
+              {autocompleteFromValue?.item.label || autocompleteFrom.props.placeholder}
```
wakooka, line 168: I need to check if this is polyfilled

mcellier: As we don't use babel on kirk, i'm not sure. But since we build the app in kairos with babel, I don't think it'll be a problem for us.

src/searchForm/SearchForm.tsx
```diff
@@ -54,18 +49,18 @@ enum Elements {
...
-  [key in keyof typeof Elements]?: string | number | boolean
+  [key in keyof typeof Elements]?: string | number | boolean | AutocompleteOnChange
 }
 
 const SearchForm = ({
   className,
   onSubmit,
-  autocompleteFromProps,
-  autocompleteToProps,
+  autocompleteFrom,
```
grosbouddha, line 58: it could be nice to add some assertions to make sure this autocomplete are really autocompletes and not some ReactElements.

src/searchForm/SearchForm.tsx
wakooka: Checked, working in SPA on IE 11 👌

src/searchForm/autoComplete/overlay/AutoCompleteOverlay.tsx
```diff
@@ -1,15 +1,27 @@
...
+
+export const AutoCompleteOverlay = (props: AutoCompleteOverlayProps) => {
   const icon = (
     <div className="kirk-autoComplete-icon">
       <Bullet type={BulletTypes.SEARCH} />
     </div>
   )
 
-  return <AutoComplete {...props} inputAddon={icon} embeddedInSearchForm />
+  return React.cloneElement(props.autocompleteComponent, {
```
simonrelet, line 19: I find cloning element to add props error prone because from the cloned component point of view it's hard to know where the props come from. You can't search for the component usages in the project to find where props are passed ([grep test](http://jamie-wong.com/2013/07/12/grep-test/)).

They must also declare those props optional in order to be used as `prop={<MyComponent />}` even if they are required for the component to work.

A solution would be to use the render prop pattern to delegate the rendering:

```tsx
export const AutoCompleteOverlay = (props: AutoCompleteOverlayProps) => {
  // Caller
  return props.renderAutocompleteComponent({
    ...props,
    autoFocus: true,
    inputAddon: icon,
    embeddedInSearchForm: true,
  })
}

// Usage
<AutoCompleteOverlay
  renderAutocompleteComponent={(props) => (
    // Props are applied explicitly
    <AutoCompleteExample {...props} placeholder="Leaving from" />
  )}

  // It is event more flexible as you apply the props as you want
  renderAutocompleteComponent={(props) => (
    // The span wrapper doesn't get the props
    <span>
      <AutoCompleteExample {...props}  placeholder="Leaving from" />
    </span>
  )}
/>
```

What do you think?

wakooka: Yes much nicer indeed. I'll do this in the next PR. Thanks!

src/searchForm/SearchForm.tsx
simonrelet: There is a way:

```tsx
// Import the component (the function)
import AutoComplete from 'autoComplete'

// React elements have a `type` attribute referencing either the native HTML tag (`'span'`, `'div'`)
// or the component (`function() { /* ... */ }`)
const isAutoComplete = autocompleteFrom.type === AutoComplete
```

The issue is that you can't compose the component any more:

```tsx
<SearchForm
  autocompleteFrom={(
    <span>
      <AutoComplete />
    <span>
  )}
/>
```

Nor extend it:


```tsx
<SearchForm autocompleteFrom={<AutoCompleteVariant />} />
```

wakooka: Thanks, interesting. It won't work for us then because we compose the AutoComplete in the SPA with another component

PR #663, opened 2020-04-15
UI fixes on SearchForm

wakooka:


src/phoneField/__snapshots__/PhoneField.unit.tsx.snap
```diff
@@ -723,6 +724,7 @@ exports[`PhoneField should have the expected markup in the DOM with default sett
 
 .c3 input {
   padding-right: 0;
+  font-family: inherit;
```
simonrelet, line 727: Don't we have some sort of [CSS reset](https://meyerweb.com/eric/tools/css/reset/) to handle this?

wakooka: We do, but only in the SPA. I added it here so that we have the same result in storybook

PR #668, opened 2020-04-20
Update hero section to accept search form

simonrelet:
The `HeroSection` component now support 2 contents:
- A link to the search page (deprecated)
- The `SearchForm` component

**DEMO**

Desktop: 
Screenshot 2020-04-20 at 15 48 12

Mobile: 
Screenshot 2020-04-20 at 15 48 35

**TESTED**: Unit test and manually on Chrome, Firefox and Safari

### DIFF
--- a/src/layout/section/heroSection/heroSection.tsx
+++ b/src/layout/section/heroSection/heroSection.tsx
@@ -1,17 +1,17 @@
 import React from 'react'
-import Button from 'button'
+import Button, { ButtonProps } from 'button'
 import Title from 'title'
 import Text, { TextTagType } from 'text'
 import cc from 'classcat'
 
 export interface HeroSectionProps {
-  readonly className?: Classcat.Class
+  readonly className?: string
   readonly heroImageUrl: string
   readonly heroImageUrlLarge: string
   readonly heroText?: string
   readonly heroDescription?: string
   readonly buttonText?: string
-  readonly buttonHref?: string | JSX.Element
+  readonly buttonHref?: ButtonProps['href']
 }
 
 function HeroSection({
--- a/src/layout/section/heroSection/heroSection.tsx
+++ b/src/layout/section/heroSection/heroSection.tsx
@@ -1,17 +1,17 @@
-import React from 'react'
 import Button, { ButtonProps } from 'button'
-import Title from 'title'
-import Text, { TextTagType } from 'text'
-import cc from 'classcat'
+import React from 'react'
+import TextDisplay1 from 'typography/display1'
+import TextTitle from 'typography/title'
 
-export interface HeroSectionProps {
-  readonly className?: string
-  readonly heroImageUrl: string
-  readonly heroImageUrlLarge: string
-  readonly heroText?: string
-  readonly heroDescription?: string
-  readonly buttonText?: string
-  readonly buttonHref?: ButtonProps['href']
+export type HeroSectionProps = {
+  className?: string
+  heroImageUrl: string
+  heroImageUrlLarge: string
+  heroText?: string
+  heroDescription?: string
+  buttonText?: string
+  buttonHref?: ButtonProps['href']
+  bottomElement?: React.ReactElement
 }
 
 function HeroSection({
@@ -20,22 +20,28 @@ function HeroSection({
   heroDescription,
   buttonText,
   buttonHref,
+  bottomElement,
 }: HeroSectionProps) {
   return (
-    <div className={cc(className)}>
+    <div className={className}>
       <div className="hero-image" aria-hidden="true" />
-      <div className="hero-info wrapper wrapper--large">
+
+      <div className="hero-info">
         <div className="hero-info-title-wrapper">
-          <Title className="hero-info-title">{heroText}</Title>
+          <h1 className="hero-info-title">
+            <TextDisplay1 className="hero-info-title--display1">{heroText}</TextDisplay1>
+          </h1>
+
           {heroDescription && (
-            <Text tag={TextTagType.PARAGRAPH} className="hero-info-text">
-              {heroDescription}
-            </Text>
+            <p className="hero-info-text">
+              <TextTitle className="hero-info-text-title">{heroDescription}</TextTitle>
+            </p>
           )}
         </div>
-        <Button className="hero-button" href={buttonHref}>
-          {buttonText}
-        </Button>
+
+        <div className="hero-content">
+          {bottomElement ?? <Button href={buttonHref}>{buttonText}</Button>}
+        </div>
       </div>
     </div>
   )
--- a/src/layout/section/heroSection/heroSection.tsx
+++ b/src/layout/section/heroSection/heroSection.tsx
@@ -1,4 +1,4 @@
-import React, { PureComponent } from 'react'
+import React from 'react'
 import Button from 'button'
 import Title from 'title'
 import Text, { TextTagType } from 'text'
@@ -14,28 +14,31 @@ export interface HeroSectionProps {
   readonly buttonHref?: string | JSX.Element
 }
 
-class HeroSection extends PureComponent<HeroSectionProps> {
-  render() {
-    const { className, heroText, heroDescription, buttonText, buttonHref } = this.props
-    return (
-      <div className={cc(className)}>
-        <div className="hero-image" aria-hidden="true" />
-        <div className="hero-info wrapper wrapper--large">
-          <div className="hero-info-title-wrapper">
-            <Title className="hero-info-title">{heroText}</Title>
-            {heroDescription && (
-              <Text tag={TextTagType.PARAGRAPH} className="hero-info-text">
-                {heroDescription}
-              </Text>
-            )}
-          </div>
-          <Button className="hero-button" href={buttonHref}>
-            {buttonText}
-          </Button>
+function HeroSection({
+  className,
+  heroText,
+  heroDescription,
+  buttonText,
+  buttonHref,
+}: HeroSectionProps) {
+  return (
+    <div className={cc(className)}>
+      <div className="hero-image" aria-hidden="true" />
+      <div className="hero-info wrapper wrapper--large">
+        <div className="hero-info-title-wrapper">
+          <Title className="hero-info-title">{heroText}</Title>
+          {heroDescription && (
+            <Text tag={TextTagType.PARAGRAPH} className="hero-info-text">
+              {heroDescription}
+            </Text>
+          )}
         </div>
+        <Button className="hero-button" href={buttonHref}>
+          {buttonText}
+        </Button>
       </div>
-    )
-  }
+    </div>
+  )
 }
 
 export default HeroSection
--- a/src/layout/section/heroSection/index.tsx
+++ b/src/layout/section/heroSection/index.tsx
@@ -1,123 +1,195 @@
-import styled from 'styled-components'
-import { responsiveBreakpoints, componentSizes, color, space } from '_utils/branding'
+import styled, { css } from 'styled-components'
+import { responsiveBreakpoints, componentSizes, color, space, font } from '_utils/branding'
+import HeroSection, { HeroSectionProps } from './heroSection'
 
-import HeroSection from './heroSection'
+// Remove these deprecated styles once all the button content parts have been removed.
+const DEPRECATED_STYLES = css`
+  & {
+    margin-bottom: 0;
+    height: initial;
+  }
+
+  & .hero-image {
+    position: static;
+    height: 136px;
+    background-image: ${(props: HeroSectionProps) => `url(${props.heroImageUrl})`};
+  }
+
+  & .hero-image::after {
+    content: none;
+  }
+
+  & .hero-info {
+    padding-top: ${space.xl};
+  }
+
+  & .hero-info-title--display1 {
+    color: ${color.primaryText};
+  }
+
+  & .hero-info-text-title {
+    color: ${color.secondaryText};
+    font-size: ${font.base.size};
+    line-height: ${font.base.lineHeight};
+  }
+
+  & .hero-content {
+    align-items: center;
+  }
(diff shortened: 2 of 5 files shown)


src/layout/section/heroSection/story.tsx
```diff
@@ -21,9 +22,54 @@ stories.add('default', () => (
     )}
     heroImageUrlLarge={text(
       'heroImageUrlLarge',
-      'https://cdn.blablacar.com/kairos/assets/build/images/home_summer_campaign-1ea3207605913c1e26410e605a467eb7.jpg',
+      'https://cdn.blablacar.com/kairos/assets/build/images/home_summer_campaign_large-240e0d9fe2123506cca634c2acedce24.jpg',
```
simonrelet, line 25: I changed the image just to have a different one from `heroImageUrl `

src/layout/section/heroSection/story.tsx
```diff
@@ -21,9 +22,54 @@ stories.add('default', () => (
...
+        'heroImageUrl',
+        'https://cdn.blablacar.com/kairos/assets/build/images/home_summer_campaign-1ea3207605913c1e26410e605a467eb7.jpg',
+      )}
+      heroImageUrlLarge={text(
+        'heroImageUrlLarge',
+        'https://cdn.blablacar.com/kairos/assets/build/images/home_summer_campaign_large-240e0d9fe2123506cca634c2acedce24.jpg',
+      )}
+      searchForm={
+        // This should not be needed...
+        // eslint-disable-next-line react/jsx-wrap-multilines
```
simonrelet, line 50: Got this ESlint/prettier conflict. Until we can fix it, I  disabled the rule as workaround.

simonrelet:
@wakooka How do you update the Changelog? Manually or with a tool?

gabrielagabriel:
CHANGELOG.md is updated manually 🧷

src/layout/section/heroSection/heroSection.tsx
```diff
@@ -1,41 +1,48 @@
...
+  return (
+    <div className={className}>
+      <div className="hero-image" aria-hidden="true" />
+
+      <div className="hero-info">
+        <div className="hero-info-title-wrapper">
+          <Title className="hero-info-title">{heroText}</Title>
+
+          {heroDescription && (
+            <Text tag={TextTagType.PARAGRAPH} className="hero-info-text">
```
gabrielagabriel, line 34: Does this component can be updated for any `typography` component? We would like to deprecate `Text`

src/layout/section/heroSection/index.tsx
```diff
@@ -1,123 +1,168 @@
...
   & .hero-info {
-    text-align: center;
-    margin-left: auto;
-    margin-right: auto;
+    width: 100%;
     max-width: ${componentSizes.largeSectionWidth};
-    padding: ${space.xl};
-    padding-top: 0;
+    padding: 0 ${space.xl};
+    padding-top: 120px;
```
gabrielagabriel, line 100: Where this magic number comes from?

src/layout/section/heroSection/index.tsx
```diff
@@ -1,123 +1,168 @@
...
+  & .hero-info-title {
     margin: 0;
+    padding-top: ${space.m};
+    padding-bottom: ${space.m};
+    color: ${color.textWithBackground};
+    font-size: 60px;
+  }
+
+  & .hero-info-text {
+    font-size: 22px;
```
gabrielagabriel, line 119: please, use font variables on `utilis/branding`

src/layout/section/heroSection/index.tsx
```diff
@@ -1,123 +1,168 @@
...
-    align-self: center;
-    display: flex;
-    padding-top: ${space.xl};
-    padding-bottom: ${space.ms};
+  & .hero-info-title {
     margin: 0;
+    padding-top: ${space.m};
+    padding-bottom: ${space.m};
+    color: ${color.textWithBackground};
+    font-size: 60px;
```
gabrielagabriel, line 115: Please use font-sizing

gabrielagabriel: Actually you should use Typography components instead

src/layout/section/heroSection/index.tsx
```diff
@@ -1,123 +1,168 @@
...
     margin-top: ${space.xl};
+    display: flex;
+    flex-direction: column;
+    align-items: center;
   }
 
-  @media (${responsiveBreakpoints.isMediaLarge}) {
+  ${props => (props.searchForm == null ? DEPRECATED_STYLES : null)}
+
+  @media (${responsiveBreakpoints.isMediaSmall}) {
```
gabrielagabriel, line 133: would be nicier as mobile first

src/layout/section/heroSection/index.tsx
```diff
@@ -1,123 +1,168 @@
...
-      left: -50%;
-      right: -50%;
-      z-index: 2;
-      background: linear-gradient(
-        0deg,
-        rgba(0, 0, 0, 0.4) 72%,
-        rgba(0, 0, 0, 0.2) 85%,
-        rgba(0, 0, 0, 0) 100%
-      );
+    & .hero-info-title {
```
gabrielagabriel, line 151: please use Typography components instead

src/layout/section/heroSection/index.unit.tsx
```diff
@@ -20,4 +20,18 @@ describe('HeroSection', () => {
...
+
+  it('should render a serch form', () => {
+    const props: HeroSectionProps = {
+      ...defaultProps,
+      // We don't need to use the _real_ `SearchForm` component as it is only used as a node.
+      searchForm: <form>Search Form</form>,
+    }
+
+    const wrapper = mount(<HeroSection {...props} />)
+    expect(wrapper.find('h1').text()).toContain('hero text')
```
gabrielagabriel, line 32: Avoid select items by tag (implementation details...)

src/layout/section/heroSection/index.tsx
Tuxosaurus: Problem is that the homepage was never fully Pixar compliant, it was copied as-is from the existing page. There's nothing in Pixar that has this size for a Title (this could be discussed with UX)

wakooka: I agree we should not have font-size that is not in the branding. However this is not the goal of this PR and @simonrelet is not introducing this, he just moved it

src/layout/section/heroSection/index.unit.tsx
wakooka: I agree you could do `.find.(Title)`

src/layout/section/heroSection/story.tsx
```diff
@@ -21,9 +22,54 @@ stories.add('default', () => (
...
-      'https://cdn.blablacar.com/kairos/assets/build/images/home_summer_campaign-1ea3207605913c1e26410e605a467eb7.jpg',
+      'https://cdn.blablacar.com/kairos/assets/build/images/home_summer_campaign_large-240e0d9fe2123506cca634c2acedce24.jpg',
     )}
     buttonText={text('buttonText', 'Rechercher un trajet')}
     buttonHref={text('buttonHref', 'http://google.fr')}
   />
 ))
+
+stories.add('search form', () => (
+  <MediaSizeProvider>
```
wakooka, line 33: I don't think you need it, since HereSection does not use it.

gabrielagabriel: And your are doing reponsive on the component itself with media queries

src/layout/section/heroSection/index.tsx
gabrielagabriel: Ok.. can he use the ones that already exist then.. and let only `60px` outside of scale?

simonrelet: I can use the typography components instead but I still need to override some `font-size` and `line-height` values as they only have one size for both small and large screens. The font sizes in the design are not the same for both devices.

From the design (small / large devices):

- "Title": 30px / 60px
- "Text": 18px / 22px

This means:
- "Title" can be a `Display1` (30px) with an override (60px) for large device.
- "Text" can either be a:
  - `SubHeader` (22px) with an override (18px) for small device
  - `Title` (18px) with an override (22px) for large device

What do you think?

gabrielagabriel: Unfortunatelly you must override. Please, start **mobile first** that means `Title`  going up to `22px` 

I also recommend override with _styled-components_ instead of css classes

simonrelet: > I also recommend override with styled-components instead of css classes
> @gabrielagabriel 

I'm not sure to understand, what are you proposing?

src/layout/section/heroSection/index.unit.tsx
grosbouddha: nit and loud thinking: Title is an implementation detail, the rendered dom (h1) is not (we have no choice, we render DOM at the end of the day).
If we start testing our own abstractions instead of the actual rendered output, we will end up writing shallow-like bad tests.

wakooka: While I mostly agree with avoiding shallow testing, in this particular case it also means that the drawback by not using `Title` is I can replace the `Title` by a `<h1>` in my component and the test will pass, while it shouldn't since Title also contains specific style.

simonrelet: If a use a snapshot for this test:
- We no longer need a selector to test the presence of an element
- The styles appear in the snapshot as we use Styled Components so replacing the `Title` component by a `h1` would break that part.

What do you think?

src/layout/section/heroSection/story.tsx
simonrelet: The `SearchForm` component used in the story needs it.

src/layout/section/heroSection/heroSection.tsx
```diff
@@ -1,41 +1,48 @@
...
-  readonly buttonHref?: string | JSX.Element
+export type HeroSectionProps = {
+  className?: string
+  heroImageUrl: string
+  heroImageUrlLarge: string
+  heroText?: string
+  heroDescription?: string
+  buttonText?: string
+  buttonHref?: ButtonProps['href']
+  searchForm?: React.ReactElement
```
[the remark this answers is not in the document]
simonrelet: I agree, the fact that this component is only used in one place and contain either a button or the search form led to this name, but from the component point of view a `bottomElement`, `children` or even `contentElement` would be better.

@wakooka Any proposition on this?

src/layout/section/heroSection/index.tsx
simonrelet: I can do that but there will be a bit more style overrides for the deprecated part.

src/layout/section/heroSection/index.tsx
gabrielagabriel:
```
const HeroTitle = styled(Title)`
// heroTitle styles here
`
```

src/layout/section/heroSection/index.tsx
gabrielagabriel: :(

src/layout/section/heroSection/story.tsx
gabrielagabriel: We already have a responsive component available on Storybook, no need to show it twice

src/layout/section/heroSection/index.unit.tsx
grosbouddha: Snapshot tests are not very focused and it's easy to not see all the diffs when we refactor heavily tests. Non-snapshot unit tests allow us to have a more detailed list of use cases that we want to test. For these reasons, we should try to minimize the use of snapshot tests; a good balance is too have a lot of non-snapshot unit tests  + a few snapshot tests to test the default high level states (like smoke integration tests).

About the original issue about selecting using Title or h1, I would still prefer h1 since this is the real thing I want to test the existence of (not some intermediate abstractions).

All in all, both are fine too me.

src/layout/section/heroSection/index.tsx
simonrelet: I'll still do it as it only affect the deprecated part.

src/layout/section/heroSection/story.tsx
simonrelet: > We already have a responsive component available on Storybook, no need to show it twice
> @gabrielagabriel 

Which one?
The `SearchForm` component need the React provider in order to work correctly, otherwise only the CSS is updated (media query).

gabrielagabriel: Oh.. if its an adaptative layout you are right! 👍

src/layout/section/heroSection/index.unit.tsx
simonrelet: I'll keep the `h1` for now as the other test is also using it, but I'll be happy to investigate a better solution for the tests.

src/layout/section/heroSection/index.tsx
simonrelet: [Looking at the design](https://sketch.cloud/s/2bKDG/a/Q5Y2Em#Inspector), the content is not vertically centered, it's a bit lower so the woman's face isn't hidden by the text, this padding shifts the content vertically to fit the design (in 1440 x 960) and other sizes that are not represented by the design but still exist.

This is clearly not perfect, if you have a better suggestion I'll take it 😁.

src/layout/section/heroSection/heroSection.tsx
simonrelet: Done.

src/layout/section/heroSection/index.tsx
simonrelet: I kept the class names as there were until we agree on the way to use styled-components.

PR #670, opened 2020-04-20
Use focus & blur for improving keyboard navigation in search form

wakooka:


wakooka:
Didn't use forwardedProps as discussed because I still need to know what was the element that blurred. I think now there's a need for an Overlay component which will wrap this blur logic. I might do this in a next PR, wdyt?

src/stepper/Stepper.tsx
```diff
@@ -196,6 +199,7 @@ export default class Stepper extends PureComponent<StepperProps, StepperState> {
           status={ButtonStatus.UNSTYLED}
           disabled={isMin}
           isBubble
+          focus={!isMin && isMax && focusFirstInteractiveElement}
```
simonrelet, line 202: If `props.value` is strictly between `min` and `max` it looks like none of the buttons will be focused, is this expected?

src/searchForm/autoComplete/overlay/AutoCompleteOverlay.tsx
```diff
@@ -14,24 +14,38 @@ export type AutoCompleteOverlayProps = Omit<
   readonly renderAutocompleteComponent: (
     props: Omit<AutoCompleteProps, 'searchForItems'>,
   ) => JSX.Element
+  readonly closeOnBlur: () => void
```
simonrelet, line 17: Just saw that the `readonly` (and `Readonly<T>`) are not required for React components' props and state because they already apply it.

[Explained here](https://raulfdm.github.io/react-typescript-cheatsheet/docs/basic/getting-started/class_components) under the question "No need for readonly".

src/searchForm/autoComplete/overlay/AutoCompleteOverlay.tsx
```diff
@@ -14,24 +14,38 @@ export type AutoCompleteOverlayProps = Omit<
...
   )
-
-  return renderAutocompleteComponent({
-    ...props,
-    autoFocus: true,
-    inputAddon: icon,
-    embeddedInSearchForm: true,
-  })
+  return (
+    <div
```
simonrelet, line 33: No `className`?
`DatePickerOverlay` and `StepperOverlay` have one.

src/searchForm/autoComplete/overlay/AutoCompleteOverlay.tsx
```diff
@@ -14,24 +14,38 @@ export type AutoCompleteOverlayProps = Omit<
...
   ) => JSX.Element
+  readonly closeOnBlur: () => void
 }
 
 export const AutoCompleteOverlay = ({
   renderAutocompleteComponent,
+  closeOnBlur,
   ...props
 }: AutoCompleteOverlayProps) => {
+  const overlayRef = useRef(null)
```
simonrelet, line 25: You can write it `useRef<HTMLDivElement>(null)` otherwise Typescript infers it as `React.MutableRefObject<any>`, and cast the `evt.relatedTarget as Node`

simonrelet:
> Didn't use forwardedProps as discussed because I still need to know what was the element that blurred. I think now there's a need for an Overlay component which will wrap this blur logic. I might do this in a next PR, wdyt?
> @wakooka 

I think having an overlay component is a good idea 👍.

src/searchForm/autoComplete/overlay/AutoCompleteOverlay.tsx
wakooka: Good to know, thanks

src/searchForm/autoComplete/overlay/AutoCompleteOverlay.tsx
wakooka: No need an extra div here, the class is applied directly on the Autocomplete

wakooka: It's applied on `Autocomplete` component directly, but for consistency I'll move it here.

src/stepper/Stepper.tsx
wakooka: You're right, was working on my case because by default the minus button was disabled. I'll try to find a better solution

src/datePicker/DatePicker.tsx
```diff
@@ -39,6 +39,7 @@ export interface DatePickerProps {
   readonly toMonth?: Date
   readonly firstDayOfWeek?: number
   readonly stickyPositionTop?: number
+  readonly focus?: boolean
```
simonrelet, line 42: Is this prop still used? I didn't find it's usages 🤔.

src/stepper/Stepper.tsx
```diff
@@ -73,12 +76,16 @@ export default class Stepper extends PureComponent<StepperProps, StepperState> {
...
-    this.ref = React.createRef()
+    this.valueElementRef = React.createRef()
+    this.containerRef = React.createRef()
   }
 
   componentDidMount() {
     this.handleFontSize()
     window.addEventListener('resize', this.handleFontSize.bind(this))
+    if (this.containerRef.current) {
+      this.containerRef.current.focus()
```
simonrelet, line 87: Do we always want to focus when the component is mounted?

simonrelet: Shouldn't this be done by the overlay?

src/stepper/Stepper.tsx
```diff
@@ -187,7 +195,10 @@ export default class Stepper extends PureComponent<StepperProps, StepperState> {
     return (
       <div
         className={cc(['kirk-stepper', `kirk-stepper-${display}`, className])}
+        ref={this.containerRef}
         aria-label={title}
+        // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+        tabIndex={0}
```
simonrelet, line 201: I think if you use `tabIndex={-1}` you don't need the ESlint disable, it would still be able to get the focus programmatically but not by user navigation.

src/stepper/Stepper.tsx
```diff
@@ -196,6 +207,7 @@ export default class Stepper extends PureComponent<StepperProps, StepperState> {
           status={ButtonStatus.UNSTYLED}
           disabled={isMin}
           isBubble
+          focus={!isMin && isMax && focus}
```
simonrelet, line 210: I'm not sure to understand the focus conditions we want here. When do we want to focus the buttons?

src/searchForm/SearchForm.tsx
```diff
@@ -143,30 +145,11 @@ const SearchForm = ({
...
 
   return (
-    <form
-      action=""
-      noValidate
-      className={cc(['kirk-searchForm', className])}
-      ref={container}
-      role="search"
-    >
+    <form action="" noValidate className={cc(['kirk-searchForm', className])} role="search">
```
simonrelet, line 152: I think you can remove the `onClick` listener on the button and replace the `action` prop of the form by a `onSubmit={event => { event.preventDefault() }}`.

A button in a form automatically triggers the form submission when clicked so you only need to _catch_ the submit event and prevent the browser's default action.

src/datePicker/DatePicker.tsx
wakooka: Yes passing it from the DatePickerOverlay.

src/searchForm/SearchForm.tsx
wakooka: Need to keep `action` so that the keyboard on iOS is the "form keyboard" because there's an issue on forms that don't have as far as I remember. 
But I agree with the other part, I'll move to the onSubmit event instead :+1

src/stepper/Stepper.tsx
wakooka: No my mistake I'll add the the condition on the focus prop

src/datePicker/DatePicker.tsx
wakooka: Just added in the condition to focus the datepicker

src/stepper/Stepper.tsx
wakooka: Removed

PR #671, opened 2020-04-20
Add a max size to the QR card to see the image in small desktops

florianeyzat:


pith:
I agree with @gabrielagabriel. I don't know why these size are defined outside of the components. 🤔 
Can you also add a comment on the 450px to explain: As there was no spec for desktop we used this max-width as a pragmatic fix.

wakooka:
I also agree, sorry @florianeyzat I asked you to put the size there. But now that I look at it the other sizes in `componentSizes `are used multiple places, but there's no need for your component to share this size.
Can you move the size?
Also update the CHANGELOG.md

florianeyzat:
Sure @gabrielagabriel @pith I actually did that after a talk with @wakooka but I'm okay with this way ! 😄

PR #674, opened 2020-04-21
[BOYSCOUT] FIX reverse import by creating a Component Object Model

gabrielagabriel:
## Description

Avoid importing index on SuccessModal

## What has been done

Export **SuccessModal Elements** for testing purposes

## Things to consider

- Styled-components syntaxe

## How was it was tested

Local


src/successModal/elements.tsx
```diff
@@ -0,0 +1,112 @@
...
+const SuccessAction = styled.div`
+  padding: ${space.xl};
+`
+
+const SuccessButton = styled(Button)`
+  justify-content: center;
+
+  & {
+    min-width: 8rem;
+  }
```
simonrelet, line 99: I think the `& {}` part can be removed to have:
```css
justify-content: center;
min-width: 8rem;
```

src/successModal/index.tsx
```diff
@@ -1,115 +1,6 @@
...
-const SuccessModalStyle = {
-  StyledSuccessModal,
-  Media,
-  Figure,
-  Content,
-  SuccessTitle,
-  SuccessAction,
-  SuccessButton,
-}
+import SuccessModalElements from './elements'
```
simonrelet, line 2: You can simplify the import/export to:

```ts
export { default, SuccessModalProps } from './SuccessModal'
export { default as SuccessModalElements } from './elements'
```

src/successModal/elements.tsx
gabrielagabriel: not for `min-width` & here increase specificity

gabrielagabriel: @simonrelet  I couldn't remove it from Button and that's why I'm increasing specificity, better than a important, for sure 🙅‍♀️

PR #676, opened 2020-04-21
Fix breakpoints values to make them exclusive betweens small & large

wakooka:
## Description

The issue was that depending on how we write media queries based on small or large breakpoints, the rendering would differ when the screen is 800px. 
It also broke the mediaSizeProvider which we could in certain case like in the search form returning large instead of small.
See example below: rendering on 800px screen, we should have Divider between the fields which are based on MediaSizeProvider returning large
image (8)

## What has been done

Have exclusive values for large and small fixes the issue

## Things to consider

This also makes it consistent with the logic we have in the SPA on breakpoints.

## How was it was tested

Tested locally on storybook

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-- [...]
+- **[FIX]** Fix breakpoints values to make them exclusive betweens small & large
 
 # v29.1.0 (21/04/2020)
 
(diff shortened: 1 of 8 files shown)


wakooka:
> We should have mobile first in mind that way we don't even need **isMediaSmall**

@gabrielagabriel I'm ok to do everything mobile first in a next PR, but this was supposed to be a quick fix for an issue we currently have. I agree with what you suggest but it would require a bigger refacto and testing to make sure inverting media queries won't break the layout.
Also the benefit of having 2 breakpoints is to lessen the override see this file for example
https://github.com/blablacar/ui-library/blob/master/src/layout/section/cardsSection/index.tsx

PR #681, opened 2020-04-23
Add focus trap & handling overflow on `searchForm` section components

wakooka:
## Description
Keep the focus trapped in Search Form section since they're like "modals".
Search form sections now handles their own scroll.
 
## What has been done
- Created a custom hooks `useFocusTrap`

## Things to consider
I couldn't find a way to put back the focus on the active element if we click on the back button in Autocomplete section for example. Maybe @simonrelet you have an idea. I'd like to keep the logic in `useFocusTrap`

## How was it was tested
Locally in Storybook


src/_utils/useOverflowHiddenOnDocument/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+const DEFAULT_OVERFLOW_VALUE = 'visible'
+const OVERFLOW_HIDDEN_VALUE = 'hidden'
+
+/**
+ * This custom hook will remove the overflow visible from the document
+ * It's useful for modal or any screen that wants to handle their own scrollbar
+ * WARNING: For now it doesn't handle 2 components using this hooks at the same time
+ */
+export const useOverflowHiddenOnDocument = () => {
```
grosbouddha, line 11: Not sure why we use a hook/useEffect for that.
It's can be a simple helper static function, no need to tight it to the component rendering lifecycle.

src/_utils/useFocusTrap/index.tsx
```diff
@@ -0,0 +1,35 @@
+import { useEffect, useRef } from 'react'
+import createFocusTrap from 'focus-trap'
+
+const ESCAPE_KEY = 'Escape'
```
simonrelet, line 4: Don't we have a key mapping in the utils?

src/_utils/useOverflowHiddenOnDocument/index.tsx
wakooka: You're right, I was on my to create custom hooks so I kept going.
I'll create a static function with a condition on being run on client side.

src/_utils/useFocusTrap/index.unit.tsx
```diff
@@ -0,0 +1,52 @@
...
+          <button type="button">3</button>
+        </div>
+      )
+    }
+  })
+
+  it('Should activate focus trap', () => {
+    mount(<TestComponent />)
+    expect(createFocusTrap).toHaveBeenCalled()
+    expect(activate).toHaveBeenCalled()
```
simonrelet, line 35: Can we also check that the first button has the focus?

src/_utils/useOverflowHiddenOnDocument/index.tsx
simonrelet: We can have something like:
```tsx
function hideOverflowOnDocument() {
  document.querySelector('html').style.overflow = OVERFLOW_HIDDEN_VALUE

  // Return a "clean up/cancel" callback
  return () => {
    document.querySelector('html').style.overflow = DEFAULT_OVERFLOW_VALUE
  }
}

function MyComponent() {
  useEffect(hideOverflowOnDocument)
  // ...
}
```

What do you think?

src/_utils/useOverflowHiddenOnDocument/index.tsx
```diff
@@ -0,0 +1,20 @@
+import { useEffect } from 'react'
+
+const DEFAULT_OVERFLOW_VALUE = 'visible'
+const OVERFLOW_HIDDEN_VALUE = 'hidden'
```
simonrelet, line 4: Do we really need these constants as there are only used once and not exported?

src/autoComplete/AutoComplete.tsx
```diff
@@ -25,10 +25,10 @@ export interface AutoCompleteProps {
...
   readonly onBlur?: (event: React.FocusEvent<HTMLInputElement>) => void
   readonly onClear?: () => void
-  readonly className?: Classcat.Class
-  readonly inputClassName?: Classcat.Class
-  readonly itemClassName?: Classcat.Class
-  readonly bodyClassName?: Classcat.Class
+  readonly className?: string
+  readonly inputClassName?: string
+  readonly itemClassName?: string
+  readonly bodyClassName?: string
```
simonrelet, line 31: 👌

src/_utils/useOverflowHiddenOnDocument/index.tsx
grosbouddha: The hook is not extremely useful as a globally shareable hook. I'm afraid we end up with a bunch of confusing barely-reusable global hooks.
I like Simon alternative or you could have a folder-local hook (but if we use it once, why create a separate file/function).

simonrelet: How about integrating it _in_ `useFocusTrap`?
I don't have a use case where we want to `useFocusTrap` without preventing a scroll on the document 🤔.

src/_utils/useOverflowHiddenOnDocument/index.tsx
wakooka: I feel using this constants make it clear which one is the default value.

src/_utils/useOverflowHiddenOnDocument/index.tsx
wakooka: > How about integrating it _in_ `useFocusTrap`?
> I don't have a use case where we want to `useFocusTrap` without preventing a scroll on the document 🤔.

I thought of that at first but ended making two separate hooks. I think this is the right approach, I don't see any case where we would have a focusTrap without preventing the scroll on document. I'll add it directly in `usefocusTrap`

src/_utils/useFocusTrap/index.unit.tsx
wakooka: I don't want to test focus-trap lib, I just want to check that the correct methods are called.

src/_utils/useFocusTrap/index.tsx
wakooka: Waiting for #659 to be merged first, I'll add it there. Conflicts incoming :D

src/_utils/useFocusTrap/index.unit.tsx
simonrelet: I don't see it as testing the focus-trap lib but as testing the functionality of our component. The fact that we use a lib is an implementation detail.

wakooka: I'll try, I'll remove the mock and check if it's possible

src/_utils/useOverflowHiddenOnDocument/index.tsx
wakooka: I removed `useOverflowHiddenOnDocument ` and integrated the logic in useFocusTrap

src/_utils/useFocusTrap/index.unit.tsx
wakooka: As seen together we can't test this due to the test environment and a condition in focus-trap lib on the .offsetParent (which is null in our test)

PR #682, opened 2020-04-23
Add specifications for hint

pith:
## Description

Add markdown specification for the hint component

## What has been done

image

image

## How was it was tested

Tested locally in storybook

### DIFF
--- a/src/hint/specifications/hint.md
+++ b/src/hint/specifications/hint.md
@@ -1,0 +1,33 @@
+# Hint
+
+Use it to show the member there’s a new action they can take on this page. Only use for actions that previously didn’t exist on this page.
+
+<!-- STORY -->
+
+## Usage
+
+`TextComponents` are exported individually.
+
+```js
+import Hint from '@blablacar/ui-library/build/hint'
+
+;<Hint
+  title="Title"
+  description="A longer description"
+  closeButtonTitle="Close button"
+  onClose={onClose}
+  hidden={isHidden}
+>
+  {a11yAttrs => <ItemChoice leftAddon={<QrCodeIcon />} label="Your tickets" {...a11yAttrs} />}
+</Hint>
+```
+
+## Props
+
+| **Props**          | **Value**        | **Description**                     |
+| ------------------ | ---------------- | ----------------------------------- |
+| `title`            | `string`         | Title content                       |
+| `description`      | `string`         | Description content                 |
+| `closeButtonTitle` | `string`         | A11y content for the close button   |
+| `position`         | `ABOVE`\|`BELOW` | Define how the content is displayed |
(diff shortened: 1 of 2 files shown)


src/hint/specifications/hint.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Props
+
+| **Props**          | **Value**        | **Description**                     |
+| ------------------ | ---------------- | ----------------------------------- |
+| `title`            | `string`         | Title content                       |
+| `description`      | `string`         | Description content                 |
+| `closeButtonTitle` | `string`         | A11y content for the close button   |
+| `position`         | `ABOVE`\|`BELOW` | Define how the content is displayed |
+| `hidden`           | `true`\|`false`  | The component is render when false  |
```
simonrelet, line 33: I suppose we can't generate this table from the prop-types? 😅

src/hint/specifications/hint.md
```diff
@@ -0,0 +1,33 @@
+# Hint
+
+Use it to show the member there’s a new action they can take on this page. Only use for actions that previously didn’t exist on this page.
```
wakooka, line 3: It would be good to have something also describing when the action is **not considered as "new" anymore**. Then maybe we could even programaticaly shows/hide the component during a certain period of time. Or just making sure that we remove it when it's not "new" anymore. I'll talk to Jonas :)

wakooka:
Thanks 👍

src/hint/specifications/hint.md
pith: That's a good point, I should also talk about it with Alban. Because it might stay new for several months with the situation (half trolling). But we will forget it for sure.

pith: We could implement an expirationDate that would hide the component and log a warning for the devs.

pith:
Oups I merged it in another PR 🤦

src/hint/specifications/hint.md
pith: I think the last time we discussed this it was not possible with TypeScript.

src/hint/specifications/hint.md
grosbouddha: Or implement a proper user backend preference/flag store :) ? The expiration date strategy makes assumption about the user visit pattern and is some client business logic (spread in 3 clients).

wakooka: > implement a proper user backend preference/flag store
This should be driven by the product team depending on what they want to have:
- is the hint component used on a user-centric point of view? e.g. This new action will be considered new until the user sees it 3 times, need to be in the backend.
- is the hint component used on a product point of view only? e.g. This action is considered new in the product for X months (regardless of the user), no need for backend.

PR #684, opened 2020-04-24
Add new ItemEditableInfo

grosbouddha:
## Description

Create new ItemEditableinfo widget

image

## What has been done

Extending the base item and adding a readonly mode for the ItemEditableinfo non-editable case.

## How was it was tested

Locally in storybook


src/itemEditableInfo/index.tsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react'
+
+import { TextDisplayType } from 'text'
+import Item from '_utils/item'
+import { A11yProps, pickA11yProps } from '_utils/interfaces'
+import { color } from '_utils/branding'
+
+export interface ItemEditableInfoProps extends A11yProps {
```
gabrielagabriel, line 8: Are we keeping interfaces instead of types?

src/itemEditableInfo/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+const ItemEditableInfo = (props: ItemEditableInfoProps) => {
+  const { className, label, value, href = null, disabled = false } = props
+  const a11yAttrs = pickA11yProps<ItemEditableInfoProps>(props)
+
+  const extraProps = {
+    isClickable: true,
+    href,
+    // For the ItemEditableInfo, the value (entered previously by the user)
+    // is the most important info and is visually bigger than the
+    // label for it.
```
gabrielagabriel, line 31: Not possible to replace for Typography?

src/itemEditableInfo/index.unit.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import { mount } from 'enzyme'
+
+import ItemEditableIInfo from 'itemEditableInfo'
```
Tip-Sy, line 4: Typo: `ItemEditableInfo`
(fix below as well)

src/itemEditableInfo/index.unit.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+it('Should render a basic ItemEditableInfo', () => {
+  const props = {
+    label: 'My label',
+    value: 'My value',
+    href: '/myhref',
+  }
+  const wrapper = mount(<ItemEditableIInfo {...props} />)
+  const anchor = wrapper.find('a')
```
gabrielagabriel, line 13: Avoid targeting tags

src/itemEditableInfo/index.unit.tsx
Tip-Sy: - [ ] fix typo

src/itemEditableInfo/index.tsx
grosbouddha: I need the old a11yProps pick mechanism :/

src/itemEditableInfo/index.tsx
grosbouddha: I would love to but I will need to recreate a new Item implmentation.
We need a separate project and PRs for that.

src/itemEditableInfo/index.unit.tsx
grosbouddha: humm I do not agree, I need to verify that we have an anchor with a href and a good value.

src/itemEditableInfo/index.unit.tsx
```diff
@@ -0,0 +1,28 @@
...
+  expect(anchor.length).toBe(1)
```
wakooka, line 14: nit: toHaveLength(1)

src/itemEditableInfo/index.unit.tsx
grosbouddha: Done

src/itemEditableInfo/index.unit.tsx
```diff
@@ -0,0 +1,28 @@
...
+})
+
+it('Should render a non-editable ItemEditableInfo', () => {
+  const props = {
+    label: 'My label',
+    value: 'My value',
+    disabled: true,
+  }
+  const wrapper = mount(<ItemEditableIInfo {...props} />)
+  expect(wrapper.find('a').length).toBe(0)
```
wakooka, line 26: nit: `expect(wrapper.find('a).exists()).toBe(false)`

src/itemEditableInfo/index.tsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react'
+
+import { TextDisplayType } from 'text'
+import Item from '_utils/item'
+import { A11yProps, pickA11yProps } from '_utils/interfaces'
+import { color } from '_utils/branding'
+
+export interface ItemEditableInfoProps extends A11yProps {
+  readonly className?: Classcat.Class
```
wakooka, line 9: After speaking with @simonrelet we have no reason to have this type, `className` should always be a string
`readonly className?: string`

src/itemEditableInfo/index.unit.tsx
wakooka: I agree with @grosbouddha on this one but I think there's a need for a conversation on how we write test.

src/itemEditableInfo/index.unit.tsx
grosbouddha: Done

src/itemEditableInfo/index.unit.tsx
grosbouddha: Done

src/itemEditableInfo/index.tsx
simonrelet: This should work:

```ts
export type ItemEditableInfoProps = A11yProps & {
  // ...
}
```

src/itemEditableInfo/index.unit.tsx
simonrelet: I added an entry for this in the next chapter weekly minutes. I only wrote my name but everyone is free to speak.

src/itemEditableInfo/index.tsx
grosbouddha: I updated it to the above, it works ! Thanks folks !

grosbouddha:
FYI, I renamed the  **disabled** props to **readonly** which is closer to the widget spec.

PR #687, opened 2020-04-24
Define 's justify-content to be centered only in large screens

florianeyzat:


src/layout/section/mediaContentSection/__snapshots__/index.unit.tsx.snap
```diff
@@ -137,8 +137,8 @@ exports[`MediaContentSection should render default MediaContentSection section 1
   -webkit-box-align: center;
   -ms-flex-align: center;
   align-items: center;
-  -webkit-transition: max-width 500ms ease-in, background-color 200ms ease-in;
-  transition: max-width 500ms ease-in, background-color 200ms ease-in;
+  -webkit-transition: max-width 500ms ease-in,background-color 200ms ease-in;
```
gabrielagabriel, line 140: We gotta decide if spaces or not are the rules for snapshots.
Ping @wakooka

florianeyzat: Yes I have an issue with them here..

florianeyzat: Basically my snapshots update something different than our CI, I'm going to force push the previous version so we don't see the issue here

wakooka: I sent a message in slack on this: `npm run test -- --clearCache` You shouldn't have the issue after that

PR #691, opened 2020-04-24
Simplify CSS prefix function

simonrelet:
## Description

Simplify the `prefix` usages for custom prefixes.

## What has been done

The `baseClass` parameter of `prefix` is now added to the `'kirk'` prefix instead of replacing it.

```ts
const BASE_CLASSNAME = prefix({ 'button-group': true })
// ↓↓↓
const BASE_CLASSNAME = 'button-group'
```

## Things to consider

I also split the tests into specific unit tests.

## How it was tested

Unit tests.

### DIFF
--- a/src/_utils/index.tsx
+++ b/src/_utils/index.tsx
@@ -1,12 +1,18 @@
 import React, { Fragment } from 'react'
 import { canUseDOM } from 'exenv'
 
-export const prefix = (
-  modifiers: Classcat.ClassObject = {},
-  baseClass: string = 'kirk',
-): string => {
+const DEFAULT_CLASS_PREFIX = 'kirk'
+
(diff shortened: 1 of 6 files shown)


src/_utils/index.tsx
```diff
@@ -1,12 +1,18 @@
 import React, { Fragment } from 'react'
 import { canUseDOM } from 'exenv'
 
-export const prefix = (
-  modifiers: Classcat.ClassObject = {},
-  baseClass: string = 'kirk',
-): string => {
+const DEFAULT_CLASS_PREFIX = 'kirk'
+
+export const prefix = (modifiers: Classcat.ClassObject, baseClass?: string): string => {
```
simonrelet, line 6: The `modifiers`  parameter is no longer optional (it was never called as `prefix()` anyway)

PR #693, opened 2020-04-27
Removed color abstraction

wakooka:
## Description

Removed the color abstraction and use the paletter colors directly.

## What has been done

All files using colors have been updated to the new color name.

## How it was tested

Locally in storybook

image


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **[BREAKING CHANGE]** Removed color abstraction and only use palette instead
```
pith, line 3: Are you ready to update Kairos with that breaking change? 😱

wakooka: Not yet, I'll prepare the branch before release ;)

gabrielagabriel:
IMHO we should NOT remove "theme" properties. Dark Theme will be **impossible** if we do that

wakooka:
> IMHO we should NOT remove "theme" properties. Dark Theme will be impossible if we do that
I'm only doing what's been decided during the chapter meeting :/
We also said that we were not against having a Theme, but right now the theme we use is a mess

PR #696, opened 2020-04-28
add disabled state to Items with icons fixes

Tuxosaurus:
We will need to show disabled ItemRadio and ItemCheckboxes as well as independent icons to illustrate the lack of results and the impossibility to filter.

The existing "disabled" prop for Items is used to add a ".kirk-item--disabled" class that overrides the color of all text in Items. It was added as a knob in the stories of all Item-based component.

A new "isDisabled" prop was added to all icons, adding a ".kirk-icon--disabled" class that override the color of the "stroke" attribute in SVGs.

Some icons needed more changes when disabled. Also, while testing overriding both the stroke and fill attributes, I found that some icons had unnecessary elements that stayed transparent until now, I removed them and tweaked other icons for improved usability.  

ItemCheckbox
itemcheckbox_checked
itemcheckbox_checked_disabled

ItemRadio
itemradio_checked
itemradio_checked_disabled

ItemChoice
itemchoice
itemchoice_disabled

Tested locally in Firefox and Chrome for MacOS


pith:
Are the padding variation due to the screenshots or is it expected?

Tuxosaurus:
> Are the padding variation due to the screenshots or is it expected?

Yes, I took the larger screenshot at a different time, with the console on the right set at a different width...

src/_utils/icon/index.tsx
```diff
@@ -14,6 +14,11 @@ const StyledBaseIcon = styled(BaseIcon)`
     box-shadow: ${shadow.icon};
   }
 
+  &.kirk-icon--disabled,
+  &.kirk-icon--disabled * {
+    stroke: ${color.fadedText};
```
gabrielagabriel, line 19: I would rather use styled-components approach and do not apply a class for change the colors based on props

```suggestion
    stroke: ${props => (props.isDisabled ? color.gray : '')};
```

in that case, you only need to pass the property `isDisabled` to the element no need to recalc any class

src/_utils/item/Item.tsx
```diff
@@ -110,6 +112,7 @@ const Item = (props: ItemProps) => {
           'kirk-item--highlighted': highlighted,
           'kirk-item--clickable': isClickable,
           'kirk-item--hideHoverBackground': hideHoverBackground,
+          'kirk-item--disabled': disabled,
```
gabrielagabriel, line 115: following my suggestion: non need to apply this class here

src/_utils/item/Item.tsx
```diff
@@ -134,7 +137,7 @@ const Item = (props: ItemProps) => {
           <Text
             className="kirk-item-body"
             display={leftBodyDisplay}
-            textColor={leftBodyColor}
+            textColor={disabled ? color.fadedText : leftBodyColor}
```
gabrielagabriel, line 140: You can override `Text` to avoid repetion
```suggestion
const StyledText = styled(Text)`
   color: ${props => (props.isDisabled ? color.disabled : '')};
`
```

src/_utils/item/index.tsx
```diff
@@ -23,6 +24,16 @@ const StyledItem = styled(Item)`
     background: none;
   }
 
+  &:disabled,
+  &.kirk-item--disabled {
+    cursor: default;
+    background-color: ${color.defaultBackground};
+
+    .kirk-text {
```
gabrielagabriel, line 32: cf: kirk-text override with **Styled-Components**

src/_utils/radioIcon/index.tsx
```diff
@@ -7,16 +7,27 @@ import CircleIcon from 'icon/circleIcon'
...
+}: RadioIconProps) => {
   if (isLoading) {
     return <Loader size={24} inline />
   }
   if (isChecked) {
-    return <CircleIcon iconColor={color.primary} innerDisc />
+    return (
+      <CircleIcon
+        iconColor={isDisabled ? color.fadedText : color.primary}
+        isDisabled={isDisabled}
```
gabrielagabriel, line 25: only this is needed

src/icon/checkIcon.tsx
```diff
@@ -24,23 +27,29 @@ const CheckIcon = (props: CheckProps) => (
...
 export const StyledCheckIcon = styled(CheckIcon)`
+  /* .kirk-icon--disabled forces a stroke color, but it needs to be overriden if a specific background is passed */
+  &.kirk-icon--disabled path {
+    stroke: ${props =>
+      props.backgroundColor !== defaultBackgroundColor ? color.white : color.fadedText};
+  }
   & {
-    background-color: ${props => props.backgroundColor};
+    background-color: ${props =>
+      props.isDisabled && props.backgroundColor !== defaultBackgroundColor
```
gabrielagabriel, line 50: cool ! you are already using this model here! why not use everywhere else?

Tuxosaurus: The checkIcon is already an exception, the only one with a "backgroundColor" prop. It was a mess to change correctly so I'm not sure this is ideal.

I'm even thinking we could split this Icon into "just a check icon" and "circled check icon".

gabrielagabriel: Typography and TheVoice uses also **styled-components**. Since styled-components is our official stack, we should start writting with its syntax. 

It will keep your code DRY and easy to maintain

src/_utils/icon/BaseIcon.tsx
```diff
@@ -18,6 +18,7 @@ export const BaseIconDefaultProps = {
   title: '',
   badgeAriaLabel: '',
   badgeContent: '',
+  isDisabled: false,
```
simonrelet, line 21: Why not disabled?

src/icon/busIcon.tsx
```diff
@@ -15,7 +15,7 @@ export const BusIcon = (props: Icon) => (
       <path
         d="M.67 5h20c1 0 1.66 1.66 2.66 1.66h4v4.67a1 1 0 0 1-1 1H4.67L1 11.1a.5.5 0 0 1-.34-.47V5z"
         stroke={props.iconColor}
-        fill={props.iconColor}
+        fill={props.isDisabled ? color.fadedText : props.iconColor}
```
simonrelet, line 18: Is there a reason not to use [`currentColor`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/color) in SVG and simply use CSS to change it?

grosbouddha:
I can barely see the content of the disabled widget and I think it won't pass even generous contrast ratio tests.
@nickoola any guidelines on minimum contrast ratio ?

src/_utils/icon/index.tsx
```diff
@@ -14,6 +14,11 @@ const StyledBaseIcon = styled(BaseIcon)`
     box-shadow: ${shadow.icon};
   }
 
+  &.kirk-icon--disabled,
+  &.kirk-icon--disabled * {
```
grosbouddha, line 18: Can we not use the star selector ?

src/itemChoice/story.tsx
```diff
@@ -26,14 +28,14 @@ stories.add(
           labelInfo={text('Title Info', 'Additional informations')}
           data={text('Data', 'Data')}
           dataInfo={text('Data Info', 'Info')}
-          leftAddon={withLeftIcon ? <ComfortIcon /> : null}
-          rightAddon={withRightIcon ? <ComfortIcon /> : null}
+          leftAddon={withLeftIcon ? <CheckIcon isDisabled={disabled} /> : null}
```
grosbouddha, line 31: Too bad we have to pass the disabled state to the left/right addons :/
Not sure how to fix it easily though.

src/_utils/icon/BaseIcon.tsx
wakooka: Not sure, it will make it consistent with html but inconnsistent with the component API (isLoading, isChecked).
I would keep it as it is

Tuxosaurus: Yes there's both `disabled` and `isDisabled` in some of the files I touched.
Personally, I prefer the explicit boolean wording `isDisabled`.

src/itemChoice/story.tsx
Tuxosaurus: Yes, I thought of checking if the ReactNode has a "disabled" prop and if so change it in the Item component but even so I'd rather try it in a separate branch

Tuxosaurus:
> I can barely see the content of the disabled widget and I think it won't pass even generous contrast ratio tests.
> @nickoola any guidelines on minimum contrast ratio ?

Agreed. I've raised the point in the Product discussion.

src/_utils/icon/BaseIcon.tsx
simonrelet: `isDisabled` is indeed a nicer API but it is not the one used by HTML and the community, which means that we are breaking the _implicit_ API for disabling things. If we need to pass it to some external component, or someone external wants to use it, it should be wrapped to rename the prop 😟.

src/icon/busIcon.tsx
Tuxosaurus: I guess none of us were familiar with svgs when we started this, so it's almost never used.
I'll do some cleaning and try to implement it in this icon but for the rest I'm guessing it will need a separate refacto, possibly with improvements to BaseIcon

src/_utils/icon/index.tsx
Tuxosaurus: Done

src/_utils/item/Item.tsx
Tuxosaurus: Overriding `Text` this way did not work since I do need to take pass a variable color in case it's not disabled.

But still, I factorized the change in a separate `getTextColor` method to avoid repetition

src/_utils/radioIcon/index.tsx
Tuxosaurus: Done

src/icon/checkIcon.tsx
Tuxosaurus: Done

src/_utils/icon/BaseIcon.tsx
Tuxosaurus: I'm keeping it for now but you're right, the API should be streamlined in a separate PR.

src/_utils/icon/index.tsx
Tuxosaurus: Done

src/_utils/item/index.tsx
Tuxosaurus: Done

src/_utils/item/Item.tsx
Tuxosaurus: Done

PR #697, opened 2020-04-29
Fix missing CSS prefix for DatePicker and ButtonGroup

simonrelet:
## Description

The base class name of `DatePicker` and `ButtonGroup` are missing the "kirk" prefix.

## What has been done

Add the missing prefix to the component's base class names.

## Things to consider

🤷‍♂️

## How it was tested

I added a test for that in each component.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
-[...]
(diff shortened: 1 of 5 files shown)


wakooka:
Update changelog

PR #699, opened 2020-04-30
add aloneInTheBack icon, implemented in TripCard

Tuxosaurus:
Add a new icon for the Alone in the Back feature. It will also be used within the TripCard component along with ComfortIcon.

Screen Shot 2020-04-30 at 09 01 41
Screen Shot 2020-04-30 at 09 00 01

Test locally on Firefox and Chrome


src/icon/aloneInTheBackIcon.tsx
```diff
@@ -0,0 +1,18 @@
+// tslint:disable:max-line-length
+import React from 'react'
+import BaseIcon from '_utils/icon'
+import { BaseIconDefaultProps } from '_utils/icon/BaseIcon'
+
+export const AloneInTheBackIcon = (props: Icon) => (
+  <BaseIcon {...props}>
```
wakooka, line 7: Should we make `statusIcon` for more flexibility so we can have the red strikethrough?

src/tripCard/TripCard.unit.tsx
```diff
@@ -84,6 +86,7 @@ describe('TripCard component', () => {
   it('Should have only the Ladies Only icon', () => {
     const tripCard = shallow(<TripCard {...mockedProps} flags={{ ladiesOnly: true }} />)
     expect(tripCard.find(LadyIcon).exists()).toBe(true)
+    expect(tripCard.find(AloneInTheBackIcon).exists()).toBe(false)
     expect(tripCard.find(ComfortIcon).exists()).toBe(false)
     expect(tripCard.find(LightningIcon).exists()).toBe(false)
   })
```
wakooka, line 92: nit: add a test "Should have the aloneInTheBackIcon"

src/tripCard/story.tsx
```diff
@@ -35,9 +35,16 @@ const tripCardConfig = () => ({
...
+    aloneInTheBack: boolean('Alone in the back', true),
     maxTwo: boolean('Max 2 in the back', true),
     autoApproval: boolean('Auto approval', true),
   },
+  titles: {
+    ladiesOnly: text('Flag title for Ladies only', 'Ladies Only'),
+    aloneInTheBack: text('Flag title for Alone in the back', 'Alone in the back'),
+    maxTwo: text('Flag title for Max 2 in the back', 'Max 2 in the back'),
+    autoApproval: text('Flag title for Auto approval', 'Auto approval'),
+  },
```
wakooka, line 47: 👍

src/icon/aloneInTheBackIcon.tsx
Tuxosaurus: I guess it's not required since I copied ComfortIcon since it's supposed to be used in the same context.

src/tripCard/TripCard.unit.tsx
Tuxosaurus: There's no such test for Comfort or Lightning icons, only for Lady, which also tests the absence of the others. As they're just booleans, I found it "good enough".

PR #701, opened 2020-04-30
[BOYSCOUT] Update spacing on TheVoice

gabrielagabriel:
## Description

TheVoice spacing not follow spec

Screenshot 2020-04-30 at 11 07 56

## What has been done

- Update spacings

## Things to consider

- Converting **margin** to **padding** to avoid collapse
- Rewrite style on **mobile-first**
- Reduce **styled-components** specificity
- Remove `isInverted` class

## How it was tested

- _Manually_ 🤙

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
-[...]
+- **[FIX]** `TheVoice` spacing.
+  [...]
 
 # v31.0.3 (04/05/2020)
 
--- a/src/theVoice/index.tsx
+++ b/src/theVoice/index.tsx
@@ -5,7 +5,7 @@ import TheVoice from './TheVoice'
 
 const StyledTheVoice = styled(TheVoice)`
   margin: 0;
-  padding: ${space.xl} 0 calc(${space.xl} + ${space.m}); /* 32px it's not en scale but on spec :() */
+  padding: ${space.xl} 0 calc(${space.xl} + ${space.m}); /* 32px it's not on scale but on spec :() */
   color: ${props => (props.isInverted ? color.white : '')};
 
   @media (${responsiveBreakpoints.isMediaLarge}) {
--- a/src/theVoice/index.tsx
+++ b/src/theVoice/index.tsx
@@ -14,6 +14,7 @@ const StyledTheVoice = styled(TheVoice)`
   margin: 0;
   padding: ${space.xl} 0 var(--space-bottom);
   color: ${props => (props.isInverted ? color.white : '')};
+  white-space: pre-line;
 
   @media (${responsiveBreakpoints.isMediaLarge}) {
     text-align: center;
--- a/src/theVoice/index.tsx
+++ b/src/theVoice/index.tsx
@@ -1,23 +1,15 @@
 import styled from 'styled-components'
-
-import { color, font, responsiveBreakpoints, space } from '_utils/branding'
+import { space, color, responsiveBreakpoints } from '_utils/branding'
 
 import TheVoice from './TheVoice'
 
 const StyledTheVoice = styled(TheVoice)`
-  & {
+  margin: 0;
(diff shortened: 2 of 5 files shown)


src/theVoice/index.tsx
```diff
@@ -5,7 +5,7 @@ import TheVoice from './TheVoice'
 
 const StyledTheVoice = styled(TheVoice)`
   margin: 0;
-  padding: ${space.xl} 0 calc(${space.xl} + ${space.m}); /* 32px it's not en scale but on spec :() */
+  padding: ${space.xl} 0 calc(${space.xl} + ${space.m}); /* 32px it's not on scale but on spec :() */
```
grosbouddha, line 8: I would just use 32 px here or create a new constant.
You are doing arithmetic on things that are not semantically related: E.g. if someone is modifying the space.m constant one day, I'm not sure he would expect to modify the voice bottom padding at the same time.

I don't feel too strongly about and can keep that too since probably nobody will ever change these values globally. It just look like you have relationships between constants but actually we have none. The only relationship we have between these values is that they are a multiple of a base grid unit size (e.g. m => 1 grid unit (8px), xl => (2 gu -> 24px), etc).

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **[BREAKING CHANGE]** `TheVoice` spacing.
```
grosbouddha, line 3: You will have a lot of visual snapshots to update, good luck :)

src/theVoice/index.tsx
```diff
@@ -1,22 +1,15 @@
...
 import TheVoice from './TheVoice'
 
 const StyledTheVoice = styled(TheVoice)`
-  & {
-    text-align: center;
-    font-size: ${font.xl.size};
-    margin: ${space.xxl} 0;
-  }
+  margin: 0;
+  padding: ${space.xl} 0 calc(${space.xl} + ${space.m}); /* 32px it's not on scale but on spec :() */
```
wakooka, line 8: I'd rather add `32px` to our scale than doing some calculation in CSS.
If you don't want to make a too big change in the codebase by doing that, you can add a scoped variable in theVoice. e.g. `const paddingBottom = '32px'`

CHANGELOG.md
wakooka, line 3: No breaking change on how we use the component then this is a `FIX`

src/theVoice/index.tsx
```diff
@@ -1,22 +1,15 @@
 import styled from 'styled-components'
-import { space, color, font, responsiveBreakpoints } from '_utils/branding'
+import { space, color, responsiveBreakpoints } from '_utils/branding'
 
 import TheVoice from './TheVoice'
 
 const StyledTheVoice = styled(TheVoice)`
-  & {
```
simonrelet, line 7: 👏

src/theVoice/index.tsx
```diff
@@ -1,22 +1,15 @@
...
-    text-align: center;
-    font-size: ${font.xl.size};
```
wakooka, line 9: Actually just noticed that, it's in the spec, on desktop the font size should be bigger. Can you add it to the mediaLarge breakpoint?

gabrielagabriel: @wakooka  the ui-library don't have the size defined on spec on the Font SIze scale :S
should we add it anyway?

src/theVoice/index.tsx
```diff
@@ -1,23 +1,24 @@
...
-    margin: ${space.xxl} 0;
-  }
+  /* @TODO: Align scale and spec
+  /* 32px it's not part of current scale. 4 it could be calculated by the base unit space.s  */
+  --space-bottom: 32px;
+  /* 40px is an exception on font scale  */
+  --font-size-desktop: 2.5rem;
```
wakooka, line 12: Can we use px so we don't introduce a new unit in this PR?

PR #702, opened 2020-04-30
[HOTFIX] ScreenModal layout to  all browsers

gabrielagabriel:
## Description

**BEFORE**
Screenshot 2020-04-30 at 13 31 33

**AFTER**
Screenshot 2020-04-30 at 13 34 00

## What has been done

- Convert `vh` into `%`

## Things to consider

- Magic numbers from spec

## How it was tested

- Manually 😬

### DIFF
--- a/src/successModal/elements.tsx
+++ b/src/successModal/elements.tsx
@@ -23,10 +23,9 @@ const StyledSuccessModal = styled(Modal)`
 
   .kirk-modal-body {
     display: flex;
+    min-height: -webkit-fill-available;
     min-height: 100vh;
     /* to avoid scroll on webkit browsers when they have a top bar */
-    min-height: -moz-available;
-    min-height: -webkit-fill-available;
     flex-direction: column;
 
     @media (${responsiveBreakpoints.isMediaLarge}) {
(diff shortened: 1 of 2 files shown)


src/successModal/elements.tsx
```diff
@@ -23,10 +23,9 @@ const StyledSuccessModal = styled(Modal)`
 
   .kirk-modal-body {
     display: flex;
+    min-height: -webkit-fill-available;
```
wakooka: Removing this to make it consistent with other modal and using min-height: 100%.
Tested in Safari, Firefox, Chrome desktop. And Safari & Chrome mobile, I didn't see any issue even with the address bar, I'm guessing since we're in % it's working, it wouldn't if we were using vh.

src/successModal/elements.tsx
```diff
@@ -55,25 +54,33 @@ const Media = styled.div`
...
-  height: 33.33vh;
-  min-height: 33.33vh;
+  height: 33.33%;
+  max-height: 33.33%;
+
+  img {
+    height: 100%;
+    max-width: 305px; /* Magic Numbers from spec */
+    max-height: 240px; /* Magic Numbers from spec */
+  }
```
wakooka, line 65: Let's see if we have a better way to handle images next week.

PR #704, opened 2020-04-30
[BBC-7850] Add import-order eslint rule

mcellier:
## Description

- Add import order check in eslint (same as kairos)
- Add lint-staged in pre-commit hook to run `prettier` and `eslint --fix` automatically on changed files.

## What has been done

- Add `eslint-plugin-simple-import-sort` and `lint-staged` packages
- Add autofix (prettier, eslint --fix) as pre-commit hook
- Fix codebase

## Things to consider

- Same as kairos: imports order impacts `styled-components` output. That's why lots of snapshot have been updated. We might have to update snapshot on kairos after update too.
- I manually configured the sorting, as the plugin can't differenciate external plugins and our files as we're using a `baseUrl` (under `src`). 

## How it was tested

With storybook.


simonrelet:
☹️
Screenshot 2020-04-30 at 18 21 24

PR #707, opened 2020-05-05
[BBC-6972] Responsive `ConfirmationModal`

gabrielagabriel:
## Description

https://www.loom.com/share/71870630fa714fc4ae9cc00d7a6518ed

## What has been done

- Fix UI issues on ConfirmationModal due to `size` property

## Things to consider

- Remove `size` property and use media query to set the layout

## How it was tested

- Manually 👋. on Chrome, Firefox, Safari (iOS)


src/confirmationModal/index.tsx
```diff
@@ -76,19 +51,44 @@ const StyledConfirmationModal = styled(ConfirmationModal)`
...
+      margin: 0 auto;
+      background: none;
+    }
+
+    .kirk-confirmationModal-footer .kirk-button {
+      margin: ${space.l} ${space.xl} 0;
+    }
+
+    .kirk-confirmationModal-confirmButton {
+      min-width: 128px;
```
wakooka, line 87: Magic value?

gabrielagabriel: 🙅‍♀️ From the spec

PR #708, opened 2020-05-06
[BBC-7928] Allow icon in ItemCheckbox

mcellier:
## Description

- Add the `leftAddon` option to `ItemCheckbox` to allow having an icon.
- Add edit your publication options screen in storybook

Capture d’écran 2020-05-06 à 10 40 01

## Things to consider

They're is still hard-coded styles in the story. I'll need to update this in other PRs

## How it was tested

Storybook, mobile & desktop layout, chrome.

### DIFF
--- a/src/itemCheckbox/ItemCheckbox.tsx
+++ b/src/itemCheckbox/ItemCheckbox.tsx
@@ -16,6 +16,7 @@ export interface ItemCheckboxProps extends A11yProps {
   readonly name: string
   readonly data?: string
   readonly className?: string
+  readonly leftAddon?: React.ReactNode
   readonly labelTitle?: string
   readonly dataInfo?: string
   readonly checked?: boolean
@@ -43,6 +44,7 @@ class ItemCheckbox extends Component<ItemCheckboxProps> {
       data,
       className,
       labelTitle,
+      leftAddon,
       dataInfo,
       checked,
       disabled,
@@ -68,6 +70,7 @@ class ItemCheckbox extends Component<ItemCheckboxProps> {
         className={cc(['kirk-item-checkbox', className])}
         leftTitle={labelTitle}
         leftBody={label}
+        leftAddon={leftAddon}
         rightTitle={data}
         rightTitleDisplay={TextDisplayType.SUBHEADERSTRONG}
         rightBody={dataInfo}
--- a/src/pages/yourrides/offer/editOptions.story.tsx
+++ b/src/pages/yourrides/offer/editOptions.story.tsx
@@ -1,0 +1,122 @@
+/* eslint-disable react/jsx-curly-newline */
+import React, { Component, Fragment } from 'react'
+import { storiesOf } from '@storybook/react'
+
+import Button, { ButtonStatus } from 'button'
+import Divider from 'divider'
+import ComfortIcon from 'icon/comfortIcon'
+import LadyIcon from 'icon/ladyIcon'
+import LightningIcon from 'icon/lightningIcon'
+import ItemCheckbox from 'itemCheckbox'
+import ItemInfo from 'itemInfo'
+import { BottomContent, Content, MainContent } from 'layout/content'
+import LayoutNormalizer from 'layout/layoutNormalizer'
+import Section from 'layout/section/baseSection'
+import Stepper, { StepperDisplay } from 'stepper'
+import SubHeader from 'subHeader'
+import Textarea from 'textarea'
+import TextTitle from 'typography/title'
+
+const stories = storiesOf('Pages|Your rides/Offer/Edit/Options', module)
+
+type EditOptionsState = Readonly<{
+  isComfort: boolean
+  isLadyOnly: boolean
+  isInstantBooking: boolean
+  buttonStatus: ButtonStatus
+}>
+
+// Note: some hard-coded styles will be removed with the following tickets: BBC-7943, BBC-7944
+
+class EditOptions extends Component<{}, EditOptionsState> {
+  state: EditOptionsState = {
+    isComfort: false,
+    isLadyOnly: false,
+    isInstantBooking: false,
+    buttonStatus: ButtonStatus.PRIMARY,
+  }
+
+  save = (): void => {
+    this.setState({ buttonStatus: ButtonStatus.LOADING })
+
+    setTimeout(() => this.setState({ buttonStatus: ButtonStatus.CHECKED }), 1000)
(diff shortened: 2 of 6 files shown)


src/itemCheckbox/ItemCheckbox.tsx
```diff
@@ -16,6 +16,7 @@ export interface ItemCheckboxProps extends A11yProps {
   readonly name: string
   readonly data?: string
   readonly className?: string
+  readonly leftAddon?: React.ReactNode
```
mcellier, line 19: Should i rename this `icon` to be more explicit? For now i just went with the same name as the base item, but as it's internal, we might as well not stay strict to its props. WDYT?

src/pages/yourrides/offer/editOptions.story.tsx
```diff
@@ -0,0 +1,103 @@
...
+    this.setState((prevState) => ({ [name]: !prevState[name] }))
+  }
+
+  render = (): JSX.Element => (
+    <MainContent>
+      <Content>
+        <Section>
+          <SubHeader>Passengers options</SubHeader>
+
+          <div style={{ display: 'flex' }}>
```
mcellier, line 52: @grosbouddha Do the horizontal behavior works?

src/pages/yourrides/offer/editOptions.story.tsx
```diff
@@ -0,0 +1,103 @@
...
+            <TextTitle style={{ flex: 1, lineHeight: '48px' }}>Number of passengers</TextTitle>
+            <div style={{ position: 'relative', right: '-14px' }}>
```
mcellier, line 54: `right: '-14px'` is needed for the + icon of the `Stepper` to be aligned with the `ItemCheckbox` circles... I don't know how to improve this :(

src/pages/yourrides/offer/editOptions.story.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+          <Divider />
+          <ItemInfo mainTitle="Additional details" />
+          <Textarea
+            placeholder="Flexible about where and when to meet? Not taking the motorway? Got limited space in your boot? Keep passengers in the loop."
+            onChange={() => {}}
+          />
+        </Section>
+      </Content>
+      <BottomContent style={{ display: 'flex', justifyContent: 'center', padding: '16px' }}>
```
mcellier, line 94: Do we have a "just center this button" behavior somewhere?
Also, do the new layout system adds de padding by itself on the button?

mcellier: @PerrineBoissieres Do we always display a centered button with 16px padding in the `BottomSection`? If so, we can add them to `BottomContent`...

src/itemCheckbox/ItemCheckbox.tsx
simonrelet: A name I often see for this kind of thing is "adornment"

src/pages/yourrides/offer/editOptions.story.tsx
```diff
@@ -0,0 +1,103 @@
...
+          />
+          <ItemCheckbox
+            onChange={() => this.updateOption('isInstantBooking')}
+            leftAddon={<LightningIcon />}
+            labelTitle="Instant booking"
+            checked={this.state.isInstantBooking}
+          />
+
+          <Divider />
+          <ItemInfo mainTitle="Additional details" />
```
mcellier, line 87: Use something else than ItemInfo + should be label for Textarea

src/pages/yourrides/offer/editOptions.story.tsx
mcellier: As talked "IRL", I'm thinking of adding a `leftAddon` prop to the `Stepper` component. It might handle these behaviors: 
- Subheader in this page 
- Itinerary in the price edit page

src/pages/yourrides/offer/editOptions.story.tsx
grosbouddha: We don't have the "just center this button" in kirk yet.
We have the SubmitButton in kairos but it should be moved (ideally) to kirk.

src/pages/yourrides/offer/editOptions.story.tsx
grosbouddha: As discussed, I would either create a new ItemStepper widget or add new props to the exisiting stepper.
As long as, we don't do CSS in kairos, I'm ok with both approaches.

grosbouddha: As discussed, I would either create a new ItemStepper widget which will be an Item that contains a Stepper. The Item knows how to manage its horizontal padding automatically.

src/itemCheckbox/ItemCheckbox.tsx
grosbouddha: I would keep a name that match the actual value that can be passed. So leftAddon will be good here IMO (unless you find a way to type the param value to be only an icon).

src/pages/yourrides/offer/editOptions.story.tsx
grosbouddha, line 52: Any CSS in this story will end up as CSS in kairos eventually. Let's make sure we have some kirk way to handle all these custom style/CSS bits.

src/pages/yourrides/offer/editOptions.story.tsx
PerrineBoissieres: BottomContent as well as Content should not add any paddings, that's not their role.
Button can be used in many other components, so as well it should have its own margins or else it will mess up with a lot of cases.
Ideal solution would be to move the SubmitButton component to Kirk, yep, so it handles proper spacing and position.

src/pages/yourrides/offer/editOptions.story.tsx
mcellier: BBC-7943

src/pages/yourrides/offer/editOptions.story.tsx
mcellier: I'll create a ticket for this

mcellier: BBC-7944

src/pages/yourrides/offer/editOptions.story.tsx
mcellier: As discussed: let's keep the ItemInfo for now, but added a11y

PR #710, opened 2020-05-06
Add leftAddon option to Stepper

mcellier:
## Description

Allows rendering a vertically-centered component on the left of small steppers.
- Updated example for Edit your trip options
- Add example for Edit your trip price (also used during publication)

Capture d’écran 2020-05-06 à 17 56 19
Capture d’écran 2020-05-06 à 17 56 09

## What has been done

- Add `leftAddon` option to `Stepper`
- Updated unit tests
- Updated related pages
- Removed custom CSS!

## How it was tested

Storybook with small/large layouts, chrome + unit tests

### DIFF
--- a/src/stepper/Stepper.tsx
+++ b/src/stepper/Stepper.tsx
@@ -1,6 +1,7 @@
 import React, { PureComponent, RefObject } from 'react'
 import cc from 'classcat'
 import { canUseDOM } from 'exenv'
+import isEmpty from 'lodash.isempty'
 
 import { isTouchEventsAvailable } from '_utils'
 import { color, delay, font, pxToInteger, space } from '_utils/branding'
@@ -38,6 +39,7 @@ export interface StepperProps {
   onChange?: (obj: OnChangeParameters) => void
   display?: StepperDisplay
   focus?: boolean
+  leftAddon?: React.ReactNode
 }
 
 interface StepperState {
@@ -174,9 +176,10 @@ export default class Stepper extends PureComponent<StepperProps, StepperState> {
         }
   }
 
-  render() {
+  render(): JSX.Element {
     const {
       className,
+      leftAddon,
       title,
       increaseLabel,
       decreaseLabel,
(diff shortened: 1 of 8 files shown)


mcellier:
> I have to main comments:
> 
> * We need to add some top padding: For the itinerary cases, the top padding is too small and users might be tempted to add custom CSS or useless spacer to make it look good. I guess the top padding is inside the Stepper, not on the top container ?
> * I understand that you need to support this to move forward and not be blocked on UX. But in mid term  I think, we need to turn this frankenstein thing into a new official widget; combining complex widgets using leftAddon is mostly a hack and it will make out design system unmaintanable if we spread the pattern. Let's submit that to unblock you but let's confront the UX team to support this officially with some specs. Maybe we can use the Chapter Pixar for that ? @wakooka

* I'll check if I can find something for the top padding. I'll make another PR if this one is already merge, as they're still work to do on theses pages
* Actually, we have, at least partially, specs for Stepper with itinerary on the left side:

Capture d’écran 2020-05-07 à 09 47 43

src/stepper/Stepper.tsx
```diff
@@ -174,9 +176,8 @@ export default class Stepper extends PureComponent<StepperProps, StepperState> {
         }
   }
 
-  render() {
+  renderStepper = (): JSX.Element => {
```
wakooka, line 179: Why did you create this method?

wakooka:
> we need to turn this frankenstein thing into a new official widget.
I agree, let's discuss this next Tuesday with UX designers.

src/stepper/Stepper.unit.tsx
```diff
@@ -37,7 +38,7 @@ it('Should have the default text & attributes', () => {
     />,
   )
 
-  expect(stepper.find('.kirk-stepper').prop('aria-label')).toEqual(
+  expect(stepper.find('.kirk-stepper-content').prop('aria-label')).toEqual(
```
gabrielagabriel, line 41: please, target components instead of classes

wakooka: There was no decision made on this topic yet, Simon has started a documents on how we write tests, but this still in progress. So for now targeting classes is fine.

src/stepper/Stepper.tsx
mcellier: A remnant from a precedent approach. I'll remove this

src/stepper/Stepper.unit.tsx
mcellier: No component here to target, and I won't rewrite all of it right now ;)

src/stepper/Stepper.tsx
mcellier: done

mcellier:
@wakooka @grosbouddha I added a vertical padding (8px) to small steppers. Didn't want to break things on the large ones. I cheched with Pantea who told me that she found 12px padding on these. @wakooka can you check that also on the next pixar meeting ?

What it looks like now:

Capture d’écran 2020-05-07 à 11 14 05
Capture d’écran 2020-05-07 à 11 13 51

PS: the itinerary adds a padding on the bottom too...

mcellier:
> @wakooka @grosbouddha I added a vertical padding (8px) to small steppers. Didn't want to break things on the large ones. I cheched with Pantea who told me that she found 12px padding on these. @wakooka can you check that also on the next pixar meeting ?
> 
> PS: the itinerary adds a padding on the bottom too...

As seen with @wakooka, i added a local const to handle the 12px padding, as it seems we took a shortcut and didn't handled it. 

Moreover, i removed the padding bottom on itinerary's arrival if it's inside a stepper component to fix the component. Now it looks nice:

Capture d’écran 2020-05-07 à 11 50 14

wakooka:
> As seen with @wakooka, i added a local const to handle the 12px padding, as it seems we took a shortcut and didn't handled it.

I'm note sure why we never had the 12px value since this is the actual design system spec.

PR #712, opened 2020-05-07
Fix mainInfo prop type in ItemData (from string to React.ReactNode)

mcellier:
## Description

Update `mainInfo` type from `string` to `React.ReactNode`

### DIFF
--- a/src/itemData/index.tsx
+++ b/src/itemData/index.tsx
@@ -9,7 +9,7 @@ export interface ItemDataProps extends A11yProps {
   readonly data: string | JSX.Element
   readonly dataStrikeThrough?: boolean
   readonly dataAriaLabel?: string
-  readonly mainInfo: string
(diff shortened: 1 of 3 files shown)


src/itemData/index.tsx
```diff
@@ -9,7 +9,7 @@ export interface ItemDataProps extends A11yProps {
   readonly data: string | JSX.Element
   readonly dataStrikeThrough?: boolean
   readonly dataAriaLabel?: string
-  readonly mainInfo: string
+  readonly mainInfo: React.ReactNode
   readonly className?: string
   readonly mainTitle?: string
```
pith, line 14: Should we do the same for `mainTitle`?

mcellier: If we do so we should update the base item also `readonly leftTitle?: string`

PR #716, opened 2020-05-07
Added animation in search form large devices

wakooka:
## Description

Adding animation for search form on large layout.
Kapture 2020-05-07 at 16 29 33

## What has been done

Factorized repeated logic in an `Overlay` component.
Added animation using `CSSTransition` component. 
Added media query on `prefers-reduced-motion` to disable transition

## How it was tested

Tested locally in storybook, updated unit test, and change MacOS settings to test `prefers-reduced-motion`

### DIFF
--- a/src/_utils/useReducedMotion/index.tsx
+++ b/src/_utils/useReducedMotion/index.tsx
@@ -1,0 +1,19 @@
+import { useEffect, useState } from 'react'
+
+/**
+ * This custom hook will return reduced motion boolean
+ * To be used to disable animation based on user preferences
+ * https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion
+ *
+ */
+export const useReducedMotion = () => {
+  const [reducedMotion, setReducedMotion] = useState(false)
+
+  useEffect(() => {
+    setReducedMotion(window.matchMedia('(prefers-reduced-motion)').matches)
+  }, [])
+
+  return reducedMotion
+}
+
+export default useReducedMotion
--- a/src/searchForm/overlay/Overlay.tsx
+++ b/src/searchForm/overlay/Overlay.tsx
@@ -1,0 +1,50 @@
+import React, { useRef } from 'react'
+import CSSTransition from 'react-transition-group/CSSTransition'
+
+import useReducedMotion from '_utils/useReducedMotion'
+
+export const enterTransitionDuration = 350
+export const exitTransitionDuration = 500
+export const reducedMotionTransitionDuration = 0
+export const TRANSITION_CLASS_NAME = 'kirk-searchForm-overlay-transition'
+
+const overlayTimeout = {
+  enter: enterTransitionDuration,
+  exit: exitTransitionDuration,
+}
+
+type OverlayProps = {
+  closeOnBlur: () => void
+  shouldDisplay: boolean
+  children: React.ReactNode
+  className: string
+}
+
+export const Overlay = ({ closeOnBlur, shouldDisplay, children, className }: OverlayProps) => {
+  const overlayRef = useRef<HTMLDivElement>(null)
+  const reducedMotion = useReducedMotion()
+
+  return (
+    <div
+      ref={overlayRef}
+      onBlur={evt => {
+        if (!overlayRef.current.contains(evt.relatedTarget as Node)) {
+          closeOnBlur()
+        }
+      }}
+      className={className}
+    >
+      <CSSTransition
+        in={shouldDisplay}
+        classNames={TRANSITION_CLASS_NAME}
+        timeout={reducedMotion ? reducedMotionTransitionDuration : overlayTimeout}
+        mountOnEnter
+        unmountOnExit
+      >
+        {children}
+      </CSSTransition>
+    </div>
+  )
+}
+
+export default Overlay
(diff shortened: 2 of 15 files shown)


src/searchForm/overlay/Overlay.tsx
```diff
@@ -0,0 +1,51 @@
...
+  children: React.ReactNode
+  className: string
+}
+
+export const Overlay = ({ closeOnBlur, shouldDisplay, children, className }: OverlayProps) => {
+  const overlayRef = useRef<HTMLDivElement>(null)
+  const [reducedMotion, setReducedMotion] = useState(false)
+
+  useEffect(() => {
+    setReducedMotion(window.matchMedia('(prefers-reduced-motion)').matches)
```
simonrelet, line 25: Maybe have a custom hook for this? I think animations are probably used elsewhere.

src/searchForm/overlay/Overlay.tsx
```diff
@@ -0,0 +1,51 @@
...
+      onBlur={evt => {
+        if (!overlayRef.current.contains(evt.relatedTarget as Node)) {
+          closeOnBlur()
+        }
+      }}
+      className={className}
+    >
+      <CSSTransition
+        in={shouldDisplay}
+        classNames="kirk-searchForm-overlay-transition"
```
simonrelet, line 40: Is there a reason to use constants for transition duration and not the class names?

src/searchForm/overlay/Overlay.tsx
wakooka: Done

src/searchForm/overlay/Overlay.tsx
wakooka: Done 👍

src/searchForm/overlay/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+const StyledOverlay = styled(Overlay)`
+  & .${TRANSITION_CLASS_NAME}-enter {
+    opacity: 0;
+    width: 0;
+  }
+
+  & .${TRANSITION_CLASS_NAME}-enter-active {
+    opacity: 1;
+    width: ${componentSizes.searchOverlayWidth};
+    transition: width ${enterTransitionDuration}ms ${transitionTimingFunction},
```
grosbouddha, line 23: does styled-component produce vendor specific transition rules ?

src/_utils/useReducedMotion/index.tsx
```diff
@@ -0,0 +1,19 @@
+import { useEffect, useState } from 'react'
+
+/**
+ * This custom hook will return reduced motion boolean
+ * To be used to disable animation based on user preferences
+ * https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion
+ *
+ */
+export const useReducedMotion = () => {
```
grosbouddha, line 9: I'm still confused by these global helper hooks with useEffect.
I would prefer something like a global util:
const USE_REDUCED_MOTION = window.matchMedia('(prefers-reduced-motion)').matches
(we can even cache the result if really needed)

This is simpler to read and does not depend on React lifecycle/hooks (it does not use useEffect).

To me it's not very valuable to have this as a global hook and even misleading: It should be called useReducedMotionPostRender to match what it's actually doing.
Am I missing some use cases ?

src/searchForm/overlay/index.tsx
wakooka: I don't think we need any
https://caniuse.com/#feat=css-transitions

src/_utils/useReducedMotion/index.tsx
wakooka: > const USE_REDUCED_MOTION = window.matchMedia('(prefers-reduced-motion)').matches

wakooka:
> The animation looks good for the 2 first fields (it's growing from the field) but not really good for the two last fields (it jumps to the left then grow to the right).

Good point, actually this was much better in the specs, I'll try to make the field is transitioning to its transition.

src/searchForm/overlay/index.tsx
wakooka: I checked it's oupouting the code with prefixes.

src/_utils/useReducedMotion/index.tsx
grosbouddha: Ok let's rename the hook to have a correct name using some suffix (postRender, onMount, etc)

simonrelet: Maybe:

```ts
const hasReducedMotion = () => {
  if (window == null) {
    return false
  }

  return window.matchMedia('(prefers-reduced-motion)').matches
}

export const useReducedMotion = () => {
  // During server side the value is `false`, during client side rendering it is initialised with
  // the correct value.
  const [reducedMotion, setReducedMotion] = useState(hasReducedMotion)

  useEffect(() => {
    // If the component was serve-sider rendered we update the value.
    // Otherwise the state won't change
    setReducedMotion(hasReducedMotion())
  }, [])

  return reducedMotion
}
```

wakooka: > To me it's not very valuable to have this as a global hook and even misleading: It should be called useReducedMotionPostRender to match what it's actually doing.

I don't see the point of renaming or even changing the code. As a hook consumer you want a value from it and use it in your component code, for me what the hook is doing and how is not important IMHO. 

@grosbouddha why do you need to know it will be run post render?

PR #717, opened 2020-05-13
Use relative imports

simonrelet:
## Description

After more research, import paths transform during the build [go against the Design goals of TypeScript](https://github.com/Microsoft/TypeScript/issues/15479#issuecomment-543329547).

[Some](https://www.npmjs.com/package/tspath) [custom](https://gist.github.com/Kitanotori/86c906b2c4032d4426b750d232a4133b) [tools](https://gist.github.com/azarus/f369ee2ab0283ba0793b0ccf0e9ec590) exists but they are monkey patching the code.

## What has been done

All imports are now relative.

```diff
- import Button from 'button'
+ import Button from '../../button'
```

## Things to consider

I splitted the refactoring into 2 commit to simplify the review:

- Update all import paths
- Fix ESlint import order

**UPDATE**: The change is marked as breaking because unresolved types are no longer resolved as `any` and can break usages.

## How it was tested

- Unit tests
- ESlint
- TypeScript compiler
- Manual check of the build output


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **[BREAKING CHANGE]** Use relative imports so applications can resolve types.
```
mcellier, line 3: Is it really breaking for the package consumers?

simonrelet: I just updated the description to explain why 👌

> **UPDATE**: The change is marked as breaking because unresolved types are no longer resolved as `any` and can break usages.

pith:
Wow! Thanks for that ❤️Did you script it?
Where you able to test this change in our main project?

tsconfig.json
```diff
@@ -15,8 +15,7 @@
     "allowJs": false,
     "declaration": true,
     "noEmitOnError": true,
-    "allowSyntheticDefaultImports": true,
-    "baseUrl": "./src"
```
grosbouddha, line 19: This should be enough to prevent people from using the previous syntax, correct ?

simonrelet:
> Wow! Thanks for that ❤️Did you script it?
> @pith 

Sort of, I generated a regex and used VS Code's search/replace:

Screenshot 2020-05-13 at 18 59 33

Then repeat it for each folder depth.

`eslint --fix` did the rest 👌

> Where you able to test this change in our main project?
> @pith 

Yes, TS compiler raised 30 errors.

tsconfig.json
simonrelet: Yes, without `baseUrl` TS compiler will used node's module resolution.

PR #718, opened 2020-05-13
Add optional prop to set a global className with FocusVisibleProvider

nickoola:
## What has been done

Add optional prop to set a global className with FocusVisibleProvider
It provide a global selector for Components/Elements which not use the focusContext for now.
It merge the behavior we already have on Kairos.

## How it was tested

- Locally with the Why component which already use the context 
- Unit tests with/without the new prop added

### DIFF
--- a/src/_utils/focusVisibleProvider/index.tsx
+++ b/src/_utils/focusVisibleProvider/index.tsx
@@ -17,15 +17,29 @@ const pointerMoveEventList = [
 
 const pointerDownEventList = ['mousedown', 'pointerdown', 'touchstart']
 
+export const FOCUS_VISIBLE_CSS_CLASS = 'focus-visible'
+
 type FocusVisibleProviderProps = {
   children: ReactNode
+  // Provide a global css selector for legacy components.
+  setGlobalClassName?: boolean
 }
 
 export const FocusVisibleContext = createContext(false)
 
-export const FocusVisibleProvider = ({ children }: FocusVisibleProviderProps) => {
-  /* When the provider first loads, assume the user is in pointer modality. */
+export const FocusVisibleProvider = ({
+  children,
+  setGlobalClassName,
+}: FocusVisibleProviderProps) => {
+  // When the provider first loads, assume the user is in pointer modality.
   const [hadKeyboardEvent, setHadKeyboardEvent] = useState(false)
+
+  if (setGlobalClassName) {
+    useEffect(() => {
+      // Add a 'focus-visible' CSS class on the <body> element.
+      document.body.classList.toggle(FOCUS_VISIBLE_CSS_CLASS, hadKeyboardEvent)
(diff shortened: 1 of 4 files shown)


src/_utils/focusVisibleProvider/index.tsx
```diff
@@ -17,16 +17,29 @@ const pointerMoveEventList = [
 
 const pointerDownEventList = ['mousedown', 'pointerdown', 'touchstart']
 
+export const FOCUS_VISIBLE_CSS_CLASS = 'focus-visible'
+
 type FocusVisibleProviderProps = {
   children: ReactNode
+  hasGlobalClassName?: boolean
```
grosbouddha, line 24: I would call it setGlobalClassName

src/_utils/focusVisibleProvider/index.tsx
```diff
@@ -17,16 +17,29 @@ const pointerMoveEventList = [
...
+export const FocusVisibleProvider = ({
+  children,
+  hasGlobalClassName,
+}: FocusVisibleProviderProps) => {
   /* When the provider first loads, assume the user is in pointer modality. */
   const [hadKeyboardEvent, setHadKeyboardEvent] = useState(false)
 
+  if (hasGlobalClassName) {
+    useEffect(() => {
+      // Provide a global css selector for legacy components
```
grosbouddha, line 38: Move this comment to the props itself to make it more discoverable.
// Add a 'focus-visible' CSS class on the < body > element.

nickoola: Done

src/_utils/focusVisibleProvider/index.tsx
nickoola: Done

src/_utils/focusVisibleProvider/index.tsx
```diff
@@ -17,16 +17,30 @@ const pointerMoveEventList = [
...
-  /* When the provider first loads, assume the user is in pointer modality. */
+export const FocusVisibleProvider = ({
+  children,
+  setGlobalClassName,
+}: FocusVisibleProviderProps) => {
+  // When the provider first loads, assume the user is in pointer modality.
   const [hadKeyboardEvent, setHadKeyboardEvent] = useState(false)
 
+  if (setGlobalClassName) {
+    useEffect(() => {
```
simonrelet, line 38: Hooks should not be called conditionally, it's [the first rule of hooks](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level).

You should move the condition inside of the effect:

```ts
useEffect(() => {
  if (setGlobalClassName) {
    document.body.classList.toggle(FOCUS_VISIBLE_CSS_CLASS, hadKeyboardEvent)
  }
}, [hadKeyboardEvent, setGlobalClassName])
```

PR #720, opened 2020-05-14
Added Search form animations for large devices

wakooka:
## Description

Adding animations for small devices on search form
Kapture 2020-05-14 at 16 44 54

## What has been done

- Regroup animation config in the same file
- Used Styled component to share animation css between search form section components
- Added same z-index value as modals (temporary until we moved z-index logic from SPA to ui-lib)
- Used `position: fixed` instead of `absolute` so that it's top-positionned regardless of scroll in main document
 
## Things to consider
There was no actual specs on this animation only a video, so this is a first shot at animation but will need to be discussed with UX designers.

## How it was tested

Tested locally in storybook:
- desktop Chrome, Firefox, Safari.
- mobile Android Chrome

### DIFF
--- a/src/searchForm/autoComplete/section/AutoCompleteSection.tsx
+++ b/src/searchForm/autoComplete/section/AutoCompleteSection.tsx
@@ -5,6 +5,7 @@ import { AutoCompleteProps } from '../../../autoComplete'
 import Button, { ButtonStatus } from '../../../button'
 import Icon from '../../../icon/chevronIcon'
 import Section from '../../../layout/section/baseSection'
+import { TransitionSection } from '../../baseStyles'
 
 export type AutoCompleteSectionProps = Omit<
   AutoCompleteProps,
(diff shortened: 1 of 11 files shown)


src/searchForm/autoComplete/section/AutoCompleteSection.tsx
```diff
@@ -19,6 +21,8 @@ export type AutoCompleteSectionProps = Omit<
   onClose: () => void
 }
 
+const TransitionAutoComplete = styled(TransitionBaseStyle)``
```
simonrelet, line 24: There is an empty styled component here ☝️

Same in _src/searchForm/datePicker/section/DatePickerSection.tsx_ and _src/searchForm/stepper/section/StepperSection.tsx_

wakooka: Typescript is complaining, it's expecting a: ``

PR #721, opened 2020-05-15
Added padding-top on Itinerary

wakooka:
## Description

We have inconsistency on vertical spacing on Itinerary in the SPA. To fix that we should use the vertical spacing given by the specs and set it in the component itself. Once done we need to remove custom styles that we set on SPA

## What has been done

Adding top padding on Itinerary component

## How it was tested

Tested on storybook

### DIFF
--- a/src/itinerary/index.tsx
+++ b/src/itinerary/index.tsx
@@ -16,6 +16,9 @@ const StyledItinerary = styled(Itinerary)`
     position: relative;
     list-style-type: none;
   }
+  & {
+    padding-top: ${space.m};
+  }
 
   & .kirk-itinerary--highlightRoad .kirk-itineraryLocation-road {
     background-color: ${color.midnightGreen};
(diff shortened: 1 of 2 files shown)


src/itinerary/index.tsx
```diff
@@ -16,6 +16,9 @@ const StyledItinerary = styled(Itinerary)`
     position: relative;
     list-style-type: none;
   }
+  & {
+    padding-top: ${space.m};
```
pith, line 20: There is no padding-bottom?

mcellier: The padding-bottom is set after all itinerary leg.

wakooka: > The padding-bottom is set after each itinerary leg.

Yes this is not ideal, but it will require a bigger refacto, if we want to change that.

PR #731, opened 2020-05-19
Fix crash on IE11 with MainContent component

PerrineBoissieres:
## Description

MainContent component was crashing on IE11 with an error on `matches` function from the `closest` polyfill.

## What has been done

Event though the polyfill was taken from Mozilla, we didn't check that everything was compatible with IE11 (and the `matches` function wasn't). So I've added the polyfill for it since it was pretty simple.

## How it was tested

Tested locally on IE11 through Saucelab. Prod version was crashing, while new version wasn't, so I guess it's ok ^^

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **[FIX]** Fix crash on IE11 for `MainContent` component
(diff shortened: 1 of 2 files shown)


wakooka:
Update changelog

PR #736, opened 2020-05-25
Deprecate old aria props and replace with A11yProps type (WIP)

PerrineBoissieres:
## Description

We have a lot of components using `ariaX` props, defined per component, alongside other custom props for sub-components a11y props. This becomes quite difficult to expand and maintain, since when we want to add new a11y props for interactions (ex: Hint component needs to add `aria-describedby` attribute on the component targeted by the Hint, which can be multiple things from a Text to an Item or a Button).
For the Hint component, we've added a new props interface named A11yProps that whitelists generic a11y attributes that can be used on any component. A few components support it today (Items mostly), but we need to expand it to all of our components that currently use a11y specific props and clean all those old props.

## What has been done

Here are the impacts on the props:

**Global:**
ariaLabel -> aria-label
ariaLabelledBy -> aria-labelledby
ariaDescribedBy -> aria-describedby

**Item:**
rightTitleAriaLabel -> rightTitleAriaProps `{ aria-label: '' }`

**ItemData:**
dataAriaLabel -> dataAriaProps `{ aria-label: '' }`

**Itinerary:**
collapsedAriaLabel -> collapsedAriaProps `{ aria-label: '' }`

## Things to consider

Those are breaking changes, so we would need to carefully update everything once released.

## How it was tested

Locally, through UT and stories :)


src/successModal/SuccessModal.tsx
```diff
@@ -51,7 +51,7 @@ export const SuccessModal = (props: SuccessModalProps): JSX.Element => {
       forwardedRef={forwardedRef}
       className={className}
       modalContentClassName={baseClassName}
-      ariaLabelledBy={successContentId}
+      aria-labelledBy={successContentId}
```
Tuxosaurus, line 54: "aria-labelledBy" => "aria-labelledby"

src/phoneField/PhoneField.tsx
```diff
@@ -256,11 +255,11 @@ export class PhoneField extends PureComponent<PhoneFieldProps, PhoneFieldState>
 
     return (
       <div className={cc([baseClassName, prefix({ error: !!error }), this.props.className])}>
-        <div id={id} className={classNames} aria-labelledby={ariaLabelledBy}>
+        <div id={id} className={classNames} {...a11yAttrs}>
```
Tuxosaurus, line 258: `id={id}` should also be generated by `{...a11yAttrs}`
I'm not sure it's a good idea to have `id` handled by A11yProps but since it's there already.

src/itemData/ItemData.unit.tsx
```diff
@@ -38,10 +37,3 @@ it("Shouldn't display left addon button if no main title", () => {
   )
   expect(wrapper.find('.kirk-item-title--withButtonAddon button').exists()).toBe(false)
 })
-
-it('Should display right data with aria-label attribute', () => {
```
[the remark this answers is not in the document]
PerrineBoissieres: You're right, done :)

src/itemRadio/story.tsx
```diff
@@ -73,7 +73,7 @@ stories.add('Multiple items with chevrons (form step with auto submit)', () => (
       onChange={action('changed group')}
       withSeparators
       withChevrons
-      ariaLabelledBy="id-title"
+      aria-labelledBy="id-title"
```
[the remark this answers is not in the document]
PerrineBoissieres: Fixed for consistency. It doesn't crash anything, but full lower case should be the way to go.

src/phoneField/PhoneField.tsx
PerrineBoissieres: For the logic of the ID, it's mainly because a lot of aria attributes are also used alongside an `id`: `aria-labelledby`, `aria-describedby`, etc... all point to an `id`. So I've added it to the list so we can have those behaviors working without  having to add extra props.

Removed the specific call to `id` there since it's handled by `a11yAttrs`.

PR #738, opened 2020-05-26
Release/34.0.1

Tuxosaurus:
# v34.0.1 (26/05/2020)
- **[UPDATE]** Add `flipped` prop in `MediaContentSection`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,10 @@
 # Unreleased
 
+[...]
+
+# v34.0.1 (26/05/2020)
+
 - **[UPDATE]** Add `flipped` prop in `MediaContentSection`.
-  [...]
 
 # v34.0.0 (25/05/2020)
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@blablacar/ui-library",
-  "version": "34.0.0",
+  "version": "34.0.1",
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 # Unreleased
 
+[...]
+
+# v34.0.1 (26/05/2020)
```
simonrelet, line 5: Shouldn't it be minor release (`34.1.0`) instead of a patch (`34.0.1`)?

Tuxosaurus: I wasn't completely sure, I guessed it was minor because it's one change on one component for a specific case. I'd put "minor" for a new component or a multi-file change/refacto

simonrelet: If we follow Semantic Versioning (as node packages should) we should have:

> 1. MAJOR version when you make incompatible API changes,
> 2. MINOR version when you add functionality in a backwards compatible manner, and
> 3. PATCH version when you make backwards compatible bug fixes.
>
> [Semantic Versioning](https://semver.org/)

Tuxosaurus: Our wiki says:
```
patch is to use for a release containing bug fixes and small additions
minor is to use for backwards-compatible features
major is to use for breaking changes
```

But I can agree to use the "official" definition, there would be less confusion

PR #739, opened 2020-05-26
Introduce testing-library

simonrelet:
## Description

Support [React Testing Library](https://testing-library.com/docs/react-testing-library/intro) in our tests.

## What has been done

- Update Jest config
- Add ESLint recommended rules for RTL and jest-dom
- Write `HeroSection` and `UneditableTextField` tests with RTL.

## How it was tested

With the unit tests 🤷‍♂️


mcellier:
Can we have an example with snapshots too?

simonrelet:
> @mcellier 
> * Can we have an example with snapshots too?

I didn't find a good use case for snapshot testing, if you have one I can update it too.

> * (Maybe) add an entry to the changelog

I didn't update it because this PR doesn't changes anything from the developer user point of view, i.e. it doesn't change the production code.

mcellier:
> > @mcellier
> > 
> > * Can we have an example with snapshots too?
> 
> I didn't find a good use case for snapshot testing, if you have one I can update it too.

Simple usage: `UneditableTextField`

> > * (Maybe) add an entry to the changelog
> 
> I didn't update it because this PR doesn't changes anything from the developer user point of view, i.e. it doesn't change the production code.

We're pretty inconsistent on that part, but I agree with you.

simonrelet:
> @mcellier
> Simple usage: `UneditableTextField`

Done in this PR, snapshots were not needed 😉.

src/uneditableTextField/index.unit.tsx
```diff
@@ -1,41 +1,74 @@
...
-    const wrapper = renderer.create(<UneditableTextField>Hello world</UneditableTextField>)
-    expect(wrapper.toJSON()).toMatchSnapshot()
+  it('should render the content', () => {
+    const props = createProps()
+    const { container } = render(<UneditableTextField {...props} />)
+
+    expect(container.firstChild).toHaveTextContent(props.children as string)
+  })
+
+  it('should apply public CSS classes', () => {
```
simonrelet, line 19: I not sure we should have CSS classes in the public API of a component but they are used so I added a specific test to be isofunctional.

src/uneditableTextField/index.unit.tsx
```diff
@@ -1,41 +1,74 @@
...
   })
 
-  it('Should have the text ellipsed', () => {
-    const wrapper = renderer.create(
-      <UneditableTextField ellipsis>
-        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec nec tristique sapien, eu
-        placerat justo. Donec tempor, risus ac cursus fringilla, lorem ipsum facilisis tortor, vel
-        molestie sapien justo nec orci.
-      </UneditableTextField>,
+  it('Should not add CSS class for text ellipsis by default', () => {
```
simonrelet, line 27: A screenshot test would be the best here.

PR #741, opened 2020-06-01
Fix back button alignment in SearchForm autocomplete sections

simonrelet:
## Description

Fix the alignment of the back button in the search form autocomplete sections.

**Before**

Screenshot 2020-06-01 at 11 11 56

**After**

Screenshot 2020-06-01 at 11 02 03

## What has been done

Override the button styles for this "specific" usage. It doesn't look supported 🤷‍♂️.

## Things to consider

Don't we have a generic component for this?

## How it was tested

- Storybook

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **[FIX]** Fix back button alignment in `SearchForm` autocomplete sections.
 [...]
 
 # v34.1.0 (27/05/2020)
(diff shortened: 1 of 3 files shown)


Tuxosaurus:
Was the height change requested? It looks more appropriate size-wise but now it's the vertical alignment that looks a bit weird.
Else it's OK

simonrelet:
> @Tuxosaurus 
> Was the height change requested? It looks more appropriate size-wise but now it's the vertical alignment that looks a bit weird.

It was not requested but it does look weird.

The design show that the input component should have a height of 72px (all spacing included) but in implementation for the place sections it is 52px, and for the date and passengers seats it is 56px (which use `Item` component).

The `Divider` also have its own "margin" (it is actually a height), which is not taken into account in the design 🤷‍♂️.

PR #743, opened 2020-06-01
Add disabled prop to Stepper

mcellier:
## Description

For trip edit, we need to be able to disable the stepper (i.e., the driver can't update the seat count if a booking is in progress).
We could cheat by setting min/max values equal to the selected value, but MEH.

## What has been done

Added `disabled` prop to the `Stepper` component

## Things to consider

Reusing the disabled display we have when hitting min/max values

## How it was tested

Storybook on chrome

### DIFF
--- a/src/stepper/Stepper.tsx
+++ b/src/stepper/Stepper.tsx
@@ -41,6 +41,7 @@ export interface StepperProps {
   display?: StepperDisplay
   focus?: boolean
   leftAddon?: React.ReactNode
+  disabled?: boolean
 }
 
 interface StepperState {
@@ -189,6 +190,7 @@ export default class Stepper extends PureComponent<StepperProps, StepperState> {
       max,
       valueClassName,
       display,
+      disabled,
     } = this.props
 
     const isMax = this.state.value >= max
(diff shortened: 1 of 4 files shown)


src/stepper/Stepper.tsx
```diff
@@ -230,11 +235,14 @@ export default class Stepper extends PureComponent<StepperProps, StepperState> {
...
             className="kirk-stepper-increment"
             status={ButtonStatus.UNSTYLED}
-            disabled={isMax}
+            disabled={disabled || isMax}
             isBubble
             {...this.createButtonListeners(this.increment)}
           >
-            <PlusIcon iconColor={isMax ? color.lightGray : color.blue} size={buttonSize} />
+            <PlusIcon
+              iconColor={disabled || isMax ? color.lightGray : color.blue}
```
Tuxosaurus, line 243: Icons all have the "isDisabled" (I know) prop that should be enough to gray them out

mcellier: Can't find the code related to that. Can you point it out?

Tuxosaurus: https://github.com/blablacar/ui-library/blob/master/src/_utils/icon/index.tsx#L14

mcellier: Thx ! I'll try to clean this a bit

mcellier: The difference: disabled icon is gray, disabled stepper icon is lightGray 

```
  gray: '#DDD',
  lightGray: '#EDEDED',
```

Is it an issue or can we just put gray everywhere?

Ping @wakooka

mcellier: Now using `isDisabled` from icons

PR #744, opened 2020-06-01
Remove hover & pointer on disabled ItemCheckbox

mcellier:
## Description

Impact both `ItemCheckbox` & `ItemRadio` components. Demo for `ItemCheckbox`:

Before:
before

After:
after

## What has been done

- Remove `cursor: pointer` when `disabled` is true
- Set `isClickable` prop to false when `disabled` is true to remove the background 

## Things to consider

Some snapshots will be updated when upgrading kairos

## How it was tested

With storybook

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 [...]
+- **[FIX]** Remove hover color & pointer when `disabled` is passed to `ItemCheckbox`/`ItemRadio`
 
 # v34.1.0 (27/05/2020)
 
--- a/src/itemCheckbox/index.tsx
+++ b/src/itemCheckbox/index.tsx
@@ -4,7 +4,7 @@ import ItemCheckbox from './ItemCheckbox'
 
 const StyledItemCheckbox = styled(ItemCheckbox)`
   & {
-    cursor: pointer;
+    cursor: ${props => (props.disabled ? 'default' : 'pointer')};
   }
 
   & input {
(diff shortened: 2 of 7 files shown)


src/itemCheckbox/index.tsx
```diff
@@ -11,6 +11,10 @@ const StyledItemCheckbox = styled(ItemCheckbox)`
     position: absolute;
     clip: rect(0, 0, 0, 0);
   }
+
+  &.kirk-item-checkbox--disabled {
+    cursor: default;
```
Tuxosaurus, line 16: Maybe remove the new class and use the prop instead (I received this same feedback when disabling icons):
```
& {
    cursor: ${props => (props.isDisabled ? 'default' : 'pointer')};
  }
```

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 [...]
+- **[FIX]** Remove hover color & pointer when `disabled` is passed to `ItemCheckbox`
```
Tuxosaurus, line 4: I guess this should also be applied to `ItemRadio`

nickoola: Yep, at least create a ticket if you prefer doing it separatly @mcellier

src/itemCheckbox/index.tsx
mcellier: Thanks, it's way cleaner that way

CHANGELOG.md
mcellier: Done

wakooka:
Would it be possible to do this on the Item directly only based on the `disabled` prop?
Probably need a bigger refacto.

mcellier:
> Would it be possible to do this on the Item directly only based on the `disabled` prop?
> Probably need a bigger refacto.

All items aren't clickable or else. I don't want to touch the base `Item` too much, as it's a mess :/

PR #746, opened 2020-06-04
Remove memoization on icons

simonrelet:
## Description

[`React.memo`](https://reactjs.org/docs/react-api.html#reactmemo) is a performance optimisation feature that skips an update if the props haven't changed. This works great for _heavy_ components with few props.

This PR remove calls to `React.memo` for the following reasons:
1. Our icons are quite small, memoizing them is not a real performance gain.
2. It looks like Enzyme's `find()` function doesn't work with memoized components which forces us to have multiple exports for tests.

## What has been done

Remove all `React.memo` in icons (it was only used there).

## Things to consider

This PR will ease the refactoring to use named exports instead of default exports because there will be less exported symbols for the tests (see _2._ above).

## How it was tested

- Unit tests
- Storybook

### DIFF
--- a/src/icon/airConditioning.tsx
+++ b/src/icon/airConditioning.tsx
@@ -15,4 +15,4 @@ export const AirConditioning = (props: IconPropsWithStatus) => (
 
 AirConditioning.defaultProps = StatusIcon.defaultProps
 
-export default React.memo(AirConditioning)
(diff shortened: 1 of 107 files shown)


grosbouddha:
Thanks for this Simon and sorry for the late review ! LGTM

PR #747, opened 2020-06-04
Add onClick to ItemEditableInfo

grosbouddha:
- Add onClick to ItemEditableInfo
- Make clickable item use a cursor:pointer style

TESTED=Locally in storybook + unit tests


Tip-Sy:
👍

PR #749, opened 2020-06-05
Use named exports

simonrelet:
## Description

As named import/export are preferred to default ones, this PR is a refactoring to use named import/export.

## What has been done

- Make sure all components have a named export.
- Only use the named imports in the code.

## Things to consider

- Aliases were used when components names were not the correct ones to avoid changing the code to much.
  ```diff
  - import Section from '../layout/section/baseSection'
  + import { BaseSection as Section } from '../layout/section/baseSection'
  ```
- Default export have been kept in index files to make sure this PR isn't breaking.
  Once application have been refactored, we can remove all default exports.

## How it was tested

- Unit tests
- Storybook
- TypeScript Compiler


src/_utils/branding.ts
```diff
@@ -144,11 +144,13 @@ export const responsiveBreakpoints = {
 
 export const pxToInteger = (size: string): number => parseInt(size.replace('px', ''), 10)
 
-export default {
+export const branding = {
```
simonrelet, line 147: This export has been kept because a `branding` object is exported from the package entry point.

simonrelet:
> Did you do that by hand ? :)

Some parts yes, other parts with RegExp 🤖

PR #750, opened 2020-06-10
[BBC-8438] `ItemBigData` Component

gabrielagabriel:
## Description

Screenshot 2020-06-10 at 17 04 13

## What has been done

- Implemented a first version of `ItemBigData` based on `Item`

## Things to consider

- Updated `ItemData` doc to be able to compare and test both components at the same time

## How it was tested

- 🤙  Manually


src/_utils/item/Item.tsx
```diff
@@ -101,6 +101,24 @@ export const Item = (props: ItemProps) => {
 
   const getTextColor = (textColor: string) => (disabled ? color.gray : textColor)
 
+  const getLeftTitle = (value: any) => {
+    if (typeof value === 'object') {
+      return leftTitle
```
wakooka, line 106: I find this way too flexible, I think the component API should be more restrictive.
I'd remove this, and use `leftTitleDisplay` prop instead, wdyt? So we can keep `leftTitle` prop as string.

gabrielagabriel: I could make it restrictive to any typography component. 
But you are right that **ONLY** `BigData` uses `Display2` and all other `Item` use `Title`.
I just don't like the idea to have one prop by component...

wakooka: > I just don't like the idea to have one prop by component...
What do you mean?

gabrielagabriel: Creating `leftTitleDisplay` for instance because the API goes with implementation details instead of what is the function of this property which is to have a `LeftTitle`. 

If you want to keep `Title` and do not touch `Item` API. I can override on `BigData` CSS.  It will have less impact, I guess.

_(really sad though, I can't override the component itself since this component doesn't follow a `styled-components` syntaxe)_

**Should I override title by CSS**?

grosbouddha: I would go for a list of typogaphic styles as a param type (VS CSS or having separate pair of props for each styles).

grosbouddha: I think ultimately we need to get rid of these leftTitleDisplay props: the combinatory explodes and allow to create typographic styles that don't exists (since you can change independantly font size/color/etc)

src/_utils/item/Item.tsx
```diff
@@ -101,6 +101,24 @@ export const Item = (props: ItemProps) => {
 
   const getTextColor = (textColor: string) => (disabled ? color.gray : textColor)
 
+  const getLeftTitle = (value: any) => {
```
grosbouddha, line 104: You can create a letTitle type and keep type safety (instead of using any):

```
type  leftTitleType = string | React.ReactNode
... 
readonly leftTitle?: leftTitleType
...
const getLeftTitle = (value: leftTitleType) => {
...
```

src/itemBigData/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+`
+
+export const ItemBigData = (props: ItemBigDataProps) => {
+  const { mainInfo, className, mainTitle, tag, ariaLabel } = props
+  const a11yAttrs = pickA11yProps<ItemBigDataProps>(props)
+
+  return (
+    <StyledItemBigData
+      className={className}
+      leftTitle={<TextDisplay2>{mainTitle}</TextDisplay2>}
```
grosbouddha, line 36: Nice !

src/_utils/item/Item.tsx
simonrelet: [Compound components](https://www.google.fr/search?q=react+compound+components) would resolve this kind of issues 😉

PR #751, opened 2020-06-12
[BBC-8503] Ride Axis

gabrielagabriel:
## Description

**From-ArrowIcon-To** is a repeated pattern. There is an opportunity to refactor:

axis-before-after

## What has been done

- New component `RideAxis`

## Things to consider

- Update `IconBase` to be able to adapt to text properties (font-size and color)

### Impact
**MessagingSummaryItem** 
https://blablacar.github.io/ui-library/?path=/story/widgets-messagingsummaryitem--with-html-sublabel

**Carpool**
https://blablacar.github.io/ui-library/?path=/story/pages-ride-details-carpool--default

**The right arrow**
https://blablacar.github.io/ui-library/?path=/story/widgets-searchrecap--default


src/_utils/icon/index.tsx
```diff
@@ -4,6 +4,8 @@ import { color, shadow } from '../branding'
 import { BaseIcon } from './BaseIcon'
 
 const StyledBaseIcon = styled(BaseIcon)`
+  stroke: ${props => (props.iconColor ? props.iconColor : 'currentColor')};
```
simonrelet, line 7: nit: You can write it using the nullish coalescing operator `??`:

```ts
props.iconColor ?? 'currentColor'
```

src/rideAxis/index.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+import styled from 'styled-components'
+
+import { space } from '../_utils/branding'
+import { ArrowIcon } from '../icon'
+
+const StyledRideAxis = styled.span`
+  display: flex;
+  align-items: center;
+  font-size: 1em;
```
simonrelet, line 10: Is this really needed? I think omitting `font-size` (and `color`) will do the same.

src/rideAxis/specifications/index.md
```diff
@@ -0,0 +1,27 @@
...
+
+<section>
+    <!-- STORY -->
+</section>
+
+---
+
+## Specifications
+
+- `Font-size` and `Color` are _inhehit from parent_
```
wakooka, line 20: typo: `inherit`

src/rideAxis/specifications/index.md
```diff
@@ -0,0 +1,27 @@
+# Ride Axis
```
wakooka, line 1: Since it's not part of the design system, should we put this kind of components in another bucket?
For example we discussed having `Internal` suffix when a component should only be used in ui-library, maybe we should have a way to identify components that we created because it's easier on the technical side. Another approach would be to ask designers to add this component in the design system. Wdyt?

gabrielagabriel: I'll move to '_utils' for now

simonrelet: I'm afraid suffixing all private symbols (components, hooks, functions, constants, etc.) will be too cumbersome.
I simple solution can be to use an __internal/_ folder.

gabrielagabriel: Yeah.. that's why I'm moving the component there and creating a "Internal" Section on Storybook taxonomy

PR #755, opened 2020-06-19
Add aria-busy, aria-live and aria-label on the Loader component

pith:
## Description

Add the `aria-busy` on the loader component to indicates its loading states along with a `aria-live="polite"` attribute.

Ressources: https://www.accessibilityresources.org/aria-busy

## What has been done

* Add aria attributes
* Rewrite the tests with RTL

## Things to consider

* nothing in particular

## How it was tested

* unit tests

### DIFF
--- a/src/loader/Loader.tsx
+++ b/src/loader/Loader.tsx
@@ -26,6 +26,7 @@ export interface LoaderProps {
   size?: number
   done?: boolean
   layoutMode?: LoaderLayoutMode
+  ariaLabel?: string
   onDoneAnimationEnd?: () => void
 }
 
@@ -79,7 +80,7 @@ export class Loader extends PureComponent<LoaderProps> {
   }
 
   render() {
-    const { className, size, done } = this.props
+    const { className, size, done, ariaLabel } = this.props
     const iconSize = {
       width: `${size}px`,
       height: `${size}px`,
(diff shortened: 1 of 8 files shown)


src/loader/Loader.tsx
```diff
@@ -79,15 +80,21 @@ export class Loader extends PureComponent<LoaderProps> {
...
       width: `${size}px`,
       height: `${size}px`,
     }
 
     return (
       <div className={cc([className, this.computeLayoutClass()])}>
-        <div className={cc([{ 'kirk-loader--done': done }])} style={iconSize}>
+        <div
+          className={cc([{ 'kirk-loader--done': done }])}
+          aria-busy={!done}
```
grosbouddha, line 93: Lot's of debatable loud thinking ahead:

I think this won't work. The aria-busy should be applied to the busy subtree that will be updated.
Here, the loader subtree contains nothing. We could just apply aria-busy to the body itself or the main role element ? that will work semantically when the loader is in fullscreen mode but not when it's in other local modes (inline for instance) but maybe that's good enough.

Similarly for aria-live, there is nothing inside the loader (during or after loading) so no 'live' content to announce. Unless combining it with with aria-label has some side effect and it announces the aria label ? In that case, I would rename the props aria-label to loadSuccesfullAriaAnnouncement to reflect how it's supposed to be used API wise (but still use it as aria-label internally).

More generally, I think it's weird that the loader takes care of a global application state and I think it should be the other way around. E.g. a global application state in kairos (depending on saga's states for instance) should propagate to the rest of the UI: then the loading container can set itself as aria-busy and the loader can change independently.

wdyt ?

grosbouddha: As well, if you want to programmatically announce the state of a loading state, we could use a programmatic announcer (in kirk or kairos) to do the job (it's mostly a more element aria-live element). For instance:
https://www.npmjs.com/package/react-a11y-announcer

pith: Thanks for your feedback, indeed I got this wrong. We can discuss how to implement the other approach you suggest.

PR #756, opened 2020-06-22
Fix Pet icon stroke color

grosbouddha:
The stroke color was defined 2 times as a prop and inside the icon SVG. The internal fill stroke color was removed (like in other icons) to allow the override from the props.

TESTED=Locally


gabrielagabriel:
We do have an issue with Icons format, they also receive the color by the `fill` attribute

PR #761, opened 2020-06-23
Move unused utils files to an _internals folder

pith:
## Description

Move all the utils files and folders that are unused in client apps to an `_internals` folder.

## What has been done

image

## Things to consider

Nothing special

## How it was tested

*In progress*


grosbouddha:
What will be the semantic of this leading _ ?
Can we remove it from _utils after this PR ?

pith:
Yes removing the `_` in the `_utils` would be the next step. I could do it in this PR, so batch the breaking changes in one major version.

However I just tested the canary version and it seems that I have few regressions.

pith:
I successfully tested this PR with a canary release. So when I'll have fix all the conflicts, it should be ready for merge 😅 .

pith:
I rebased the PR and tested it in BBC with the canary version `0.0.0-2e35a272`. 
All the tests are green and I didn't notice any changes in localhost.

PR #764, opened 2020-06-25
Restoring Icons Fill

gabrielagabriel:
## Description

BBC Icons have a irregular stroke, to try to reduce this difference I've used `stroke` attributes to set colors` but it makes the icon look stronger than they really are.

**STROKE**
Screenshot 2020-06-25 at 16 57 10

Back to the old fill version, still irregular, but _we don't see them_

**FILL**
Screenshot 2020-06-25 at 16 57 55

## How it was tested

👀  eyes

### DIFF
--- a/src/_utils/icon/index.tsx
+++ b/src/_utils/icon/index.tsx
@@ -4,7 +4,7 @@ import { color, shadow } from '../branding'
 import { BaseIcon } from './BaseIcon'
 
 const StyledBaseIcon = styled(BaseIcon)`
-  stroke: ${props => props.iconColor ?? color.lightMidnightGreen};
+  fill: ${props => props.iconColor ?? color.lightMidnightGreen};
 
   &,
   & rect,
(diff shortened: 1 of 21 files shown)


src/_utils/icon/index.tsx
```diff
@@ -4,7 +4,7 @@ import { color, shadow } from '../branding'
 import { BaseIcon } from './BaseIcon'
 
 const StyledBaseIcon = styled(BaseIcon)`
-  stroke: ${props => props.iconColor ?? color.lightMidnightGreen};
+  fill: ${props => props.iconColor ?? color.lightMidnightGreen};
```
Tuxosaurus, line 7: This line didn't exist before the change

src/_utils/icon/index.tsx
```diff
@@ -13,7 +13,7 @@ const StyledBaseIcon = styled(BaseIcon)`
   & polyline,
   & path,
   & g {
-    stroke: ${props => (props.isDisabled ? color.gray : '')};
+    fill: ${props => (props.isDisabled ? color.gray : '')};
```
Tuxosaurus, line 16: `stroke` was already there before the last change

PR #766, opened 2020-06-26
Fix ItemsListDivider type comparison when using react-hot-loader

grosbouddha:
This will fix our garbled UserMenu (broken icons and links) and remove the start up console error (about bad hydratation of SSR UserMenu) in DEV mode.

TESTED=Locally


Tip-Sy:
👍

PR #767, opened 2020-06-26
Implement Loader basic a11y

grosbouddha:
Implement basic Loader a11y:
- add role progressbar on loader itself (it had no role before)
- use aria-valuetext for the progressbar with the default non-localized value: loading and loaded
- allow customization of these loading and loaded strings with 2 new optional props loadingAriaLabel and loadedAriaLabel
- Add a new non-visual role=status element to get announcements when the state changes (the progressbar does not do that).

It's only basic a11y since that will solve discoverability for screenreaders and tests (selection with a11y-based react testing lib queries).
Still needs to be done to be considered complete a11y-wise:
 - the state announcement should be done using some programmatic announcer instead of this role=status. That will allow us to get rid/factorize properly the 'visually-hidden' CSS class for instance and encourage people to announce their state if they have an easy to use tool.
 - a loader should mark some area in the page as aria-busy (probably using some callback mechanism, TBD)
 - when the loading resume, we should probably manage the focus in some cases (when the loader is fullscreen but maybe not when it's inline or block. Again we should probably use some callbacks to delegate this to the application, not the Loader itself).

TESTED=Unit tests

### DIFF
--- a/src/loader/Loader.unit.tsx
+++ b/src/loader/Loader.unit.tsx
@@ -30,6 +30,16 @@ describe('Loader', () => {
     expect(wrapper.find('.kirk-loader--inline').exists()).toBe(true)
   })
 
+  it('Should use custom loading announcements', () => {
+    const wrapper = mount(<Loader loadingAriaLabel="localized custom loading" />)
+    expect(wrapper.text()).toBe('localized custom loading')
+  })
+
+  it('Should use custom loaded announcements', () => {
+    const wrapper = mount(<Loader done loadedAriaLabel="localized custom loaded" />)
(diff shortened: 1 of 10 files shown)


src/loader/Loader.unit.tsx
```diff
@@ -30,6 +30,16 @@ describe('Loader', () => {
     expect(wrapper.find('.kirk-loader--inline').exists()).toBe(true)
   })
 
+  it('Should use custom loading announcements', () => {
```
pith, line 33: It would be nice to use RTL for new tests

PR #768, opened 2020-06-29
[boyscout] RideAxis polish style

gabrielagabriel:
## Description
Support longer texts on `RideAxis`. 

**BEFORE:**

Screenshot 2020-06-29 at 12 40 37

**AFTER:**
Screenshot 2020-06-29 at 12 38 43

## Things to consider

-  No official specs from Pixar

## How it was tested
- Applied to *Search History*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - [...]
+- **[UPDATE]** `RideAxis` alignment polish: when text goes from 1 to 2 lines.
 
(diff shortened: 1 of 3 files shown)


pith:
Why is this component in `_utils` ? It doesn't seem to be an utils method.

gabrielagabriel:
> @pith  Why is this component in _utils ? It doesn't seem to be an utils method. Is it internal or public?
It has no official spec from Pixar. I've been considering to move it to components, since its gonna be used on BBC App

PR #769, opened 2020-06-29
Update Node version to form 10.15.0 to 12.16.1

simonrelet:
## Description

Update Node version to use the LTS.

## What has been done

Update Node version to form 10.15.0 to 12.16.1 in the following files:

## How it was tested

* Unit test
* Build
* CI


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - '10.15'
+  - '12.16'
 before_install:
   - echo "//registry.npmjs.org/:_authToken=\${NPM_TOKEN}" > .npmrc
   - dpkg --compare-versions `npm -v` ge 6.0 || npm i -g npm@^6.0
```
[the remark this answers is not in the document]
simonrelet: Yes we could even though I'm not sure this line is really useful as [npm 6.13.4 is always installed with node 12.16.1](https://nodejs.org/en/download/releases/).

package.json
```diff
@@ -9,6 +9,10 @@
     "type": "git",
     "url": "git@github.com:blablacar/ui-library.git"
   },
+  "engines": {
+    "node": "^12.16.1",
```
grosbouddha, line 13: Why not freeze the exact version ?

simonrelet: Good question 👌
Because If I do, it will:

- Require direct and transitive dependencies that declare an `engines` key to have the exact same version.
- Require all applications using this library that declare an `engines` key to have the exact same version.

If these conditions are not meet, an "Unsupported engine for @blablacar/ui-library" will be thrown by npm during installation.

PR #777, opened 2020-07-01
Normalizing Item, TheVoice, Divider

wakooka:
## Description
Normalized horizontally `Item`, `Divider`, and `TheVoice` component.

## What has been done
- Adding 24px horizontally differently when Item is in the page or Item is in a Section or Card
- Simplified the Divider and use semantic tag `hr`, the use of absolute positioning was breaking the spacing.
- Adding `noHorizontalSpacing` optional on `Section`
- Updated `Item*` with the new `hasHorizontalSpacing` prop

## Things to consider

Trying to break the less possible things by using explicit rules padding-left/right and margin-left/right so that the changes don't apply to the vertical spacing.

On ride details view: 
- Keep existing `Section` for children that have not been normalized.
- Update `Section` with a new `noHorizontalSpacing` prop so that the section padding does not apply.

This is the approach I found the less risky, we can work iteratively component by component. That way having normalized components and not-normalized components in the same page is not an issue.
See https://github.com/blablacar/ui-library/pull/777/files#diff-4db8d39277f2b246fb53549b429444de
 
## How it was tested

Storybook 

image

### DIFF
--- a/src/_utils/branding.ts
+++ b/src/_utils/branding.ts
@@ -73,6 +73,12 @@ export const space: defaultBranding = {
   xxl: '48px',
 }
 
+export const horizontalSpace: defaultBranding = {
+  inner: '8px',
+  outer: '16px',
+  global: '24px',
+}
+
 export const radius: defaultBranding = {
   none: '0',
   s: '4px',
--- a/src/divider/Divider.tsx
+++ b/src/divider/Divider.tsx
@@ -3,28 +3,34 @@ import cc from 'classcat'
 import styled from 'styled-components'
 
 import { color, space } from '../_utils/branding'
+import { normalizeHorizontally, NormalizeProps } from '../layout/layoutNormalizer'
 
-export interface DividerProps {
+export interface DividerProps extends NormalizeProps {
   readonly className?: string
 }
 
 const StyledDivider = styled.div`
+  /* HorizontalNormalization */
   & {
-    // Height is used instead of margins to prevent collapsing margin issues with other
-    // margin-based components.
-    height: calc(2 * ${space.m});
-    position: relative;
+    ${normalizeHorizontally}
   }
 
-  &:after {
-    position: absolute;
-    top: ${space.m};
-    content: ' ';
-    border-top: solid ${color.lightGray} 1px;
-    width: 100%;
+  & {
+    /* Using padding to avoid collapsing margins with above and below components */
+    padding-top: ${space.m};
+    padding-bottom: ${space.m};
+  }
+
+  & > hr {
+    border: none;
+    background-color: ${color.lightGray};
+    height: 1px;
+    margin: 0;
   }
 `
 
 export const Divider = ({ className }: DividerProps) => (
-  <StyledDivider className={cc(className)} aria-hidden="true" />
+  <StyledDivider className={cc(className)} aria-hidden="true">
+    <hr />
+  </StyledDivider>
 )
--- a/src/layout/section/baseSection/baseSection.tsx
+++ b/src/layout/section/baseSection/baseSection.tsx
@@ -14,6 +14,7 @@ export interface BaseSectionProps {
   readonly role?: string
   readonly children: JSX.Element | string | React.ReactNode
   readonly contentSize?: SectionContentSize
+  readonly noHorizontalSpacing?: boolean
 }
 
 /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-- [...]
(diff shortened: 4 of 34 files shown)


src/_utils/branding.ts
```diff
@@ -73,6 +73,12 @@ export const space: defaultBranding = {
   xxl: '48px',
 }
 
+export const horizontalSpace: defaultBranding = {
```
gabrielagabriel, line 76: maybe we should keep t-shirt naming convention `s`,  `m`, `l`,...

src/_utils/item/index.tsx
```diff
@@ -1,13 +1,18 @@
 import styled from 'styled-components'
 
+import { normalizeHorizontally } from '../../layout/layoutNormalizer'
 import { color, space } from '../branding'
 import { Item } from './Item'
 
 const StyledItem = styled(Item)`
+  /* HorizontalNormalization */
+  & {
+    ${normalizeHorizontally}
```
gabrielagabriel, line 10: by the css cascade if you apply a padding here it will be replace for declations just after... 
so you can move after component declarion or
remove the `&` on css definition for the component and let the & here to bring specificity

src/divider/Divider.tsx
```diff
@@ -3,8 +3,11 @@ import cc from 'classcat'
 
 export interface DividerProps {
   readonly className?: string
+  readonly inBorderedComponent?: boolean
 }
 
 export const Divider = ({ className }: DividerProps) => (
-  <div className={cc(className)} aria-hidden="true" />
+  <div className={cc(className)} aria-hidden="true">
+    <hr />
```
gabrielagabriel, line 11: can't you use _only_  `hr`?

src/divider/index.tsx
```diff
@@ -1,22 +1,25 @@
...
   }
 
-  &:after {
-    position: absolute;
-    top: ${space.m};
-    content: ' ';
-    border-top: solid ${color.lightGray} 1px;
-    width: 100%;
+  & {
+    padding-top: ${space.m};
```
gabrielagabriel, line 14: I see.. using padding for normalization to avoid collapsing.. I think we should document that outter spaces are going to be used with padding to avoid margin collapse

src/layout/section/baseSection/baseSection.tsx
```diff
@@ -14,6 +14,7 @@ export interface BaseSectionProps {
   readonly role?: string
   readonly children: JSX.Element | string | React.ReactNode
   readonly contentSize?: SectionContentSize
+  readonly normalizedChildren?: boolean
```
gabrielagabriel, line 17: for the API naming, I think we should avoid technical terms and be more general like `nospacing`

src/_utils/item/Item.tsx
```diff
@@ -47,6 +47,7 @@ export interface ItemProps extends A11yProps {
   readonly onMouseDown?: (event: React.MouseEvent<HTMLElement>) => void
   readonly hideHoverBackground?: boolean
   readonly disabled?: boolean
+  readonly inBorderedComponent?: boolean
```
gabrielagabriel, line 50: API: _wrapped_? _isWrapped_? _contained_? _hasContainer_? _hasSpacing_? (to have an alusion with `nospacing`)

src/_utils/branding.ts
wakooka: I'd rather be explicit on their usage.

src/divider/Divider.tsx
wakooka: You answered your own question :D  https://github.com/blablacar/ui-library/pull/777#discussion_r448848822

src/divider/index.tsx
wakooka: Added a comment

src/_utils/item/Item.tsx
wakooka: Went with `hasHorizontalSpacing` & `noHorizontalSpacing`

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@blablacar/ui-library",
-  "version": "36.0.0",
+  "version": "0.0.0-6182619d",
```
gabrielagabriel, line 3: reminder to do not forget versionning :)

src/divider/Divider.tsx
```diff
@@ -3,28 +3,34 @@ import cc from 'classcat'
...
 }
 
 const StyledDivider = styled.div`
+  /* HorizontalNormalization */
   & {
-    // Height is used instead of margins to prevent collapsing margin issues with other
-    // margin-based components.
-    height: calc(2 * ${space.m});
-    position: relative;
+    ${normalizeHorizontally}
```
gabrielagabriel, line 15: you cans add that at same specificity leve defined at line 18

PR #780, opened 2020-07-07
[BBC-7102] Using Storybook Swatch and Type Scaling Components

gabrielagabriel:
## Description

https://www.loom.com/share/64767267ac394e3ba3786a19e01d66c3

## What has been done

- Replace _Colors_ and _Font_ export for the Storybook components
- `Spacing` still an script
- Rename typography  spec from `.md` to `mdx`
- Star to group components by "function" (new taxonomy to promote discoverability). 

## Things to consider

Notice that I have not re-written any documentation yet.. polish on all that will come later

## How it was tested

UT


src/_utils/rideAxis/story.tsx
```diff
@@ -7,7 +7,7 @@ import { color, font } from '../branding'
 import { RideAxis } from './index'
 import readme from './specifications/index.md'
 
-const stories = storiesOf('Internal|RideAxis', module)
+const stories = storiesOf('Widgets|RideAxis', module)
```
grosbouddha, line 10: this should not be here ?

src/typography/body/story.tsx
```diff
@@ -1,13 +1,12 @@
...
+import { boolean, text } from '@storybook/addon-knobs'
 import { storiesOf } from '@storybook/react'
 
 import { color } from '../../_utils/branding'
 import { BaseSection as Section } from '../../layout/section/baseSection'
 import { TextBody } from './index'
 
-const stories = storiesOf('Brand|typography', module)
-stories.addDecorator(withKnobs)
+const stories = storiesOf('Design Tokens|Typography/', module)
```
pith, line 9:
```suggestion
const stories = storiesOf('Design Tokens|Typography', module)
```

gabrielagabriel: The `/` is used to create a folder

src/subHeader/story.tsx
```diff
@@ -1,17 +1,12 @@
...
 
 import { BaseSection as Section } from '../layout/section/baseSection'
 import { SubHeader } from '../subHeader'
-import spec from './specifications/subHeader.md'
 
-const stories = storiesOf('Widgets|SubHeader', module)
-stories.addDecorator(withKnobs)
-const specTemplateFn = () => <SubHeader>SubHeader content</SubHeader>
-stories.add('Specifications', specTemplateFn, { readme: spec })
+const stories = storiesOf('Text|', module)
```
grosbouddha, line 7: Why not keep Text|SubHeader like other stories ?

src/typography/body/story.tsx
pith, line 9: Is the `/` expected in `'Design Tokens|Typography/'`? Could we share constants for these strings?

src/_utils/rideAxis/story.tsx
gabrielagabriel: I'm moving the widgets by group of function like "Design Tokens", `RideAxis` where all alone so I move it back to Widgets, I can do it on other PR, though.. no problem

src/typography/titleStrong/story.tsx
```diff
@@ -1,13 +1,12 @@
...
+import { boolean, text } from '@storybook/addon-knobs'
 import { storiesOf } from '@storybook/react'
 
 import { color } from '../../_utils/branding'
 import { BaseSection as Section } from '../../layout/section/baseSection'
 import { TextTitleStrong } from './index'
 
-const stories = storiesOf('Brand|typography', module)
-stories.addDecorator(withKnobs)
+const stories = storiesOf('Design Tokens|Typography/', module)
```
grosbouddha, line 9: Do we need the 'Design Tokens' level ? could we just go for Typography ?
It's not clear what's a design token (the meaning might vary depending on which medium article you read :/)

src/typography/body/story.tsx
gabrielagabriel: Yes.. it creates a "folder" structure. The way we declare hierarchy will change on 6.0, But good point we may think about this optimisation

src/subHeader/story.tsx
gabrielagabriel: 'cos "Default" can be declared direct as story, no need to create another level of hierarchy

src/typography/titleStrong/story.tsx
gabrielagabriel: A design token is an 'atom' basically. We can't do any design without colors, type, scalling... 
We can redefine a new taxonomy. I'm going to open a JIRA for that.

PR #781, opened 2020-07-08
[BBC-8963] ButtonGroup: first story on MDX

gabrielagabriel:
## Description

https://www.loom.com/share/8f61a7f2590541bd94a7749a4e2acf65

## What has been done

- Convert specifications.md to `Component.story.mdx``
- Update components exports

## Things to consider

- How we export our components with Styled-Components must be changed
**Before** switching to a simple export, `styled-components` props are part of components props

Screenshot 2020-07-08 at 14 44 43

Exporting the component now is only:
```
export { ButtonGroup, ButtonGroupProps } from './ButtonGroup'
 ```

This method of export avoids reverse import (@simonrelet  can explain better than me the issues on that), and avoid importing the component without style as @pith mentioned on the last chapter

## How it was tested

- Storybook

### CODE REVIEW

File: src/buttonGroup/ButtonGroup.tsx [hunk 1]
```diff
@@ -1,9 +1,10 @@
 import React, { cloneElement } from 'react'
```
wakooka, line 1: Could we rename this file index.tsx? To avoid doubling name in the import?
`import { ButtonGroup } form `@blablacar/ui-library/...../buttonGroup/ButtonGroup`
vs
`import { ButtonGroup } form `@blablacar/ui-library/...../buttonGroup`

Wdyt?

src/buttonGroup/buttonGroup.story.mdx
```diff
@@ -0,0 +1,73 @@
...
+
+Use it at the end of a flow. Allows the member to make a final decision.
+
+- ✅ Always at the bottom of the screen
+- ✅ Can be sticky
+- ⛔️ Labels can't go on 2 lines. If too long we use "…"
+
+## TODO
+
+- [ ] Force button types
```
pith, line 73: Is it a Pixar todo or is it for us?

gabrielagabriel: It was there, I think we should remove the "TODOs"

src/buttonGroup/ButtonGroup.tsx
gabrielagabriel: Usually, `index.js` is used only to export files. I'm ok on keeping it named as the component name.

@simonrelet  what are the best practices on exporting files?

pith: I think this was to avoid breaking changes, but I agree on the long term, we should need only one file. However if we do this refactoring on all the files it could be nice to batch the breaking changes into one PR. I don't know what you have in mind @gabrielagabriel for that?

pith:
I really like this new documentation, thanks for that!

src/buttonGroup/ButtonGroup.tsx
gabrielagabriel: yes.. this avoids breaking changing because for now we are only updating the documentation. I rather move files on another PR

simonrelet: > @gabrielagabriel 
> @simonrelet what are the best practices on exporting components?

What I think is:

- Imports ending with `/buttonGroup/ButtonGroup` are probably mistakes, I don't think components were ment to be imported without their styles, that being said, it wasn't clearly explain what is public and what is internal, so it can be considered a braking change.

- I think we should only be allowed to import from the package root: `import { ButtonGroup } from '@blablacar/ui-library'`, the internal structure is not part of the public API and tree shaking is [not a problem with these imports/exports with Webpack](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free). As it would require a longer change (for a next time), I think we should now have a minimal path: `/buttonGroup`. The index file either containing the component or just import/export would work, I have a small preference for the second one (just import/export) as it ease discoverability and is more consistent:
  ```
  ┬╴ buttonGroup/
  ├╴ index.tsx
  ├╴ ButtonGroup.tsx       ⬅ Component code here.
  ├╴ ButtonGroup.test.tsx
  ├╴ ButtonGroup.story.mdx
  └╴ ...
  ```

PR #783, opened 2020-07-08
[BBC-8974] Divider.mdx

gabrielagabriel:
## Description

screencapture-localhost-60907-2020-07-08-18_18_51

## What has been done

- Divider.md to .mdx

## How it was tested

- UT
- Storybook


src/divider/Divider.story.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+```jsx
+import { Divider } from '@blablacar/ui-library'
+
+Some content to divide...
+<Divider />
+...Rest of the divided content.
+```
+
+<Props of={Divider} />
```
grosbouddha, line 33: 😮

src/divider/Divider.tsx
```diff
@@ -1,10 +1,30 @@
...
 import cc from 'classcat'
+import styled from 'styled-components'
+
+import { color, space } from '../_utils/branding'
 
 export interface DividerProps {
   readonly className?: string
 }
 
+const StyledDivider = styled.div`
```
grosbouddha, line 11: I think we had separate files to not mix a bunch of CSS bloat with the actual code of the component. It's ok-ish for small components but for bigger ones, I think it's nicer to keep them separated. 
@wakooka thoughts ?

gabrielagabriel: Next bigger one, I'll put styles aside.

simonrelet: Putting the styles in a separate file can lead to move out prop types if they are needed in the styles:

```tsx
// ComponentProps.ts
export type ComponentProps = {
   hasError: boolean
}

// StyledComponent.ts
import { ComponentProps } from './ComponentProps'

export const StyledComponent = styled.div<ComponentProps>`
  color: ${props => props.hasError ? 'red' : 'inherit'}
`

// Component.tsx
import { ComponentProps } from './ComponentProps'
import { StyledComponent } from './StyledComponent'

export function Component({ hasError }: ComponentProps) {
  return <StyledComponent hasError={hasError} />
}
```

I don't think we should have one large styled component, but rather multiple small ones.

```tsx
// Instead of
const StyledComponent = styled.div`
  // ...

  & .element-aaa {
    // ...
  }

  & .element-bbb {
    // ...
  }

  & .element-ccc {
    // ...
  }
`

// Have
const StyledComponent = styled.div`
  // ...
`

const StyledElementAAA = styled.div`
  // ...
`

const StyledElementBBB = styled.div`
  // ...
`

const StyledElementCCC = styled.div`
  // ...
`
```

PR #795, opened 2020-07-15
Grip component

PerrineBoissieres:
## Description

I still have a lot of work on the layout part, but at least here's the cleaned-up/basic new Grip component before my PR gets too massive.

## What has been done

This component adds the GripHandle in the DOM, and listens to touch events for slide up and slide down interactions. The story is really basic with no layout shenanigans yet (this will be another component, in another PR, still WIP).

## Screenshot

Capture d’écran 2020-07-15 à 12 07 57

## How it was tested

Local Chrome with emulated touch events, and Safari iOS (both latest versions)


src/grip/index.tsx
```diff
@@ -0,0 +1,9 @@
+import styled from 'styled-components'
```
gabrielagabriel, line 1: Hey, we are changing the way we export our files on ui-library. Since this is a new component, could you please refeer to `ItemBigData` to see the template?
TLDR;
- index will export only the component (not a styled-component)
- styled-component is declared "inside" the component
- you can use an external file .style if you will
- stories also changed, you can write it on CSF or MDX

src/grip/specifications/grip.md
```diff
@@ -0,0 +1,31 @@
...
+  }}
+  onSlideDown={() => {
+    /* ... */
+  }}
+>
+  <TheVoice>What is your route?</TheVoice>
+</Grip>
+```
+
+## Props
```
gabrielagabriel, line 25: If you write the story on .MDX, it will have properties for free ;)
Please refeer to `ItemBigData` component for an exemple or any of my open PRs

 https://github.com/blablacar/ui-library/pulls/gabrielagabriel

src/grip/story.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
```
gabrielagabriel, line 1: With MDX you can merge story and md files

src/grip/Grip.tsx
```diff
@@ -0,0 +1,62 @@
...
+  // difference with starting point
+  useEffect(() => {
+    const delegatedTouchEndListener = (e: TouchEvent) =>
+      touchEndListener(e.changedTouches.item(0).clientY, fingerYPosition, setFingerYPosition, props)
+    window.addEventListener('touchend', delegatedTouchEndListener)
+
+    return function cleanup() {
+      window.removeEventListener('touchend', delegatedTouchEndListener)
+    }
+  })
```
simonrelet, line 48: As there are no dependencies to this effect, it will be run on each update of the component, which means the event listener will be removed and added on each update.

Can't `onTouchEnd` prop be used on the div element?

PerrineBoissieres: I'd like to allow the user to drag his finger outside the panel and still have the gesture be registered properly, so sadly I have to listen to the whole page...
I could revert the component back to a class component to avoid this, unless there is a way to do it with hooks?

src/grip/index.tsx
PerrineBoissieres: Thanks for the tips, I'll have a look :)

src/grip/GripHandle.tsx
```diff
@@ -0,0 +1,20 @@
+import styled from 'styled-components'
+
+import { color, space } from '../_utils/branding'
+
+export const GripHandle = styled.div`
+  & {
```
gabrielagabriel, line 6: It's a brand new component, you don't need to declare higher specificity, so you can remove the `&`

src/grip/specifications/grip.md
```diff
@@ -0,0 +1,31 @@
...
+
+<!-- STORY -->
+
+## Usage
+
+`TextComponents` are exported individually.
+
+```js
+import Grip from '@blablacar/ui-library/build/grip'
+;<Grip
```
gabrielagabriel, line 13: I have the same issue on formating :(
Remove the `;` before `<Grip...`

PerrineBoissieres: Thanks! Done ;)

src/grip/GripHandle.tsx
PerrineBoissieres: Done ;)

src/grip/Grip.tsx
PerrineBoissieres: Ok, I found a solution: having an empty array as a second argument for `useEffect` limits the call to this hook to the mount and unmount of the component, solving the issue :)

Source: https://stackoverflow.com/questions/53464595/how-to-use-componentwillmount-in-react-hooks

src/grip/story.tsx
PerrineBoissieres: Thanks!

src/grip/specifications/grip.md
PerrineBoissieres: Thanks!

src/grip/index.tsx
PerrineBoissieres: Updated the PR with the new docs format and removed the unused style from index :)

PerrineBoissieres:
> Since it's a brand new component, would be interesting to apply the new syntax. Don't hesitate if you have questions

Thanks a lot for the examples! I updated the PR to match the new guidelines, it's pretty awesome 👍

src/grip/Grip.tsx
simonrelet: Oh ok then maybe you can:

1. Use a Ref instead of a state as `fingerYPosition` value is only used in a futur `touchend` event.
  ```tsx
  const fingerYPosition = useRef<number | null>(null)
  // Reset it with:
  fingerYPosition.current = null
  ```

2. Move the disabled check at the top of the effect function, so you only add the event listener when you need it.
  ```tsx
  useEffect(() => {
    if (!disabled) {
      window.addEventListener()
      return () => {
        window.removeEventListener()
      }
    }
  }, [disabled]) // <= Still need to be a dependency
  ```

3. Put `onSlideUp` and `onSlideDown` in Refs so they can be removed from the dependencies while being defined inline.
  ```tsx
  const onSlideUpRef = useRef(onSlideUp)
  useEffect(() => (onSlideUpRef.current = onSlideUp)) // <= Need to be in an effect to make sure we only update the Ref when the component has been updated.

  onSlideUpRef.current() // <= Calls the "latest" reference of `onSlideUp`.

  <Grip
    onSlideUp={() => { /* ... */ }} // <= Won't remove/add the global event listener on each render.
  />
  ```

src/grip/Grip.story.mdx
```diff
@@ -0,0 +1,44 @@
...
+import { TheVoice } from '../theVoice'
+import { Grip } from './index'
+
+<Meta title="Widgets|Grip" />
+
+# Grip
+
+<Preview>
+  <Story name="Default">
+    <Grip onSlideUp={action('Slide Up')} onSlideDown={action('Slide Down')}>
```
grosbouddha, line 15: Not a biig fan of the naming, If we put actual content in the Grip, IMO it's not a grip anymore but more a container being decorated with a Grip and the Grip naming is misleading.

PerrineBoissieres: Yeah, it's a bit of the stretch, but it comes from the way I've decided to split up the responsibilities of the components that I'll build for the final screen.
From a technical point of view, I wanted the `Grip` component to carry the responsibility of handling the gestures, and since the specs mention that a larger area than the small `Grip` is responsive to finger gestures (in my case, the whole routes list) I needed the `Grip` to be able to have children to expand the area that is "grippable".
I could have done something like:
```
<GripContainer>
  <Grip />
  {children}
</GripContainer>
```
But I preferred to have the main component that we are going to use in our code `Grip` for consistency with the specs.

PR #797, opened 2020-07-20
add SearchForm props initialFrom initialTo disabledFrom disabledTo

Tuxosaurus:
## Description

Add `initialFrom` and `initialTo` to prefill the form. The values are placed instead of the placeholder.
Add `disabledFrom` and `disabledTo` to disable From and To inputs. It is not clear if we'll use it but it's a safe move in case we want to protect the initial values.

With `initialFrom` = "Grenoble" (and `disabledTo` set to true but it doesn't show)
Screen Shot 2020-07-20 at 17 39 10

## Things to consider

I didn't change the story to MDX yet. The original story has a case for each component of the Search Form so I'd rather separate the update in another PR.

## How it was tested

Tested locally on Firefox within the main project.

### DIFF
--- a/src/searchForm/SearchForm.tsx
+++ b/src/searchForm/SearchForm.tsx
@@ -29,6 +29,10 @@ import { TRANSITION_SECTION_CLASS_NAME, transitionSectionTimeout } from './trans
 export interface SearchFormProps {
   className?: string
   onSubmit: (formValues: SearchFormValues) => void
+  disabledFrom?: boolean
+  disabledTo?: boolean
+  initialFrom?: string
+  initialTo?: string
   autocompleteFromPlaceholder: AutoCompleteProps['placeholder']
   autocompleteToPlaceholder: AutoCompleteProps['placeholder']
   renderAutocompleteFrom: AutoCompleteOverlayProps['renderAutocompleteComponent']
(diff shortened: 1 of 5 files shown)


src/searchForm/SearchForm.tsx
```diff
@@ -212,6 +220,21 @@ export const SearchForm = ({
     formValues[SearchFormElements.AUTOCOMPLETE_FROM] != null ||
     formValues[SearchFormElements.AUTOCOMPLETE_TO] != null
 
+  const getPlaceholderText = (
```
simonrelet, line 223: It looks like this function can be static.

src/searchForm/SearchForm.tsx
```diff
@@ -248,7 +272,11 @@ export const SearchForm = ({
                   { 'kirk-search-placeholder': !autocompleteFromValue },
                 ])}
               >
-                {autocompleteFromValue?.item.label || autocompleteFromPlaceholder}
+                {getPlaceholderText(
```
simonrelet, line 275: Can the `initial(From|To)` values be displayed in black as if the user selected it?
I didn't see it at first in the screenshot in the description because it had the same color as a placeholder.

src/searchForm/SearchForm.tsx
Tuxosaurus: Not sure what you have in mind

src/searchForm/SearchForm.tsx
Tuxosaurus: You're right, it's probably better to leave it "normal" in the absence of product feedback for now. 
I removed the placeholder class when an initial value is present.

src/searchForm/SearchForm.tsx
simonrelet: You can move it out of the component as it doesn't depend on props so you don't recreate it on each render 😉.

Tuxosaurus: Oh sorry, sure I moved it

PR #801, opened 2020-07-21
Migrate Loader to new storybook

grosbouddha:
## Description
Migrate loader to new stoybook format.

## How it was tested

Locally


src/loader/Loader.story.mdx
```diff
@@ -0,0 +1,41 @@
...
+  <Story name="Default">
+    <Loader
+      layoutMode={LoaderLayoutMode.BLOCK}
+      size={48}
+      done={false   }
+    />
+  </Story>
+</Preview>
+
+# ** Legacy inline attribute loader (use layoutMode LoaderLayoutMode.INLINE instead) **
```
PerrineBoissieres, line 19: Nit: Do we want to keep this?

grosbouddha: Good idea, removed !

PR #817, opened 2020-07-27
Modal Layout for Filters

wakooka:
## Description

Layout for Modal that will be used for Filters feature in SPA

## What has been done

- Used styled components `ModalBody` `ModalFooter` `ModalFog` 
- Kept current modal retro-compatible
- Removed `!important` from `normalizeHorizontally`, I checked and it has no impact.

## Things to consider

- Couldn't really unit test this, I'd like to move the modal to RTL in a next PR
- This new layout should be moved out of the `Modal` and should be a Page layout, but right now we need this to progress on the filtres feature

## How it was tested

Storybook locally
Kapture 2020-07-23 at 15 08 21

### DIFF
--- a/src/modal/Modal.tsx
+++ b/src/modal/Modal.tsx
@@ -35,6 +35,8 @@ export type ModalProps = A11yProps &
     closeButtonTitle?: string
     forwardedRef?: Ref<HTMLDivElement>
     noHorizontalSpacing?: boolean
+    layoutModeEnabled?: boolean
+    isLoading?: boolean
   }>
 
 export class Modal extends Component<ModalProps> {
@@ -50,6 +52,11 @@ export class Modal extends Component<ModalProps> {
     size: ModalSize.MEDIUM,
     displayDimmer: true,
     forwardedRef: null,
+    isLoading: false,
+  }
+
+  state = {
+    showFooterBorder: true,
   }
 
   constructor(props: ModalProps) {
@@ -140,6 +147,14 @@ export class Modal extends Component<ModalProps> {
     this.focusTrap.activate()
   }
 
+  onScroll = (e: React.UIEvent<HTMLElement>): void => {
+    const element = e.currentTarget
+    const isAtBottom = element.clientHeight + element.scrollTop >= element.scrollHeight
+    this.setState({
+      showFooterBorder: !isAtBottom,
+    })
+  }
+
   render() {
     const baseClassName = 'kirk-modal'
     const a11yAttrs = pickA11yProps<ModalProps>(this.props)
@@ -171,6 +186,7 @@ export class Modal extends Component<ModalProps> {
                 {...a11yAttrs}
                 role="dialog"
                 aria-modal="true"
+                onScroll={this.onScroll}
               >
                 <div className={`${baseClassName}-dialog`}>
                   {this.props.displayCloseButton && (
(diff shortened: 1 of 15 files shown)


src/modal/Modal.tsx
```diff
@@ -183,7 +199,14 @@ export class Modal extends Component<ModalProps> {
                       <CrossIcon size="18" iconColor={color.blue} />
                     </Button>
                   )}
-                  <div className={`${baseClassName}-body`}>{this.props.children}</div>
+                  <div
+                    className={cc([
+                      `${baseClassName}-body`,
+                      this.state.showFooterBorder ? `${baseClassName}-modalFooterBorder` : '',
```
simonrelet, line 205: I'm not sure to understand where this class (`${baseClassName}-modalFooterBorder`) is defined.

src/modal/story.tsx
```diff
@@ -47,12 +52,48 @@ class ModalOpener extends Component<ModalProps> {
...
-          </div>
+        <Modal
+          {...this.props}
+          onClose={this.closeModal2}
+          isOpen={this.state.modalOpen2}
+          noHorizontalSpacing
+          layoutModeEnabled
+        >
+          <ModalBody>
+            <ModalFog aria-hidden="true" isLoading={this.props.isLoading} />
```
simonrelet, line 63: Is there a way to prevent the body from scrolling?

src/modal/index.tsx
```diff
@@ -103,6 +105,45 @@ const StyledModal = styled(Modal)`
...
+
+const ModalFooter = styled.div`
+  position: -webkit-sticky;
+  position: sticky;
+  bottom: 0;
+  background-color: ${color.white};
+  flex-shrink: 0;
+  z-index: 2;
+
+  .kirk-modal-modalFooterBorder & {
```
simonrelet, line 123: How will this class be toggled?

src/modal/story.tsx
wakooka: Do we need to prevent the scroll? I think we discussed it but I wasn't sure and there's nothing about this in the specifications

src/modal/index.tsx
wakooka: This class is toggled based on the `showFooterBorder` state in the Modal. Without a bigger refacto (and probably breaking changes) that's how far I could get.
1. Added the `onScroll` event on Modal scrollable content
2. When reaching the bottom we update the `showFooterBorder` state which activate the `kirk-modal-modalFooterBorder` class
3. `ModalFooter` then applies its border based on `kirk-modal-modalFooterBorder`

src/modal/Modal.tsx
wakooka: Explained here: https://github.com/blablacar/ui-library/pull/817#discussion_r460950021

src/modal/story.tsx
simonrelet: We discussed it with Jonas but the specs were never updated, so I'm fine with keeping it like this 👍.

src/modal/index.tsx
simonrelet: Oh! I didn't see the `&` at the end of the selector. This changes everything 👍

src/modal/story.tsx
wakooka: Ok I'll have a look if I can make this work in another PR.

PR #821, opened 2020-07-28
[BBC-9106] Title.MDX

emaghe:
## Description

Migrated storybook for Title component

## What has been done

- migrated storybook
- updated tests to use RTL

## Things to consider

To avoid breaking changes and the need to change code in Kairos too I declared headingLevel both with number and strings

## How it was tested

Locally

Screenshot 2020-07-28 at 16 49 36


src/title/Title.story.mdx
```diff
@@ -0,0 +1,43 @@
...
+  2: '2',
+  3: '3',
+  4: '4',
+  5: '5',
+  6: '6',
+}
+
+<Preview>
+  <Story name="Default">
+    <BaseSection>
```
gabrielagabriel, line 23: No need to have section here

src/title/Title.style.tsx
```diff
@@ -0,0 +1,12 @@
+import styled from 'styled-components'
+
+import { color, font, fontWeight } from '../_utils/branding'
+
+export const StyledTitle = styled.div`
+  & {
```
gabrielagabriel, line 6: no need to add specifity. You can remove this `&` here

src/title/Title.tsx
```diff
@@ -1,25 +1,32 @@
-import React, { Fragment, ReactNode } from 'react'
+import React, { ReactNode } from 'react'
 import cc from 'classcat'
 
+import { StyledTitle } from './Title.style'
+
 export interface TitleProps {
```
gabrielagabriel, line 6: also convert to types? @simonrelet

src/title/Title.tsx
```diff
@@ -1,25 +1,32 @@
...
-    `h${headingLevel}`,
-    {
-      id,
-      className: cc(['kirk-title', className]),
-    },
-    <Fragment>{children}</Fragment>,
+
+  return (
+    <StyledTitle
+      as={HEADING_LEVEL_MAPPING[headingLevelNumber - 1]}
```
gabrielagabriel, line 25: you could simplify by interpolating `h$headingLevel`

src/title/Title.unit.tsx
```diff
@@ -1,38 +1,61 @@
...
-    expect(wrapper.hasClass('kirk-title')).toBe(true)
-    expect(wrapper.text()).toContain(title)
-    expect(wrapper.name()).toBe('h1')
+    const props = createProps({
+      children: title,
+    })
+
+    render(<Title {...props} />)
+
+    expect(screen.getByRole('heading').className.toLowerCase()).toContain('kirk-title')
```
gabrielagabriel, line 25: IMHO we don't need to have a test to define implementation details, thus testing if this class is there is useless

src/title/Title.tsx
emaghe: Unfortunately interpolating is not working as expected with the as polymorphic and this workaround is needed.
Happy to give you more details on this where @simonrelet helped me (thanks!)

src/title/Title.tsx
simonrelet: Yes please 🙏

src/title/Title.tsx
simonrelet: ``` `h${headingLevel}` ``` cannot be typed as `'h1' | 'h2' | ... | 'h6'`, it has to be explicitly `'h1'`, `'h2'`, ..., `'h6'`.

@emaghe, I would add a comment explaining the reason of this mapping.

src/title/Title.style.tsx
```diff
@@ -0,0 +1,12 @@
+import styled from 'styled-components'
+
+import { color, font, fontWeight } from '../_utils/branding'
+
+export const StyledTitle = styled.div`
+   {
```
simonrelet, line 6: You can also remove the `{}`:

```ts
export const StyledTitle = styled.div`
  color: ${color.midnightGreen};
  // ...
`
```

src/title/Title.tsx
```diff
@@ -1,25 +1,32 @@
...
 import cc from 'classcat'
 
+import { StyledTitle } from './Title.style'
+
 export interface TitleProps {
   readonly id?: string
   readonly className?: string
   readonly children: ReactNode
-  readonly headingLevel?: number | string
+  readonly headingLevel?: 1 | 2 | 3 | 4 | 5 | 6 | '1' | '2' | '3' | '4' | '5' | '6'
```
simonrelet, line 10: nit: I would add a comment explaining that the string levels are there for retro-compatibility.

src/title/Title.unit.tsx
```diff
@@ -1,38 +1,60 @@
...
-    expect(wrapper.hasClass('kirk-title')).toBe(true)
-    expect(wrapper.text()).toContain(title)
-    expect(wrapper.name()).toBe('h1')
+    const props = createProps({
+      children: title,
+    })
+
+    render(<Title {...props} />)
+
+    expect(screen.getByRole('heading')).toHaveTextContent(title)
```
simonrelet, line 25: nit: You can use `props.title` instead of `title` and remove the local variable 👌.

src/title/Title.unit.tsx
```diff
@@ -1,38 +1,60 @@
...
+    expect(screen.getByRole('heading').nodeName.toLowerCase()).toBe('h1')
```
simonrelet, line 26: nit: I would move this assertion to a dedicated test: "should render a `h1` when no `headingLevel` is specified"

src/title/Title.unit.tsx
```diff
@@ -1,38 +1,60 @@
...
-    expect(wrapper.prop('id')).toBe('my-id')
+    const title = 'blabla'
+    const id = 'my-id'
+    const props = createProps({
+      children: title,
+      id,
+    })
+
+    render(<Title {...props} />)
+    expect(screen.getByRole('heading').id).toBe(id)
```
simonrelet, line 58: nit: `element.id` is valid but uncommon, you can use `.haveAttribute()` to be iso with other attributes:

```ts
expect(screen.getByRole('heading')).toHaveAttribute('id', id)
```

PR #822, opened 2020-07-28
[BBC-9100] Fix storybook BaseSection MDX

emaghe:
## Description

Quick improvement in the MDX file

## What has been done

Just added the <BaseSection> to make sure the storybook behaviour is exactly the same as before the migration

## How it was tested

Locally and comparing with the old version currently in production


emaghe:
Ok to close this PR and do not add BaseSection

PR #826, opened 2020-07-29
Add support for adyen CSE library in inputs

florianeyzat:
## Description

Adding a new attribute for inputs, for support of Adyen's payment interface

## How it was tested

Tested locally with kairos.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **[UPDATE]** Added `adyenEncryptedName` props on `TextField`
 [...]
 
 # v38.1.0 (28/07/2020)
--- a/src/textField/TextField.tsx
+++ b/src/textField/TextField.tsx
@@ -68,6 +68,7 @@ export interface TextFieldProps extends CommonFormFields {
   format?: (value: string, previousValue: string) => string
   focusBorder?: boolean
   loader?: JSX.Element
+  adyenEncryptedName?: string
 }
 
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **[UPDATE]** Added `adyenEncryptedName` props on `TextField`
```
grosbouddha, line 3: This is not some very generic reusable API.
Ideally, I would prefer to have some extraInputAttributes API that can be used for adyen but as well any other weird cases in the future.
I can live with that too, your call.

wakooka: Yes we discussed this with @florianeyzat yesterday. I was thinking of having a sort of `extraInputAttributes` typed `InputHTMLAttributes<HTMLInputElement>` that we can spread safely on the input, so we totally control the DOM ouput of the `input` from the outside.
```
 <input {...extraInputAttributes} />
```

simonrelet: Just adding my 2 cents: I agree with @wakooka, but, as `TextField` is basically a wrapper around the native input, I think it should _inherit_ the native inputs props and spread them on the native input.

We wouldn't have to do these PR anymore and this is the reason I added the "Use rest/spread operator" in the propositions to ease the Kirk/Kairos workflow.

florianeyzat: Hey guys, yes we discussed it with @wakooka, I'm working on a suitable solution here, a PR will come really soon, and I'm closing this one in the meantime

PR #827, opened 2020-07-30
Fixed highlight section title color

wakooka:
## Description

Trying to fix title color in `highlightSection` but I can not override `SubHeader` using styled-component
@gabrielagabriel ?

### DIFF
--- a/src/layout/section/highlightSection/index.tsx
+++ b/src/layout/section/highlightSection/index.tsx
@@ -1,6 +1,7 @@
 import styled from 'styled-components'
 
 import { color, font, responsiveBreakpoints, space } from '../../../_utils/branding'
+import { SubHeader } from '../../../subHeader'
 import { HighlightSection } from './highlightSection'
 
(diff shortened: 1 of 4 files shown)


src/layout/section/highlightSection/index.tsx
```diff
@@ -12,18 +13,16 @@ const StyledHighlightSection = styled(HighlightSection)`
...
   }
+`
 
-  & .kirk-subheader {
-    color: ${color.white};
-    font-size: ${font.xl.size};
-    padding-top: ${space.xl};
-    padding-bottom: ${space.xl};
-  }
+export const StyledHighlightSectionSubHeader = styled(SubHeader)`
```
simonrelet, line 18: I think you need to add a `&` to add specificity because `SubHeader` has one.

simonrelet: I also had issues with it and never understood why it was here.

wakooka: I already tried, but still not working :(

PR #833, opened 2020-08-04
[BBC-9062] PushInfo.MDX

Tuxosaurus:
I had to remove the icon color in the story as it caused "too much recursion".
Also moved the animationDuration & animationDelay variables to the style file.

Screen Shot 2020-08-04 at 14 51 52
Screen Shot 2020-08-04 at 14 52 16


src/pushInfo/PushInfo.story.mdx
```diff
@@ -0,0 +1,42 @@
+import { Meta, Story, Preview, Props } from '@storybook/addon-docs/blocks'
+
+import { ProximityIcon } from '../icon/proximityIcon'
+import { PushInfo } from './index'
+
+<Meta title="Widgets|PushInfo" />
```
gabrielagabriel, line 6: What do you think to create an "Alert" category?

src/pushInfo/PushInfo.story.mdx
```diff
@@ -0,0 +1,42 @@
...
+    <PushInfo
+      icon={<ProximityIcon title="" />}
+      headline="This is the headline!"
+      content="But we can also give you more explanations here"
+    />
+  </Story>
+</Preview>
+
+## Specifications
+
```
gabrielagabriel, line 34: Maybe say that we should add some colors to the Icon?

src/pushInfo/PushInfo.story.mdx
Tuxosaurus: What else would you put in there?

src/pushInfo/PushInfo.style.tsx
```diff
@@ -0,0 +1,58 @@
+import styled from 'styled-components'
+
+import { color, font, fontWeight, radius, space } from '../_utils/branding'
+
+export const animationDuration = 700
```
gabrielagabriel, line 5: IMO it is a good practice to declare animations with style, since it's more a UI thing.. 
So I'm ok for declaring here.

src/pushInfo/PushInfo.story.mdx
gabrielagabriel: "Alert" elements like `Snackbar`

src/pushInfo/PushInfo.story.mdx
Tuxosaurus: I don't think it's necessary, it's part of the icon, not the PushInfo.
I added/adapted some of the specs from the designers.

src/pushInfo/PushInfo.story.mdx
Tuxosaurus: I'm not sure it qualifies as an alert, but either way, I'd rather to keep the category out of the MDX migration for now.

Tuxosaurus:
Update: added changelog + tested for regression on the main app tests with canary release 0.0.0-69e01640, it's fine.

PR #837, opened 2020-08-04
[BBC-] ItemsList.MDX

PerrineBoissieres:


src/itemsList/ItemsList.style.tsx
```diff
@@ -0,0 +1,9 @@
+import styled from 'styled-components'
+
+export const StyledItemsList = styled.ul`
+  & {
```
gabrielagabriel, line 4: you can remove & {}

stories/index.tsx
```diff
@@ -61,7 +61,6 @@ import '../src/itemBigData/story'
 import '../src/itemEditableInfo/story'
 import '../src/itemInfo/story'
 import '../src/itemRadio/story'
```
gabrielagabriel, line 63: we don't need to touch this file. I'll remove it

PR #839, opened 2020-08-04
[BBC-9056] Message.MDX

florianeyzat:


src/message/Message.story.mdx
```diff
@@ -0,0 +1,89 @@
...
+  <Story name="Default">
+    <Message active={false}>{text('text', 'A simple message')}</Message>
+  </Story>
+</Preview>
+
+## Conversation
+
+<Preview>
+  <Story name="Conversation">
+    <Fragment>
```
gabrielagabriel, line 23: Lots of messages, it would be better to separate on "variations", Check `ButtonGroup` to an example

src/message/index.tsx
```diff
@@ -1,69 +1,2 @@
...
-  }
-  &.kirk-active .kirk-message-annotation {
-    padding-right: ${space.l};
-  }
-`
-
-export { MessageProps } from './Message'
-export { StyledMessage as Message }
-export default StyledMessage
+export { Message } from './Message'
```
gabrielagabriel, line 1: Missing `MessageProps` export

PR #841, opened 2020-08-04
TripCard.mdx

wakooka:
## Description
localhost_4000__path=_story_widgets-tripcard--default-story

## How it was tested

Tested locally in storybook


src/tripCard/index.tsx
```diff
@@ -1,160 +1,2 @@
...
-  & .kirk-tripCard-flags .kirk-icon {
-    display: inline-block;
-    margin-left: ${space.m};
-  }
-`
-
-export { TripCardProps, User } from './TripCard'
-export { StyledTripCard as TripCard }
-export default StyledTripCard
+export { TripCard, TripCardProps } from './TripCard.style'
```
mcellier, line 1: You should import `TripCard.style.tsx` only in `TripCard.tsx`. 
Look at the avatar component for an example

mcellier (commented):
Missing changelog update

src/tripCard/index.tsx
wakooka: Makes sense for Avatar because it's wrapped with the `StyledAvatar` but `TripCard` is not. It's wrapped with `Card` component maybe I should migrate `Card` component first.

wakooka:
> Missing changelog update

Not sure, since it's an internal change I think we said that we don't need to update the changelog. Poke @gabrielagabriel

src/tripCard/index.tsx
wakooka: Actually `Card` is an internal component so there's nothing to migrate.

wakooka: I updated the `.style` so now it's overriding `internals/card` instead of `TripCard` component. PTAL @mcellier

mcellier (approved):

src/tripCard/index.tsx
mcellier: > > Missing changelog update
> 
> Not sure, since it's an internal change I think we said that we don't need to update the changelog. Poke @gabrielagabriel

I've seen several in the changelog already

gabrielagabriel: We have decided to update the changelog since it may break some tests on the App.

src/tripCard/TripCard.story.mdx
```diff
@@ -0,0 +1,94 @@
...
+Use backgroundless components to represent card’s content.
+
+- ✅ Always clickable.
+
+## Usage
+
+```jsx
+import { TripCard } from '@blablacar/ui-library/build/tripCard'
+
+<TripCard />
```
gabrielagabriel, line 91: no props exemples?

gabrielagabriel (approved):
🎉

src/tripCard/index.tsx
wakooka: I'll update the changelog. Thanks

PR #842, opened 2020-08-04
Add general support for input attribute

florianeyzat:
## Description

The goal is to be able to pass any html input attribute in the TextField component, that will be passed to its child input.

## Things to consider

Some attributes can be removed from the props in the future, to be cleaner (not done here to avoid breaking changes for now, can be done in multiple times)

## How it was tested

Locally with generated TextFields


src/textField/TextField.tsx
```diff
@@ -68,16 +68,7 @@ export interface TextFieldProps extends CommonFormFields {
...
-
-interface FormAttributes extends CommonFormFields {
-  value: string
-  ['aria-invalid']?: 'true' | 'false'
-  ['aria-labelledby']?: string
-  inputMode?: inputModes
-  pattern?: string
-  ref?: (input: textfield) => void
-  onChange?: (event: React.ChangeEvent<textfield>) => void
+  inputAttributes: InputHTMLAttributes<HTMLInputElement>
```
wakooka, line 71: @simonrelet was suggesting to directly extend `TextFieldProps` to `InputHTMLAttributes<HTMLInputElement>` wdyt?
I'll leave it up to you, both approaches are fine to me. 
On one hand I like to differentiate what we output as DOM attributes from the input by expliciting a prop `inputAttributes`.
And on the other hand you can have a more declarative component when using it e.g `<TextField pattern="" inputMode="" />`

florianeyzat: Yes I thought of doing that at first, but then I thought it could be a bit confusing to use it as an usual input directly, since it has more possible attributes, and does not generate only an input in the DOM, I mean I found that a bit ambiguous to do it like that

gabrielagabriel: I vote for @simonrelet  approach. It followings the same approach used for a11y elements and aria props, so we should keep consistent. Also, we should avoid "props" that doesn't add value do the API itself and should be part of the intrinsic intelligence of the component itself

PR #849, opened 2020-08-11
Refactor Avatar tests with RTL

simonrelet:
## Description

Rewrite Avatar's tests with RTL.

## Things to consider

The component is not fully accessible which cause some tests to depend on tag names. 

## How it was tested

- Unit tests

### DIFF
--- a/src/avatar/Avatar.unit.tsx
+++ b/src/avatar/Avatar.unit.tsx
@@ -1,28 +1,57 @@
 import React from 'react'
-import { shallow } from 'enzyme'
 
-import { Badge } from '../badge'
-import { Avatar } from './index'
+import { render, screen } from '@testing-library/react'
 
-it('Should have the proper default state.', () => {
-  const avatar = shallow(<Avatar image="//placehold.it/80x80" alt="my alternative txt" />)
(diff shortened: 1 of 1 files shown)


src/avatar/Avatar.unit.tsx
```diff
@@ -1,28 +1,57 @@
...
+    const props = createProps({ isIdChecked: true })
+    render(<Avatar {...props} />)
+    // There are no accessible way to check the badge is rendered.
+    expect(document.querySelector('svg')).toBeInTheDocument()
+  })
+
+  it('should render a small avatar when `isSmall` is `true`', () => {
+    const props = createProps({ isSmall: true })
+    const { container } = render(<Avatar {...props} />)
+    expect(container.firstChild).toHaveClass('kirk-avatar--small')
```
gabrielagabriel, line 43: Cool to have 100 of tests covered, but I'm afraid that on the next refactoring we may change these tests since classes would not be available

simonrelet: Probably, but the issue is the lack of accessibility of the component ☹️

PR #853, opened 2020-08-25
Bbc 9604 fix a11y item editable info button

nickoola:
## What has been done
Fix keyboard navigation:
- Use `<button>` instead of `<div role="button">` to make it focusable
- Use the FocusVisibleProvider to display the focus ring

## How it was tested
Locally:
https://www.loom.com/share/7864c97601ad422a83e8768e80ca5d26

### DIFF
--- a/src/itemEditableInfo/ItemEditableInfo.story.mdx
+++ b/src/itemEditableInfo/ItemEditableInfo.story.mdx
@@ -1,21 +1,39 @@
 import { Meta, Story, Preview, Props } from '@storybook/addon-docs/blocks'
 
 import { ItemEditableInfo } from './index'
+import { FocusVisibleProvider } from '../_utils/focusVisibleProvider'
 
 <Meta title="Items|ItemEditableInfo" />
 
 # ItemEditableInfo
 
+### Link
+
 <Preview>
-  <Story name="Default">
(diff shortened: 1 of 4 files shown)


src/itemEditableInfo/ItemEditableInfo.story.mdx
```diff
@@ -1,29 +1,47 @@
 import { Meta, Story, Preview, Props } from '@storybook/addon-docs/blocks'
 
 import { ItemEditableInfo } from './index'
+import { FocusVisibleProvider } from '../_utils/focusVisibleProvider'
 
 <Meta title="Items|ItemEditableInfo" />
 
-# ItemEditableInfo
+# ItemEditableInfo link
```
Tuxosaurus, line 8: I think you want to have a main title "ItemEditableInfo" and "Link" as a subtitle

```
# ItemEditableInfo

### link
```

Tuxosaurus (approved):

grosbouddha (approved):
FYI, styling < button> is sometimes more buggy than a simple < div>, in particular when you start using flex styling. Hopefully, we won't have issues with this change... otherwise, just use a < div tabindex='0' role='button'>.

PR #856, opened 2020-09-01
[BBC-9714] Fix MediaContentSection display

mcellier:
## Description

Noticed a display issue on BBC homepage on mobile device. 
What it looks like:
Capture d’écran 2020-09-01 à 14 01 08
What it should looks like:
Capture d’écran 2020-09-01 à 14 01 18

## What has been done

Issue was introduced via 9c4dffecad14de4bc51f54461d0f4fd5222b4cfd. `SubHeader` component has default margins, but it has been replaced by a `TextDisplay1`.
Just wrapped the component to add some style.

## Things to consider

Snapshots will have to be updated

## How it was tested

Storybook + locally on kairos

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,12 @@
 # Unreleased
 
+- **[FIX]** Fixed `MediaContentSection` margins on the title
 [...]
 
 # v40.0.1 (28/08/2020)
 
 - **[FIX]** Fixed `ItemCheckbox` a11y focus ring display for keyboard navigation
-- **[FIX]** Fixed `ItemEditableInfo` a11y keyboard navigation by Providing a focus ring 
+- **[FIX]** Fixed `ItemEditableInfo` a11y keyboard navigation by Providing a focus ring
 
 # v40.0.0 (24/08/2020)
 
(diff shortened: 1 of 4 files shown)


src/layout/section/mediaContentSection/index.tsx
```diff
@@ -52,6 +52,10 @@ const StyledMediaContentSection = styled(MediaContentSection)`
     & .kirk-media-content-wrapper {
       align-items: center;
     }
+
+    & .kirk-media-content-title {
```
gabrielagabriel, line 56: Instead of using classes why not use **styled-components** redefinition?
```
const SubHeaderTitle = styled(TextDisplay1)`
margin: ${space.xl} 0 ${space.m};
`
<SubHeaderTitle> {title} </SubHeaderTitle>
```

gabrielagabriel (approved):

wakooka (approved):

src/layout/section/mediaContentSection/index.tsx
mcellier: What do you think of the following?

```
${TextDisplay1} {
  margin: ${space.xl} 0 ${space.m};
}
```

Using your solution (if I undestand correcly), I'll have to create another component directly inside the `MediaContentSection` ?

gabrielagabriel: The first solution is the official styled-components sytaxe. Yes `SubHeaderTitle` would be a component inside `SubHeader`

mcellier: Updated the fix. Can you take another look?

gabrielagabriel (approved):
🎉 Cool !

PR #864, opened 2020-09-04
[BBC-9071] MDX Rating

mcellier:
## Description

Migrate `Rating` component to MDX

## How it was tested

Comparison between prod storybook. This component doesn't seems to be used anywhere?


gabrielagabriel (approved):
This component should be used on `TripCard`, I think...

PR #865, opened 2020-09-07
Update text prop to allow jsx element in EmptyState

Tuxosaurus:
Update EmptyState so that "text" prop can also be a JSX element. It still defaults to a "Title" component if a string is provided.

Context: for SEO purposes, we want to be able to add more info in the title but with different styles (else you might end up with the inclusion of a very long city name written as a big title).

Tested locally on Firefox, with a canary deployment to test in the main app.
Screen Shot 2020-09-07 at 16 17 33

### DIFF
--- a/src/emptyState/EmptyState.tsx
+++ b/src/emptyState/EmptyState.tsx
@@ -7,14 +7,23 @@ import { StyledEmptyState } from './EmptyState.style'
 export type EmptyStateProps = Readonly<{
   className?: string
   image: string
-  text: string
+  text: string | JSX.Element
   button?: JSX.Element
 }>
 
-export const EmptyState = ({ className, image, text, button }: EmptyStateProps) => (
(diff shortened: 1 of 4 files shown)


src/emptyState/EmptyState.tsx
```diff
@@ -7,14 +7,23 @@ import { StyledEmptyState } from './EmptyState.style'
 export type EmptyStateProps = Readonly<{
   className?: string
   image: string
-  text: string
+  text: string | JSX.Element
```
simonrelet, line 10: You can use `text: React.ReactNode` which includes both `string` and React elements but is more flexible

simonrelet (approved):

src/emptyState/EmptyState.tsx
Tuxosaurus: Unfortunately the story page sends me a "too much recursion" error if I use `React.ReactNode`

PR #868, opened 2020-09-09
Increase line-height in tabs to support all diacritics

simonrelet:
## Description

Make sure all diacritics are properly displayed in the tabs.

## What has been done

Set the line height to `normal`: it's the default value.

**Before**

Screenshot 2020-09-09 at 11 18 55

**After**

Screenshot 2020-09-09 at 11 19 15

## Things to consider

For accessibility, the line-height should be at least 1.5, `normal` is roughly around 1.2.

## How it was tested

- Storybook

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
-[...]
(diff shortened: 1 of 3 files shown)


wakooka (approved):
Update changelog. Thanks for the fix.

PR #869, opened 2020-09-09
Added focus state on ItemCheckbox

wakooka:
## Description

Added focus state for item checkbox + various improvements.

## What has been done

- Use `useFocusVisible` hook for removing focus border when clicking on radios or checkboxes
- Migrated both `itemRadio` and `itemCheckbox` as functional component
- Migrated both `itemRadio` and `itemCheckbox` to RTL
- Factorised styled between `itemRadio` and `itemChecbox`

## How it was tested

Locally in storybook, currently testing in SPA where we use checkboxes and radios


src/itemCheckbox/ItemCheckbox.unit.tsx
```diff
@@ -1,76 +1,51 @@
...
+  it('Should have a label around the input', () => {
+    const props = createProps({
+      label: 'This is the field label',
+      labelTitle: '',
+      data: '',
+      dataInfo: '',
     })
+    render(<ItemCheckbox {...props} />)
+    expect(
+      screen.getByLabelText('This is the field label', { selector: 'input[type="checkbox"]' }),
```
simonrelet, line 35: I think this should also work:

```ts
screen.getByRole('checkbox', { name: "This is the field label" })
```
If you want to avoid the `input[type="checkbox"]`

src/itemCheckbox/ItemCheckbox.unit.tsx
```diff
@@ -1,76 +1,51 @@
...
-  })
-  it('Should call the onChange prop with name and value when the input changes (uncheck)', () => {
-    const onChangeMock = jest.fn()
-    const itemCheckbox = shallow(<ItemCheckbox {...defaultProps} onChange={onChangeMock} checked />)
-    itemCheckbox.instance().onChange()
-    expect(onChangeMock).toHaveBeenCalledWith({
-      name: defaultProps.name,
-      value: false,
-    })
+  it('Should display a Loader when the component is in loading status', () => {
```
simonrelet, line 39: This test doesn't assert on the visibility of a loader 🤔

simonrelet (approved):

src/itemRadio/ItemRadio.tsx
```diff
@@ -34,100 +35,73 @@ export type ItemRadioProps = A11yProps &
...
-  }
+      {(!chevron || isLoading) && (
+        <RadioIcon isChecked={checked} isLoading={isLoading} isDisabled={disabled} />
+      )}
+    </Fragment>
+  )
+
+  return (
+    <StyledItemRadioCheckbox
+      className={cc(['kirk-item-radio', className, { focus: focusVisible }])}
```
nickoola, line 85: Does it apply the `focus-visible` class when you navigate with the keyboard?

src/_internals/item/ItemRadioCheckbox.style.tsx
```diff
@@ -0,0 +1,24 @@
...
+  &.focus-visible {
+    outline: ${inputBorderSize.focus} solid ${color.blue};
+  }
+
+  & input {
+    position: absolute;
+    clip: rect(0, 0, 0, 0);
+  }
+
+  &.focus {
```
nickoola, line 21: I'll rename it `.focus-visible` as it's only applied when keyboard nav is used, provided by the `useFocusVisible` and it's named like this in other component and in Kairos.

nickoola (approved):

src/_internals/item/ItemRadioCheckbox.style.tsx
wakooka: Yes as discussed I remove this class, the same rules are already applied on `.focus-visible` above, and I'll change it in components for consistency.

src/_internals/item/ItemRadioCheckbox.style.tsx
```diff
@@ -0,0 +1,24 @@
+import styled from 'styled-components'
+
+import { color, inputBorderSize } from '../../_utils/branding'
+import { Item } from './index'
+
+export const StyledItemRadioCheckbox = styled(Item)`
```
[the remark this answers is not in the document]
wakooka: Not sure, by an (implicit) convention in our codebase we prefix all our components wrapped with `styled` by `Styled*`

src/itemCheckbox/ItemCheckbox.unit.tsx
wakooka: It does work thanks. Much nicer indeed.

src/itemCheckbox/ItemCheckbox.unit.tsx
wakooka: Oops, good catch! I think I just forgot to write this test 😅

wakooka: Updated

src/_internals/item/ItemRadioCheckbox.style.tsx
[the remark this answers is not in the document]
simonrelet: I though we prefixed them only to avoid name conflict with the exported component using the styled component.
I find it weird having an exported `StyledItemRadioCheckbox` from a _ItemRadioCheckbox.style.tsx_ file, it feels like we're exposing implementation details.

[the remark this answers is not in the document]
wakooka: You're right, I renamed it to `ItemRadioCheckbox`

PR #872, opened 2020-09-11
Migrate `TabsSection` component to MDX

mcellier:
## Description

Migrate `TabsSection` component to MDX

## How it was tested

Storybook + search on kairos

PR #881, opened 2020-09-17
Only use knobs in stories

simonrelet:
## Description

When starting storybooks, there are some knobs displayed on all stories as a side effect of one story.

## What has been done

Make sure the knobs are called only in the scope of a story.

**Example with `ButtonGroup` story**

Before | After
------------ | -------------
Screenshot 2020-09-17 at 16 43 39 | Screenshot 2020-09-17 at 16 44 08

## Things to consider

Nothing.

## How it was tested

- Storybook

### DIFF
--- a/src/layout/section/tabsSection/story.tsx
+++ b/src/layout/section/tabsSection/story.tsx
@@ -28,30 +28,32 @@ const panels = [
   </BaseSection>,
 ]
 
-const tabs = {
-  activeTabId: 'tab1',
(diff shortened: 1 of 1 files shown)


gabrielagabriel (approved):
okay.. but how about migrating the story?

wakooka (approved):
Wow. Ok! I don't understand what's going on here 😄

simonrelet:
Fixed by #872.

PR #890, opened 2020-09-21
Propim 480/breadcrumb

Tuxosaurus:
- New Breadcrumb component: a list of links with its spacing and a divider
- Update to HeroSection: add "noMarginBottom" to accomodate the fact that it can be followed directly by the breadcrumb.
_Note: I keep the margin on small devices to make sure there's space between the main action button and the breadcrumb links._

Large device (desktop)
Screen Shot 2020-09-21 at 16 30 28

Small device (mobile)
Screen Shot 2020-09-21 at 16 30 57

Tested locally on Firefox, within ui-library and in the main project using a canary release


src/breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export const Breadcrumb = ({ crumbs }: BreadcrumbProps) => (
+  <Fragment>
+    <StyledBreadcrumb itemScope itemType="https://schema.org/BreadcrumbList">
+      {crumbs.map((crumb, index) => {
+        const position = Number(index + 1)
+
+        return (
+          <li
+            key={`crumb-${crumb.name}-${position}`}
```
wakooka, line 25: Why do you add the `index`, isn't the name unique?

src/breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,55 @@
...
+                <span itemProp="name">{crumb.name}</span>
+              </a>
+
+              {position < crumbs.length && (
+                <span role="separator" className="breadcrumb-separator">
+                  {' > '}
+                </span>
+              )}
+            </TextCaption>
+            <meta itemProp="position" content={`${position}`} />
```
wakooka, line 48:
```suggestion
            <meta itemProp="position" content={position} />

wakooka (commented):
If possible I would remove `noMarginBottom` so that we don't accomodate `HeroSection`. If it's too low I would argue that these components need to be in the design system, while they're not we shouldn't create more & more use cases.

Tuxosaurus:
> If possible I would remove `noMarginBottom` so that we don't accomodate `HeroSection`. If it's too low I would argue that these components need to be in the design system, while they're not we shouldn't create more & more use cases.

src/breadcrumb/Breadcrumb.tsx
Tuxosaurus: `content` is required as a string, not a number. I'm already using `position` as a number for this comparison `position < crumbs.length`

src/breadcrumb/Breadcrumb.tsx
Tuxosaurus: `name` isn't enforced as unique, it sounds stupid that 2 links in breadcrumb would be called the same but I prefer to use the `position` for the uniqueness. And I prefer to keep `name` just to make it more readable.

Tuxosaurus:
> If possible I would remove `noMarginBottom` so that we don't accomodate `HeroSection`. If it's too low I would argue that these components need to be in the design system, while they're not we shouldn't create more & more use cases.

Alright I removed it.

src/breadcrumb/Breadcrumb.tsx
wakooka: Oh yes I missed that, then it's a super nit: `content={String(position)} />`

wakooka (approved):

src/breadcrumb/Breadcrumb.tsx
Tuxosaurus: Done

PR #894, opened 2020-09-24
Review/Caption improvements and fixes

grosbouddha:
- **[FIX]** Add normalized horizontal padding to `Review` component.
- **[UPDATE]** Add support for reply link to `Review` component.
- **[UPDATE]** Update `Caption` to support additional links properly: Some code was already there to display a link but it's not used. I modified to display a link to the far right.

TESTED=Locally + new stories

### DIFF
--- a/src/review/Review.style.tsx
+++ b/src/review/Review.style.tsx
@@ -1,12 +1,14 @@
 import styled from 'styled-components'
 
 import { space } from '../_utils/branding'
+import { normalizeHorizontally } from '../layout/layoutNormalizer'
 
 export const StyledReview = styled.blockquote`
-  padding: ${space.m} 0;
-  margin: 0;
+  padding-top: ${space.m};
+  padding-bottom: ${space.m};
+  ${normalizeHorizontally}
 
-  &.kirk-is-review-response {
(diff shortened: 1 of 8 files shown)


src/review/Review.style.tsx
```diff
@@ -1,11 +1,16 @@
...
 import { space } from '../_utils/branding'
+import { normalizeHorizontally } from '../layout/layoutNormalizer'
 
 export const StyledReview = styled.blockquote`
   padding: ${space.m} 0;
   margin: 0;
 
+  & {
+    ${normalizeHorizontally}
+  }
```
wakooka, line 12: Do you need the override? Or can you just do?
```suggestion
  ${normalizeHorizontally}

wakooka (approved):
Sorry I missed that. Thanks for the fix 🙏

src/review/Review.style.tsx
grosbouddha: I changed it (I removed the code being overriden too).

grosbouddha:
I broke the Review with response in the previous iteration (the extra left padding was missing). I added it back and fixed the story too (I put two Review widget in one story and only the first one was showing :/).

PR #898, opened 2020-09-25
[BBC-9084] Migrate `Stepper` to MDX

mcellier:
## Description

Migrate `Stepper` to MDX

## How it was tested

Local with storybook + kairos


.storybook/preview.js
```diff
@@ -18,6 +18,6 @@ addParameters({
   },
 })
 
-addDecorator(withSmartKnobs())
+addDecorator(withSmartKnobs({ ignoreProps: ['leftAddon'] }))
```
mcellier, line 21: I added this to prevent storybook to go infinite loop on these. LeftAddon generally correspond to other components so i think we can ignore subcomponents like that.

gabrielagabriel: Can you comment this in the code then, 'cos it is a pretty generic change

src/stepper/Stepper.style.tsx
```diff
@@ -0,0 +1,66 @@
+import styled from 'styled-components'
+
+import { color, space } from '../_utils/branding'
+import { StepperButtonSize, StepperDisplay } from './constants'
+
+// These components have a 12px vertical padding. We probably took a shortcut when creating our
+// sizes, as we should have handled 4px multiples (checked with @wakooka)
+const betweenMandL = '12px'
```
gabrielagabriel, line 8: Now that we have sizes can we review that? @wakooka

wakooka: What do you mean "now that we have size"?

wakooka (approved):

src/stepper/Stepper.style.tsx
gabrielagabriel: I meant **spaces** https://github.com/blablacar/ui-library/blob/master/src/_utils/branding.ts#L67 
Actually, we should avoid this kind of exception. And make sure design follow its own rules, hehe

wakooka: IMHO, we shouldn't have an abstraction that doesn't exist on the design side, so I would be up to remove `space`. 
The design follows its own rule since the only rule is that the spacing should a multiple of 4. 

We should only enforce the multiples of 4 through a eslint rule for example. And each `space` value should be local in each component (unless there's a pattern and that the value convey the same meaning).

Example: ItemInfo, ItemAction, Item*, ect, they all have a 24px horizontal padding. Therefore it should be set as a unique value.
But it doesn't make any sense to use `space.m` for example on 2 components which don't share anything and are not related, juste because the value is the same.

PR #902, opened 2020-09-30
Support trip changes in trip card

simonrelet:
## Description

Allow the TripCard component to display a sub text in the driver section.

## What has been done

Add an optional `subText` attribute in the `driver` prop.

Screenshot 2020-09-30 at 18 16 34

## Things to consider

The `driver` prop makes no sens for buses but until we change it, `subText` should be near `ratings` as it is displayed in its place.

## How it was tested

- Storybook
- Unit tests

### DIFF
--- a/src/tripCard/TripCard.tsx
+++ b/src/tripCard/TripCard.tsx
@@ -65,7 +65,7 @@ export type TripCardProps = A11yProps &
 
 const renderPassenger = (passenger: User) => (
   <li className="kirk-tripCard-avatar" key={`${passenger.firstName}-${passenger.avatarUrl}`}>
-    <Avatar image={passenger.avatarUrl} isSmall />
+    <Avatar image={passenger.avatarUrl} alt={passenger.firstName} isSmall />
   </li>
 )
 
@@ -190,7 +190,11 @@ export const TripCard = (props: TripCardProps) => {
                   <div className="kirk-tripCard-bottom-left">
                     {driver && (
                       <div className="kirk-tripCard-driver">
-                        <Avatar className="kirk-tripCard-avatar" image={driver.avatarUrl} />
+                        <Avatar
+                          className="kirk-tripCard-avatar"
+                          image={driver.avatarUrl}
+                          alt={driver.firstName}
+                        />
                         <div className="kirk-tripCard-driver-info">
                           <Text
                             className="kirk-tripCard-driver-name"
--- a/src/tripCard/TripCard.tsx
+++ b/src/tripCard/TripCard.tsx
@@ -211,11 +211,7 @@ export const TripCard = (props: TripCardProps) => {
                   <div className="kirk-tripCard-bottom-left">
                     {driver && (
                       <div className="kirk-tripCard-driver">
-                        <Avatar
-                          className="kirk-tripCard-avatar"
-                          image={driver.avatarUrl}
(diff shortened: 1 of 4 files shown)


src/tripCard/TripCard.unit.tsx
```diff
@@ -1,161 +1,225 @@
...
+      },
+    })
+
+    render(<TripCard {...props} />)
+    expect(screen.getByAltText('Jane')).toBeInTheDocument()
+  })
 
-    const component = shallow(<TripCard {...mockedProps} driver={driver} passengers={null} />)
-    expect(component.find(Avatar)).toHaveLength(1)
+  it('Should render driver sub text', () => {
```
Tuxosaurus, line 176: Nothing wrong but it made me realize we were not testing for the ratings

Tuxosaurus (approved):

src/tripCard/TripCard.tsx
```diff
@@ -65,7 +69,7 @@ export type TripCardProps = A11yProps &
 
 const renderPassenger = (passenger: User) => (
   <li className="kirk-tripCard-avatar" key={`${passenger.firstName}-${passenger.avatarUrl}`}>
-    <Avatar image={passenger.avatarUrl} isSmall />
+    <Avatar image={passenger.avatarUrl} alt={passenger.firstName} isSmall />
```
wakooka, line 72: 👍

src/tripCard/TripCard.tsx
```diff
@@ -190,7 +211,11 @@ export const TripCard = (props: TripCardProps) => {
                   <div className="kirk-tripCard-bottom-left">
                     {driver && (
                       <div className="kirk-tripCard-driver">
-                        <Avatar className="kirk-tripCard-avatar" image={driver.avatarUrl} />
+                        <Avatar
+                          className="kirk-tripCard-avatar"
+                          image={driver.avatarUrl}
+                          alt={driver.firstName}
```
wakooka, line 217: I wouldn't add the alt here, since it's repeating the driver first name displayed juste after, and does not provide additional information.

wakooka (approved):
Thanks for the RTL migration 👍

src/tripCard/TripCard.tsx
simonrelet: From MDN:

> Omitting alt altogether indicates that the image is a key part of the content and no textual equivalent is available. Setting this attribute to an empty string (alt="") indicates that this image is not a key part of the content (it’s decoration or a tracking pixel), and that non-visual browsers may omit it from rendering. Visual browsers will also hide the broken image icon if the alt is empty and the image failed to display.
> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Img#attr-alt

Maybe we should use `alt=""` as the avatar don't add relevant information, the user's name is enough.
What do you think?

cc @nickoola

simonrelet: It looks like it's already the default value in `Avatar`.

src/tripCard/TripCard.unit.tsx
simonrelet: I added one

src/tripCard/TripCard.tsx
wakooka: Yes I forgot to mention it but `Avatar` already handles it :)

PR #903, opened 2020-10-01
Fix `statusInformation` padding in `TripCard`

mcellier:
## Description

Fix `statusInformation` position in `TripCard`

After: 
Capture d’écran 2020-10-01 à 10 55 46
Before: 
Capture d’écran 2020-10-01 à 10 55 13

## What has been done

Override item padding in `TripCard` styles

## Things to consider

Had to use `!important` cause `Item`s are using it...

## How it was tested

Storybook

### DIFF
--- a/src/tripCard/TripCard.style.tsx
+++ b/src/tripCard/TripCard.style.tsx
@@ -26,6 +26,7 @@ export const StyledTripCard = styled(Card)`
 
   & .kirk-item.kirk-tripCard-top-item {
     padding-top: 0;
+    padding-left: 0 !important;
   }
 
   & .kirk-item--highlighted .kirk-text-body {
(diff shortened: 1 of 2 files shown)


src/tripCard/TripCard.style.tsx
```diff
@@ -26,6 +26,7 @@ export const StyledTripCard = styled(Card)`
 
   & .kirk-item.kirk-tripCard-top-item {
     padding-top: 0;
+    padding-left: 0 !important;
```
wakooka, line 29: Ouch! I'm sad to see this, if we had time to finish the normalization we wouldn't have to do this :/
Normally, a `Card` should have a 8px padding (not 16px) and 8px on the Item and other components inside it (the one you removed).

wakooka (approved):

src/tripCard/TripCard.style.tsx
mcellier: `TripCard` might be the most complex component to normalize as it aggregates lots of others :/

PerrineBoissieres (approved):

PR #904, opened 2020-10-01
Fix external touch events triggering Grip touch callbacks

PerrineBoissieres:
## Description

Touching the media component (ex: map zooming) would trigger the `Grip` `onTouchMove` and `onTouchEnd` callbacks. Looks like I wasn't checking my refs right...

## What has been done

Fixed the condition that was checking that a previous `touchstart` event was done on the Grip itself before triggering `move` and `end` events.

## How it was tested

Locally with Chrome and iOS Safari

### DIFF
--- a/src/grip/Grip.tsx
+++ b/src/grip/Grip.tsx
@@ -98,7 +98,6 @@ export const Grip = (props: GripProps): JSX.Element => {
     <div
       className={className}
       onTouchStart={e => {
-        console.log('START')
         fingerYPosition.current = e.touches.item(0).clientY
       }}
       {...a11yAttrs}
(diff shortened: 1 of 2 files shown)


src/grip/Grip.tsx
```diff
@@ -98,6 +98,7 @@ export const Grip = (props: GripProps): JSX.Element => {
     <div
       className={className}
       onTouchStart={e => {
+        console.log('START')
```
mcellier, line 101: Remove console.log

mcellier (changes requested):

src/grip/Grip.tsx
PerrineBoissieres: 🤦

wakooka (approved):

mcellier (approved):

PR #905, opened 2020-10-01
Adding segments compatibilty to Itinerary

wakooka:
## Description

Make `Itinerary` component work with `segments`
image

## What has been done

- Added `ConnectionIcon`
- Added `segments` prop to keep `places` prop retro-compatible 
- Calculated height for svg programmatically, I couldn't find a way to have a `height: auto` without having the user-agent defaulting the value to 350px
 
## Things to consider

This PR aims to unblock Connecting Trips project, for long term I'm writing a design docs on a bigger refacto to make Itinerary more flexible and more aligned with the designs specifications.

## How it was tested

Storybook locally
Currently testing in SPA with a canary release


src/itinerary/index.tsx
```diff
@@ -76,6 +76,10 @@ const StyledItinerary = styled(Itinerary)`
     font-size: ${font.s.size};
   }
 
+  & .kirk-itineraryLocation-chevron {
+    align-self: ${props => (props.segments ? 'baseline' : 'center')};
```
wakooka, line 80: Approved by Jonas. This is because right now it's not possible to add a new item in the Itinerary list without being a place. The hack is to have the "Change of vehicle"  inside the place where the connection occurs.

Tuxosaurus (approved):

src/itinerary/story.tsx
```diff
@@ -25,6 +25,33 @@ const places = [
...
+    href: '#',
+    connectionLabel: 'Change of vehicle',
+  },
+]
+
+const places2 = [
+  {
+    time: '15:00',
+    isoDate: '2017-12-11T15:00',
+    stepAriaLabel: 'Dropoff location',
```
nickoola, line 37: It's just a content detail but the second place of a segment is a **Pickup location** instead of a **Dropoff location**

src/itinerary/story.tsx
```diff
@@ -25,6 +25,33 @@ const places = [
...
+    mainLabel: "Gare d'Agen",
+    subLabel: 'Agen',
+  },
+  {
+    time: '19:00',
+    isoDate: '2017-12-11T19:00',
+    stepAriaLabel: 'Dropoff location',
+    mainLabel: 'Gare De Toulouse',
+    subLabel: 'Toulouse',
+    href: '#',
```
nickoola, line 54: I also suggest to add the `actionAriaLabel` on every place as link, like you do on the first place (Paris).
This way the demo will be more coherent.
example here: `actionAriaLabel: '19:00 Gare De Toulouse (New page with a map)`

nickoola (approved):

src/_internals/itineraryLocation/ItineraryLocation.tsx
```diff
@@ -134,6 +138,12 @@ export const ItineraryLocation = ({
     return renderSubLabel(subLabel)
   }
 
+  const refConnection = useRef<HTMLDivElement>(null)
+  const [height, setHeight] = useState(0)
+  useEffect(() => {
```
simonrelet, line 143: As `useEffect` is called _some times after_ the update is committed to the DOM, there might be some visual glitch. You can use [`useLayoutEffect`](https://reactjs.org/docs/hooks-reference.html#uselayouteffect) which is called right after the DOM update.

> Use this to read layout from the DOM and synchronously re-render. Updates scheduled inside `useLayoutEffect` will be flushed synchronously, before the browser has a chance to paint.
> https://reactjs.org/docs/hooks-reference.html#uselayouteffect

simonrelet (approved):

src/_internals/itineraryLocation/ItineraryLocation.tsx
wakooka: Nice!

src/_internals/itineraryLocation/ItineraryLocation.tsx
```diff
@@ -150,10 +160,30 @@ export const ItineraryLocation = ({
             className="kirk-itineraryLocation-bullet"
             type={isSmall ? BulletTypes.SMALL : BulletTypes.DEFAULT}
           />
-          {hasRoad && <div className="kirk-itineraryLocation-road" />}
+          {displayConnection && (
+            <svg width={8} height={height}>
```
gabrielagabriel, line 164: this is mostly an Icon instead of an Image. We should add it to the IconLibrary to be able to re-use it

src/_internals/itineraryLocation/index.tsx
```diff
@@ -63,6 +62,15 @@ const StyledItineraryLocation = styled(ItineraryLocation)`
     min-width: 0; /* reset flex default to allow shrinking below content width */
   }
 
+  & .kirk-itineraryLocation-connection {
+    display: flex;
+    align-items: center;
+    margin-top: 20px;
+  }
+  & .kirk-itineraryLocation-connection svg {
```
gabrielagabriel, line 70: maybe give a class for it instead of relying on the tag element

src/itinerary/index.tsx
gabrielagabriel: Could you comment the code then, it will be easier to understand the "why" behind it without reading this PR

src/_internals/itineraryLocation/index.tsx
wakooka: Done

src/_internals/itineraryLocation/ItineraryLocation.tsx
wakooka: It's not an icon in Figma, I'd rather not add new discrepancies.

PR #911, opened 2020-10-07
Set label as optional in ItemRadio

mcellier:
## Description

We're using `labelTitle` for the `ReasonStep` in kairos, so this field shouldn't be required

## What has been done

Updated type

## How it was tested

On kairos

### DIFF
--- a/src/itemRadio/ItemRadio.tsx
+++ b/src/itemRadio/ItemRadio.tsx
@@ -22,7 +22,7 @@ export type ItemRadioProps = A11yProps &
     value: string | number
     leftAddon?: React.ReactNode
     labelTitle?: string
-    label: string
+    label?: string
     data?: string
     dataInfo?: string
     checked?: boolean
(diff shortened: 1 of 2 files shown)


src/itemRadio/ItemRadio.tsx
```diff
@@ -22,7 +22,7 @@ export type ItemRadioProps = A11yProps &
     value: string | number
     leftAddon?: React.ReactNode
     labelTitle?: string
-    label: string
+    label?: string
```
gabrielagabriel, line 25: have you checked on #a11y matters? a radio field without label seems quite odd. Should we be able to hide the content instead?

mcellier: Not an actual label on the field. It's just another random prop transfered to `Item`.
The rendered content is actually a label containing the text + the radio. See https://github.com/blablacar/ui-library/blob/master/src/itemRadio/ItemRadio.tsx#L96.

Moreover, with RTL we HAVE to be accessible, otherwise we can't select elements. That's what I checked on kairos related PR.

Tip-Sy (approved):

wakooka (approved):

PR #913, opened 2020-10-09
change separator character used in breadcrumb

Tuxosaurus:
BEFORE
Screen Shot 2020-10-09 at 12 16 54

AFTER
Screen Shot 2020-10-09 at 12 17 11

Tested locally on Firefox Macos

### DIFF
--- a/src/breadcrumb/Breadcrumb.tsx
+++ b/src/breadcrumb/Breadcrumb.tsx
@@ -41,7 +41,7 @@ export const Breadcrumb = ({ crumbs }: BreadcrumbProps) => (
 
               {position < crumbs.length && (
                 <span role="separator" className="breadcrumb-separator">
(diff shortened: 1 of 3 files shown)


mcellier (approved):

src/breadcrumb/Breadcrumb.tsx
```diff
@@ -41,7 +41,7 @@ export const Breadcrumb = ({ crumbs }: BreadcrumbProps) => (
 
               {position < crumbs.length && (
                 <span role="separator" className="breadcrumb-separator">
-                  {' > '}
+                  {' › '}
```
gabrielagabriel, line 44: Please, use an Icon instead of a character

gabrielagabriel (commented):
Use Icon for UI consistency instead of char

PR #914, opened 2020-10-09
Add missing `spellCheck` prop to `TextField`

mcellier:
## How it was tested

On kairos

### DIFF
--- a/src/textField/TextField.tsx
+++ b/src/textField/TextField.tsx
@@ -40,6 +40,7 @@ export type CommonFormFields = Readonly<{
   maxLength?: number
   autoCorrect?: 'on' | 'off'
   autoComplete?: string
+  spellCheck?: 'true' | 'false'
   disabled?: boolean
   readOnly?: boolean
   autoFocus?: boolean
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Unreleased
 
 [...]
-
+- **[UPDATE]** Add `spellCheck` prop to `TextField` type
 - **[UPDATE]** Change separator character used in `Breadcrumb`
 
 # v41.1.0 (08/10/2020)


src/textField/TextField.tsx
```diff
@@ -40,6 +40,7 @@ export type CommonFormFields = Readonly<{
   maxLength?: number
   autoCorrect?: 'on' | 'off'
   autoComplete?: string
+  spellCheck?: 'true' | 'false'
```
wakooka, line 43: I don't think we need this, there was a work @florianeyzat started on this and we should use `inputAttributes` prop to pass the attributes we want.

mcellier: I agree for non-standard props such as data attributes, but this one is from the spec. 
Another solution (in a more global PR) might be to extends `React.HTMLAttributes<HTMLInputElement>`

Tuxosaurus: So you mean replacing these CommonFormFields with InputAttributes? I guess this can be done in another issue.

But I believe whatever we choose to do we also need to add it to Textarea

mcellier: > So you mean replacing these CommonFormFields with InputAttributes? 
Yes
> But I believe whatever we choose to do we also need to add it to Textarea
Do you want me to add this in this PR?

wakooka: > I agree for non-standard props such as data attributes, but this one is from the spec
> Another solution (in a more global PR) might be to extends `React.HTMLAttributes<HTMLInputElement>`

I think it has the wrong type today it's using `InputHTMLAttributes<HTMLInputElement>` and you're right it should use `React.HTMLAttributes<HTMLInputElement>` but it would be good to differentiate props (such as `addon` or `error` and DOM attrtibutes, maybe we need to dicuss this more in chapter.
https://github.com/blablacar/ui-library/blob/36b05fe1932beb9bf6e1290c57aaacfdc493bb56/src/textField/TextField.tsx#L74

wakooka (approved):

Tuxosaurus (approved):

src/textField/TextField.tsx
Tuxosaurus: If you want to add it to Textarea, do it but I'm not requiring it

mcellier: @wakooka I think I'll prefer to have the following:

```
export type TextFieldProps = React.HTMLAttributes<HTMLInputElement> & CommonFormFields &
  Readonly<{
    defaultValue?: string
    labelledBy?: string
    onChange?: (obj: OnChangeParameters) => void
    onClear?: () => void
    className?: string
    errorClassName?: string
    error?: errorField
    addon?: JSX.Element
    label?: string
    buttonTitle?: string
    focus?: boolean
    inputMode?: inputModes
    pattern?: string
    inputRef?: (input: textfield) => void
    format?: (value: string, previousValue: string) => string
    focusBorder?: boolean
    loader?: JSX.Element
  }>
```

I'm not fond of this `inputAttributes` object. We might have duplication with existing fields (in CommonFormFields for instance) and it seems quite unnatural. WDYT?

wakooka: I think there are 2 things:
1. From consumer: passing any html attributes that will ouput directly on the <input>
1. using a `inputAttributes` object prop (of course removing them from commonFormField to avoid duplication) VS passing the props at the same level of non-html attribute props (such as `error`, `errorClassName`, `loader`, etc)

On 1) I think we all agree that it would be great
On 2) I tend to prefer not to mix attributes and props but not a super strong opinion on this one

mcellier: Let's talk about it in the next chapter. Should I merge this PR in the meantime or use the inputAttributes solution?

wakooka: Yes you can merge it's approved ✅

PR #917, opened 2020-10-12
Add TextFieldsSection to display multiple TextFields in the same row

florianeyzat:
## Description

Everything is in the name 🙂 

Goal is to be able to do that using only ui-library components:

Capture d’écran 2020-10-12 à 14 19 19

## How it was tested

Locally with main project + storybook


src/layout/section/textFieldsSection/TextFieldsSection.tsx
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Inputs Section: display two or more TextFieldProps on a row.
+ */
+const TextFieldsSectionDefinition = (props: TextFieldsSectionProps) => {
+  const { className, children, tag = <div /> } = props
+
+  return (
+    <BaseSection
+      {...props}
+      tagName={tag.type}
```
wakooka, line 24: It looks like  `BaseSection` is expecting a string, maybe it would be to keep it that way in this component too.
`tagName : string` and  tagName={tagName} for consistency. WDYT?

wakooka (approved):

src/layout/section/textFieldsSection/TextFieldsSection.tsx
florianeyzat: Agree 🙂 Updating that

PR #919, opened 2020-10-15
Itinerary withTime support for places only

wakooka:
## Description

Itinerary withTime support for places only

## What has been done

withTime now only applies on places prop

## Things to consider

This is quick fix as discussed.

## How it was tested

Locally storybook


CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Unreleased
 
-[...]
+- **[FIX]** In `Itinerary` Only supports `withTime` for `places`
```
mcellier, line 3: nit: `- **[FIX]** Always display time in `Itinerary` with segments

mcellier (approved):

CHANGELOG.md
wakooka: Much better, thanks

PR #923, opened 2020-10-22
Remove DEFAULT position for SlideSection on small screens

PerrineBoissieres:
## Description

Small phone screens can't handle displaying both the map and the list, so we remove this layout position if the screen height is too small to handle it. The initial position will be with the list fully open by default for those screens, and you can only switch between this position, and the position where the media is fully shown. We also only show the GripHandle (with a small additional height for touch comfort) for the latter position.

## How it was tested

Locally on Chrome, Firefox and Safari iOS (main reference for screen size: iPhone SE)

### DIFF
--- a/src/layout/section/slideSection/SlideSection.tsx
+++ b/src/layout/section/slideSection/SlideSection.tsx
@@ -5,6 +5,8 @@ import { Grip } from '../../../grip'
 import { GRIP_HANDLE_HEIGHT } from '../../../grip/GripHandle'
 import { StyledSlideLayout, StyledSlidePanel } from './SlideSection.style'
 
+const SMALL_SCREEN_BREAKPOINT = 450
+
 export enum SlideSectionPosition {
   DEFAULT = 'default',
   REDUCED = 'reduced',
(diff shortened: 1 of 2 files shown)


src/layout/section/slideSection/SlideSection.tsx
```diff
@@ -67,7 +81,8 @@ export const SlideSection = (props: SlideSectionProps): JSX.Element => {
   const [expandedHeight, setExpandedHeight] = useState<number>(0)
   useEffect(() => {
     // Reduced height is the height of reducedContent element + GripHandle height
-    const reducedContentHeight = reducedContentRef.current.clientHeight
+    // On small screen, we only show a 48px height element for touch control
+    const reducedContentHeight = isScreenTooSmall ? 16 : reducedContentRef.current.clientHeight
```
mcellier, line 85: use constant

src/layout/section/slideSection/SlideSection.tsx
```diff
@@ -45,7 +57,9 @@ export const SlideSection = (props: SlideSectionProps): JSX.Element => {
   const slideDown = useCallback(() => {
     if (position === SlideSectionPosition.DEFAULT) {
       setPosition(SlideSectionPosition.REDUCED)
-    } else if (position === SlideSectionPosition.EXPANDED) {
+    } else if (position === SlideSectionPosition.EXPANDED && isScreenTooSmall) {
```
mcellier, line 60: Unit tests will be welcome for theses

src/layout/section/slideSection/SlideSection.tsx
gabrielagabriel: no magic numbers.. why `16`?

src/layout/section/slideSection/SlideSection.tsx
PerrineBoissieres: I can't find a way to test this with RTL and Jest: the component height on which I'm based is always set to zero in tests for whatever reason =/

src/layout/section/slideSection/SlideSection.tsx
PerrineBoissieres: Reason was to have a bit more height to the reduced element so you can manipulate it with touch events with ease. I'll change that so it's not a magic number.

mcellier (approved):

PR #924, opened 2020-10-22
Add noHorizontalSpacing props on ItemsSection component

pith:
## Description

Forward the noHorizontalSpacing props to the BaseSection to normalise the ItemsSection component.
image

## How it was tested

locally on storybook

### DIFF
--- a/src/layout/section/itemsSection/itemsSection.tsx
+++ b/src/layout/section/itemsSection/itemsSection.tsx
@@ -6,6 +6,7 @@ import { BaseSection } from '../../../layout/section/baseSection'
 export type ItemsSectionProps = Readonly<{
   children: React.ReactElement<ItemInfoProps>[]
   className?: string
+  noHorizontalSpacing?: boolean
   tag?: JSX.Element
 }>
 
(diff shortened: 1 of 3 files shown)


src/layout/section/itemsSection/itemsSection.tsx
```diff
@@ -14,13 +15,14 @@ export type ItemsSectionProps = Readonly<{
...
 export const ItemsSection = (props: ItemsSectionProps) => {
-  const { className, children, tag = <div /> } = props
+  const { className, children, noHorizontalSpacing, tag = <div /> } = props
 
   return (
     <BaseSection
       tagName={tag.type}
       className={className}
       contentClassName="kirk-items-section-content"
+      noHorizontalSpacing={noHorizontalSpacing}
```
wakooka, line 25: I'm guessing `ItemSection` should only take `Item*` as children. So I think it's safe to hardcode noHorizontalSpacing to true instead of forwarding it.
This is because all `Item*`s have been normalized.

wakooka (changes requested):

grosbouddha (approved):

src/layout/section/itemsSection/itemsSection.tsx
pith: Even better! Done you can check it again

wakooka (approved):

src/layout/section/itemsSection/itemsSection.tsx
wakooka: Sorry didn't see the notification :(

PR #932, opened 2020-11-02
Migrate pages examples to MDX

mcellier:
## Description

- No more storiesOf
- Moved pages to _internal/pages
- Added some noHorizontalSpacing to fix some layout issues
- Added baseUrl to tsconfig to avoid LONG relative imports

## Things to consider

We have lots of layout issues on these pages due to:
- css being in kairos
- layout normalization not being finished

## How it was tested

Locally on storybook


gabrielagabriel:
hey! Could you move your PR to `mdx` branch, plz?

gabrielagabriel (changes requested):
IMHO Pages shouldn't move to "internals" since it's not utilitarian or a reusable function, etc...
Pages are actually a "Playground / Demo"

mcellier:
> IMHO Pages shouldn't move to "internals" since it's not utilitarian or a reusable function, etc...
> Pages are actually a "Playground / Demo"

Putting them in `internals` directory allows us to avoid importing them from kairos with our eslint rules. 
Letting them in the `src` directory is also misleading IMHO. 
Last but not least, I asked into our migration channel this morning :(

tsconfig.json
```diff
@@ -15,7 +15,8 @@
     "allowJs": false,
     "declaration": true,
     "noEmitOnError": true,
-    "allowSyntheticDefaultImports": true
+    "allowSyntheticDefaultImports": true,
+    "baseUrl": "./src"
```
wakooka, line 19: I understand why you added this but I don't think we can, see #717.

wakooka (changes requested):

tsconfig.json
mcellier: I removed, thanks!

wakooka (approved):

gabrielagabriel (approved):

PR #934, opened 2020-11-02
[BOYSCOUT] Migrate typography components to MDX

mcellier:
## Description

Migrate the whole `src/typography` components to MDX

## What has been done

- Migrated components to the new file structure
- Updated their stories

## Things to consider

- Not using the exact structure, as all these are just styled components. ie: we just export the styled `Text` from `Component.style.tsx`, so we don't have a `Component.tsx`

## How it was tested

Locally in storybook


src/typography/body/Body.style.tsx
```diff
@@ -0,0 +1,11 @@
+import styled from 'styled-components'
+
+import { color, font, fontWeight } from '../../_utils/branding'
+import { Text as TextStyle } from '../index'
```
gabrielagabriel, line 4: Why alias it?

mcellier: Didn't change that code. Fixed it.

wakooka (approved):

gabrielagabriel (approved):

PR #959, opened 2020-11-04
[BBC-9072] Storybook: Migrate Button to MDX

mcellier:
## Description

screencapture-localhost-4000-2020-11-04-10_50_28

## Things to consider

- Can't use styled components the same way we do for all other components, as buttons are created dynamically with `React.createElement` (and can be a `a`, a `button` or even every other React component => SlugLink for instance).
- Weird issue on the storybook: I have to randomly click multiple times on the button to trigger the action (loading state). Hope this is related to 5.3.

## How it was tested

Locally on storybook + kairos


src/button/index.tsx
```diff
@@ -1,181 +1,6 @@
...
-`
-
+// Note: can't use styled components the same way we do for all other components,
+// as buttons are created dynamically with React.createElement (and can be a "a", "button" or
+// even every other React component => SlugLink for instance).
 export { ButtonStatus, ButtonProps } from './Button'
-export { StyledButton as Button }
-export default StyledButton
+export { StyledButton as Button } from './Button.style'
+export { StyledButton as default } from './Button.style'
```
gabrielagabriel, line 6: `Button` should be the default

src/button/Button.style.tsx
```diff
@@ -0,0 +1,177 @@
+import styled from 'styled-components'
+
+import { color, componentSizes, font, shadow, space, transition } from '../_utils/branding'
+import { Button } from './Button'
```
gabrielagabriel, line 4:
```suggestion
import { Button } from './index'
```

src/button/Button.story.mdx
```diff
@@ -0,0 +1,237 @@
+import { useState } from 'react'
+import { action } from '@storybook/addon-actions'
+import { Meta, Preview, Props } from '@storybook/addon-docs/blocks'
+
+import { color } from '../_utils/branding'
+import { Button, ButtonStatus } from '../button'
```
gabrielagabriel, line 6:
```suggestion
import { Button, ButtonStatus } from '../index'
```

src/button/Button.story.mdx
```diff
@@ -0,0 +1,237 @@
...
+  )
+}
+
+<Meta title="Buttons|Button" />
+
+# Button
+
+## Primary button
+
+> With text
```
pith, line 42: I know this format is widely used in our current doc but it is meant for quotes not for title. I would recommend using ### instead.

src/button/index.tsx
mcellier: nop. Read the comment

src/button/Button.style.tsx
mcellier: nop

src/button/Button.story.mdx
```diff
@@ -0,0 +1,237 @@
...
+  onClick: action('clicked'),
+  onMouseDown: action('mouseDown'),
+  onMouseUp: action('mouseUp'),
+  onTouchStart: action('touchStart'),
+  onTouchEnd: action('touchEnd'),
+  onFocus: action('focused'),
+  onBlur: action('blured'),
+}
+
+export const renderPrimaryButton = () => {
```
pith, line 19: This would be an interesting example in the "Usage" section below.

src/button/Button.story.mdx
mcellier: All stories are impacted due to the generator we're using. Let's do a boyscout PR later for this

src/button/Button.story.mdx
```diff
@@ -0,0 +1,237 @@
+import { useState } from 'react'
+import { action } from '@storybook/addon-actions'
+import { Meta, Preview, Props } from '@storybook/addon-docs/blocks'
+
+import { color } from '../_utils/branding'
+import { Button, ButtonStatus } from '../button'
+import { ArrowIcon } from '../icon/arrowIcon'
+
+export const commonProps = {
+  onClick: action('clicked'),
```
pith, line 10: 😮 I didn't know that.

src/button/Button.story.mdx
```diff
@@ -0,0 +1,254 @@
...
+
+## Secondary button
+
+These buttons can be used as a secondary option when there’s already a primary button action.
+
+When user taps on it there’s a pressed state.
+After tapping if request is already loaded we either go to the next page or close the modal or if it’s a final decision we display the checked state then close the page (in some cases we don’t change the page like create alert in Search Results).
+If request is not loaded we display the loading state of these buttons. When loaded, we either go to the next page or close the modal or if it’s a final decision we display the checked state.
+
+<Preview>
```
pith, line 124: Note if you don't use `<Story name="...">` each example won't be displayed in the menu or in the canvas tab.

mcellier: I know, but right now if we put different stories in different previews, props are mixed up and everything looks the same. Tried to also add the `<Div>` I saw on slack yesterday, but it didn't work :(

florianeyzat (approved):

src/button/Button.story.mdx
wakooka: I thought we couldn't use it anymore in mdx for some reasons.

src/button/Button.story.mdx
wakooka: Would be good that we only have one workaround, it will be easier to spot when migrated to 6.0. And I'm totally ok to remove my `<Div />` from ConfirmationModal to do like what you did here. wdyt?

wakooka (approved):

src/button/Button.story.mdx
mcellier: Honestly, I'm hoping we'll upgrade soon and pass on all stories to fix these kind of things

src/button/Button.story.mdx
mcellier: Well, it seems to be working

PR #964, opened 2020-11-04
[BBC-9083] Storybook: SearchForm

florianeyzat:
## Description

Storybook migration of the SearchForm. Few things worth noticing before review:
* I didn't change the sub elements since they were not directly concerned by the new "architecture" with the story file
* I tried creating a sub-story for each component to organize a bit and it worked BUT the generated doc was broken into pieces so I removed that
* I organized a bit the story + titles, to avoid having just one big preview block
* I had to keep the text knob for the transitions

## How it was tested

Storybook + main project


src/searchForm/SearchForm.story.mdx
```diff
@@ -0,0 +1,199 @@
...
+import { DatePickerOverlay } from './datePicker/overlay'
+import { DatePickerSection } from './datePicker/section'
+import { SearchForm } from './index'
+import { SlideSwitchTransition, SlideSwitchTransitionSide } from './SlideSwitchTransition'
+import { StepperOverlay } from './stepper/overlay'
+import { StepperSection } from './stepper/section'
+
+<Meta title="Widgets|SearchForm" />
+
+# **QrCard**
```
gabrielagabriel, line 21:
```suggestion
# **SearchForm**
## QrCard
```

src/searchForm/SearchForm.story.mdx
```diff
@@ -0,0 +1,199 @@
...
+import { AutoCompleteOverlay } from './autoComplete/overlay'
+import { AutoCompleteSection } from './autoComplete/section'
+import { DatePickerOverlay } from './datePicker/overlay'
+import { DatePickerSection } from './datePicker/section'
+import { SearchForm } from './index'
+import { SlideSwitchTransition, SlideSwitchTransitionSide } from './SlideSwitchTransition'
+import { StepperOverlay } from './stepper/overlay'
+import { StepperSection } from './stepper/section'
+
+<Meta title="Widgets|SearchForm" />
```
gabrielagabriel, line 19: I would suggest that you split this story on stories and keep only the "SearchForm" here

wakooka: Following @florianeyzat s comment in the description, I won't be updating this part:
> I tried creating a sub-story for each component to organize a bit and it worked BUT the generated doc was broken into pieces so I removed that

wakooka (approved):

src/searchForm/SearchForm.story.mdx
wakooka: Done

src/searchForm/SearchForm.story.mdx
wakooka: I removed the other stories for some reasons we couldn't scroll on the docs page. The other stories were good for dev when we were working on those component, but since they're search form internals they shouldn't have their own stories

mcellier (approved):

PR #971, opened 2020-11-05
[BBC-10943] Storybook: Text MDX

mcellier:
## Description

Convert `text` component to MDX

## What has been done

Same as `button`: we have a dynamic html tag, so I used style-components the old way 

## How it was tested

Storybook (Text + various Item that use it)


src/text/Text.story.mdx
```diff
@@ -0,0 +1,42 @@
+import { Meta, Preview, Props, Story } from '@storybook/addon-docs/blocks'
+
+import { color } from '../_utils/branding'
+import { Text, TextDisplayType, TextTagType } from './index'
+
+<Meta title="Text" />
```
gabrielagabriel, line 6: Does it means
Text => Default on the SideBar?

mcellier: Yes

PR #981, opened 2020-11-13
Add media size hooks and test set-up helper prop

grosbouddha:
## Description

- Add new useIsSmallMediaSize and useIsLargeMediaSize custom hooks
- Add new mediaSizeForTestsOnly to set up the MediaSizeProvider in tests

## TESTED

Unit tests

### DIFF
--- a/src/_utils/mediaSizeProvider/index.tsx
+++ b/src/_utils/mediaSizeProvider/index.tsx
@@ -10,14 +10,26 @@ export enum MediaSize {
 
 export type MediaSizeProviderProps = Readonly<{
   children: ReactNode
+  mediaSizeForTestsOnly?: MediaSize
 }>
 
-export const MediaSizeContext = React.createContext(MediaSize.SMALL)
+const DEFAULT_MEDIA_SIZE = MediaSize.SMALL
+export const MediaSizeContext = React.createContext(DEFAULT_MEDIA_SIZE)
 
 const DEBOUNCE_VALUE = 500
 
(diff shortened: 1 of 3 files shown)


wakooka (approved):
Thanks 👏

src/_utils/mediaSizeProvider/index.tsx
```diff
@@ -10,14 +10,26 @@ export enum MediaSize {
 
 export type MediaSizeProviderProps = Readonly<{
   children: ReactNode
+  mediaSizeForTestsOnly?: MediaSize
```
simonrelet, line 13: nit: An other solution (seen in open source projects) is to export a `TestMediaSizeProvider`/`MockMediaSizeProvider`/`MemoryMediaSizeProvider`. It avoid mistakes in production code.

simonrelet (approved):

src/_utils/mediaSizeProvider/index.tsx
grosbouddha: I will move forward with mediaSizeForTestsOnly to unblock my test PRs. But we can easily migrate to a fake implementation later (we will have only one usage in kairos for this test param).
As well, mistake wise, if somebody can use a mediaSizeForTestsOnly props in prod, they can use a fake implemetation too :) unless we combine it with some test-linting rules (which we don't have yet).

I'm wondering how much test-only code is shipped in production binary as we speak :)

PR #985, opened 2020-11-17
Don't deactivate/activate the trap in useFocusTrap each time onClose …

simonrelet:
…changes

## Description

When `useFocusTrap` is used like this:

```ts
function MyComponent() {
  const element = React.useRef<HTMLElement>(null!)
  const onClose = ()  => {}
  useFocusTrap(element, onClose)
  //...
}
```

At each render, `onClose` is re-created and forces the trap to deactivate then re-activate.

## What has been done

- Use RTL and RHTL for the tests.
- Use a Ref inside `useFocusTrap` to avoid adding `onClose` to the effect dependencies.

## How it was tested

- Unit tests
- Locally in app


package.json
```diff
@@ -58,7 +58,8 @@
     "@storybook/addons": "6.0.28",
     "@storybook/react": "6.0.28",
     "@testing-library/jest-dom": "5.8.0",
-    "@testing-library/react": "10.0.4",
+    "@testing-library/react": "11.1.2",
```
simonrelet, line 61: I need the latest version of @testing-library/react in order for @testing-library/react-hooks to work properly

wakooka (approved):
Great! 👍  I didn't even think of focus trap on this issue :/

PR #990, opened 2020-11-19
Bbc 8588 search form strange styles during loading

simonrelet:
## Description

Fix the search form strange behaviour during the page load.

## What has been done

- Move global `matchMedia` mock to _/bin/jestScript.js_.
- Use responsive dividers in the form to avoid conditional rendering.

Before | After
------------ | -------------
before | after

## Things to consider

Two dividers were created that are specific to the search form: `VerticalDivider` and `ResponsiveDivider`.

## How it was tested

- Locally with storybook


src/searchForm/SearchForm.style.tsx
```diff
@@ -14,8 +14,7 @@ const formWidth = '928px'
 const formHeight = '56px'
 const primaryFieldsWidth = '240px'
 const secondaryFieldsWidth = '160px'
-const buttonHeight = '48px'
-const buttonHeightSmall = '40px'
```
simonrelet, line 18: After looking at the designs specs in Figma, both small and large height are 40px.

src/searchForm/SearchForm.tsx
```diff
@@ -103,9 +103,8 @@ export const SearchForm = ({
   datepickerProps,
   stepperProps,
 }: SearchFormProps) => {
-  // Use React.useContext syntax so we can mock it
-  // https://github.com/enzymejs/enzyme/issues/2176#issuecomment-533582429
-  const mediaSize = React.useContext(MediaSizeContext)
+  const isLargeMediaSize = useIsLargeMediaSize()
+  const isSmallMediaSize = !isLargeMediaSize
```
wakooka, line 107: nit: 
```suggestions
const isSmallMediaSize = useIsSmallMediaSize()

wakooka (approved):

src/searchForm/SearchForm.tsx
simonrelet: Why do we need a hook for this? 🤔

wakooka: How I see it: `useIsSmallMediaSize` already exists so I thought we should use it since it already abstracts the knowledge on whether it's small or not. Here we're "duplicating" the logic by saying `isSmallMediaSize = !isLargeMediaSize`. But this logic is so small that I'm okay to keep it as it is that's why it's really really really nitpicking 🙂

simonrelet: I understand, my question was actually "Why did we create the hook in the first place?".
I think I'll keep it like this because I don't think this abstraction is useful.

wakooka: Ahhh sorry, then I'm guessing it's for readability & flexibility (in case we have a `medium` size in the future which I'm doubting) but I agree it's not super useful and maybe a bit overkill

Tuxosaurus:
Nice, thanks for that fix!

PR #991, opened 2020-11-19
Add dataEncryptedName prop on TextField

florianeyzat:
## Description

We've talked about this for a long time so now it's done 😄 

## What has been done

Added a `dataEncryptedName` prop on the text field so it can be passed directly instead of via `inputAttributes`

## How it was tested

Integrated in the main project

### DIFF
--- a/src/textField/TextField.tsx
+++ b/src/textField/TextField.tsx
@@ -72,6 +72,7 @@ export type TextFieldProps = CommonFormFields &
     focusBorder?: boolean
     loader?: JSX.Element
     inputAttributes: InputHTMLAttributes<HTMLInputElement>
+    dataEncryptedName?: string
   }>
 
 export type TextFieldState = {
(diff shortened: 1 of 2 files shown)


src/textField/TextField.tsx
```diff
@@ -272,6 +275,10 @@ export class TextField extends PureComponent<TextFieldProps, TextFieldState> {
       attrs['aria-invalid'] = 'true'
     }
 
+    if (dataEncryptedName) {
+      attrs['data-encrypted-name'] = dataEncryptedName
```
gabrielagabriel, line 279: I don't see any value in passing this prop instead of adding the `data-ecrypted-name` on the component itself, is there a strong reason for this verbbose?

wakooka:
I'm confused, didn't we say that we wanted any prop to be spread on the `<input />` see: https://blablacar.atlassian.net/browse/BBC-10897

src/textField/TextField.tsx
florianeyzat: Every prop is using camelCase, so I just respected that format 🙂

florianeyzat:
@wakooka I don't understand, I've been asked in a weekly chapter to re-do this 🤔

wakooka:
> @wakooka I don't understand, I've been asked in a weekly chapter to re-do this 🤔

Look at the comment from @mcellier on https://blablacar.atlassian.net/wiki/spaces/SPA/pages/805339357/2020-10-20+SPA+Weekly
We should first do this: https://blablacar.atlassian.net/browse/BBC-10897

PR #993, opened 2020-11-20
[BBC-11168] Add 8 car icons

mcellier:
## Description

Capture d’écran 2020-11-20 à 11 20 57

## What has been done

Optimized through SVGOMG

## How it was tested

Storybook


src/icon/carConvertibleIcon.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import { BaseIcon, BaseIconDefaultProps, Icon } from '../_utils/icon'
+
+export const CarConvertibleIcon = (props: Icon) => (
+  <BaseIcon {...props}>
+    <path
+      fillRule="evenodd"
+      clipRule="evenodd"
+      d="M14.046 8.21A.495.495 0 0114 8a.5.5 0 01.492-.5h.008c.245 0 .487.112.799.313a7.646 7.646 0 01.71.531L19.18 11h.318c1.488 0 2.5 1.02 2.5 2.5v1.666c0 .46-.448.834-1 .834h-1.55a2.5 2.5 0 01-4.9 0h-5.1a2.5 2.5 0 01-4.9 0H3c-.552 0-1-.373-1-.834v-3.333c0-.46.448-.833 1-.833h8v-1a.5.5 0 011 0v1h3.242l-1.196-2.79zM21 15h-1.55a2.5 2.5 0 00-4.9 0h-5.1a2.5 2.5 0 00-4.9 0H3v-3h16.5a1.5 1.5 0 011.5 1.5V15zm-4.67-4h1.293l-2.016-1.687L16.33 11zM8.5 15.5a1.5 1.5 0 11-3 0 1.5 1.5 0 013 0zM17 17a1.5 1.5 0 100-3 1.5 1.5 0 000 3z"
+      fill={props.iconColor}
```
gabrielagabriel, line 11: you can use css(in-js) instead

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **[NEW]** Add `CarConvertibleIcon`, `CarEstateIcon`, `CarHatchbackIcon`, `CarMinivanIcon`, `CarMvpIcon`, `CarSaloonIcon`, `CarSuvIcon`, `CarVanIcon`
```
gabrielagabriel, line 3: I've let a comment on Figma, but _Estate_ is _Break_ in English and it should be reviewed
https://www.linguee.fr/anglais-francais/traduction/estate+car.html

src/icon/carConvertibleIcon.tsx
mcellier: Just doing the same as LOTS of icons. We should update all of them at once if we wanna change

CHANGELOG.md
mcellier: Renamed it.

PerrineBoissieres (approved):

src/icon/carConvertibleIcon.tsx
Tuxosaurus: Agreed, and it would have a bigger impact as a "point of reference". Also when doing that we may want to find a place in ui-library (icons doc?) to explicit this.

Tuxosaurus (approved):

src/icon/carConvertibleIcon.tsx
gabrielagabriel: Just sad that I'm reviewing a LOTS of PRs and getting the very same issue and I can't avoid it to keep going :(

pith: I agree with @mcellier, we should updated the other icons, I also copied this `fill={props.iconColor}` on new icons. And to be honest, I'm not sure to know what is the good way to do it with `css-in-js`. Can you share an example of what you expect?

PR #994, opened 2020-11-20
[BBC-11165] Add `ColorIcon`

mcellier:
## Description

Capture d’écran 2020-11-20 à 11 44 48

## What has been done

New icon using `iconColor` to fill the icon. Added a new `borderColor` prop to this icon to handle some specific cases

## Things to consider

The white is the only color that uses a border (black). On iOS, the border is set on the black icon when the user is in dark mode. I chose to stay generic in the library.

## How it was tested

Storybook


src/icon/colorIcon.tsx
```diff
@@ -0,0 +1,25 @@
...
+    borderColor?: string
+  }>
+
+export const ColorIcon = (props: ColorIconProps) => (
+  <BaseIcon {...props}>
+    <circle
+      cx="12"
+      cy="12"
+      r="9.5"
+      stroke={props.borderColor || props.iconColor}
```
gabrielagabriel, line 16: please use **CSS**(in-js) for that

src/icon/colorIcon.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+
+import { BaseIcon, BaseIconDefaultProps, Icon } from '../_utils/icon'
+
+export type ColorIconProps = Icon &
+  Readonly<{
+    borderColor?: string
```
gabrielagabriel, line 7: Also which colors are allowed? the whole palette? Should you type the colors?

wakooka:
Do you think we can use `CircleIcon` instead with a few modifications?

mcellier:
> Do you think we can use `CircleIcon` instead with a few modifications?

I thought about it, but `CircleIcon` already has LOT of logic in it (spinning & so on). I prefered the smaller, more specialized approach. Especially if one day we want to be smarter about the border thing (dark mode)

src/icon/colorIcon.tsx
mcellier: Color is sent by the backend (as explained in the FADR). We don't know what they'll send us, even if I don't think it changes that often.

src/icon/colorIcon.tsx
mcellier: Same answer as my previous PR.

wakooka (approved):

PerrineBoissieres (approved):

Tuxosaurus (approved):

src/icon/colorIcon.tsx
gabrielagabriel: not the same issue! here is even more problematic since we are treating fill and stroke

PR #998, opened 2020-11-24
[PROPIM-609] Add prop divider with DividerPosition enum to Breadcrumb

Tuxosaurus:
The Breadcrumb component on SEO pages will be moved from the top of the page to the bottom to make way for new content.
Hence, the ContentDivider that is included in Breadcrumb should shift from being below the actual breadcrumb to above. I also chose to create a case for no divider, which would be useful in the case of "directory pages" since the list of items ends with their own divider, so we wouldn't want to double that.

Screen Shot 2020-11-24 at 17 06 00

Tested locally on Firefox MacOs, with canary release

### DIFF
--- a/src/breadcrumb/Breadcrumb.style.tsx
+++ b/src/breadcrumb/Breadcrumb.style.tsx
@@ -1,9 +1,22 @@
 import styled from 'styled-components'
 
 import { color, space } from '../_utils/branding'
+import { DividerPosition } from './Breadcrumb'
 
-export const StyledBreadcrumb = styled.ol`
-  padding: ${space.l} ${space.xl} ${space.m};
+export const StyledBreadcrumb = styled.ol<{
+  divider: DividerPosition
+}>`
+  padding: ${props => {
+    if (props.divider === DividerPosition.TOP) {
+      return `${space.m} ${space.xl} ${space.l}`
+    }
+    if (props.divider === DividerPosition.BOTTOM) {
+      return `${space.l} ${space.xl} ${space.m}`
+    }
+
+    return `${space.l} ${space.xl} ${space.l}`
+  }};
+
   margin: 0;
   list-style: none;
 
--- a/src/breadcrumb/Breadcrumb.tsx
+++ b/src/breadcrumb/Breadcrumb.tsx
@@ -10,13 +10,22 @@ export type CrumbProps = Readonly<{
   url: string
 }>
 
-export type BreadcrumbProps = {
-  crumbs: Array<CrumbProps>
+export enum DividerPosition {
+  TOP = 'top',
+  BOTTOM = 'bottom',
+  NONE = 'none',
 }
 
-export const Breadcrumb = ({ crumbs }: BreadcrumbProps) => (
+export type BreadcrumbProps = {
+  crumbs: Array<CrumbProps>
+  divider?: DividerPosition
+}
+
+export const Breadcrumb = ({ crumbs, divider }: BreadcrumbProps) => (
   <Fragment>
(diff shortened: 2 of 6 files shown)


src/breadcrumb/Breadcrumb.style.tsx
```diff
@@ -1,9 +1,22 @@
 import styled from 'styled-components'
 
 import { color, space } from '../_utils/branding'
+import { DividerPosition } from './Breadcrumb'
+
+export const StyledBreadcrumb = styled.ol<{
+  divider: DividerPosition
+}>`
+  padding: ${props => {
```
gabrielagabriel, line 9: the beauty of css in js! Good job!

src/breadcrumb/Breadcrumb.tsx
```diff
@@ -10,13 +10,22 @@ export type CrumbProps = Readonly<{
...
+  NONE = 'none',
+}
+
 export type BreadcrumbProps = {
   crumbs: Array<CrumbProps>
+  divider?: DividerPosition
 }
 
-export const Breadcrumb = ({ crumbs }: BreadcrumbProps) => (
+export const Breadcrumb = ({ crumbs, divider }: BreadcrumbProps) => (
```
gabrielagabriel, line 24: The render function still a function and you can sort the elements by composing instead of repeating yourself.
Maybe you should try something like:

```suggestion
const dividerElm = <ContentDivider />
const breadcrumElm = <StyledBreadcrumb divider={divider} itemScope itemType="https://schema.org/BreadcrumbList" />

return (
<Fragment>{divider === DividerPosition.TOP ? [dividerElm, breadcrumElm] : [breadcrumElm, dividerElm]}</Fragment>
)
```

src/breadcrumb/Breadcrumb.unit.tsx
```diff
@@ -20,9 +20,17 @@ describe('Breadcrumb', () => {
...
     expect(screen.getAllByRole('listitem')).toHaveLength(2)
     expect(screen.getByRole('list')).toHaveTextContent('Fhtagn › Cthulhu')
   })
+  it('renders with no divider ', () => {
+    render(<Breadcrumb crumbs={[crumb1]} />)
+    expect(screen.queryByRole('separator')).not.toBeInTheDocument()
+  })
+  it('renders with a divider on top', () => {
+    render(<Breadcrumb crumbs={[crumb1]} divider={DividerPosition.TOP} />)
+    expect(screen.queryByRole('separator')).toBeInTheDocument()
```
gabrielagabriel, line 34: you should check the order of the elements on the DOM to make sure its before or after

gabrielagabriel (approved):

src/breadcrumb/Breadcrumb.tsx
wakooka: I disagree, I'd rather keep JSX rendering declarative, it's easier to read than building it programatically.

wakooka (approved):
Sorry for the delayed review

src/breadcrumb/Breadcrumb.tsx
Tuxosaurus: I don't have a strong opinion, I'd say in this case there's not much complexity so I prefer to read it as a whole in the render. Also there's not just TOP or BOTTOM but also NONE it's not just a ternary condition.

src/breadcrumb/Breadcrumb.unit.tsx
Tuxosaurus: Done

PR #999, opened 2020-11-25
Fix SearchForm autocomplete styles

simonrelet:
## Description

Fix style override in `SearchForm`.
The issue only appears in production mode.

## What has been done

Use `!important` as quick fix, better fix would be not to use nested selectors for styles override, it breaks components isolation and introduce these kinds of errors.

Before | After
------ | ------
Screenshot 2020-11-25 at 13 23 00 | Screenshot 2020-11-25 at 13 22 43

## Things to consider

Nothing.

## How it was tested

- Locally

### DIFF
--- a/src/searchForm/baseStyles.tsx
+++ b/src/searchForm/baseStyles.tsx
@@ -32,9 +32,9 @@ export const sectionBaseStyle = `
 
 export const autoCompleteBaseStyle = `
   .kirk-textField-wrapper {
-    background: transparent;
-    border-radius: 0;
-    border: 0;
+    background: transparent !important;
+    border-radius: 0 !important;
+    border: 0 !important;
 
     input {
       font-size: ${font.m.size};


src/searchForm/baseStyles.tsx
```diff
@@ -32,9 +32,9 @@ export const sectionBaseStyle = `
 
 export const autoCompleteBaseStyle = `
   .kirk-textField-wrapper {
-    background: transparent;
-    border-radius: 0;
-    border: 0;
+    background: transparent !important;
```
gabrielagabriel, line 35: can't we avoid using this class on this component then?
 it is really why css-in-js should be used instead of nested classes :(

wakooka (approved):

src/searchForm/baseStyles.tsx
wakooka: We can and this is what is desribed in the PR description
> Use !important as quick fix, better fix would be not to use nested selectors for styles override, it breaks components isolation and introduce these kinds of errors.

gabrielagabriel: still don't get it

wakooka: We're doing a quick fix for a prod issue, it's probably not the right time to change everything :)

PR #1001, opened 2020-11-25
Fixed aria props getting overriden

wakooka:
## Description

Everytime we were trying to pass `aria-label` to `SlugLink` in `place.href` the prop was getting overriden by the `pickA11yProps`. See here for more context: https://github.com/blablacar/kairos/pull/4487#discussion_r529610375
~We also reproduced it with `ItemChoice` `href` prop.~

As a long term fix we might want to replace `a11yProps` with `React.AriaAttributes` instead

@PerrineBoissieres We need your input on this.

## What has been done

Moved a11y props first so it doesn't override the aria-label passed 

## Things to consider

This is a quick fix, I don't think other things will breaki

## How it was tested

~In storybook by passing a component with `aria-label` in `ItemChoice` `href` prop.~
Unit test + Storybook

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,10 @@
 # Unreleased
 
-- [...]
+- **[FIX]** Stopped `aria-*` props from getting overriden in `Item`
+
 # v41.9.2 (26/11/2020)
 
-- **[UPDATE]** Add prop `divider` with `DividerPosition` enum to  `Breadcrumb` 
+- **[UPDATE]** Add prop `divider` with `DividerPosition` enum to `Breadcrumb`
 
 # v41.9.1 (25/11/2020)
 
--- a/src/_internals/itineraryLocation/ItineraryLocation.tsx
+++ b/src/_internals/itineraryLocation/ItineraryLocation.tsx
@@ -153,7 +153,7 @@ export const ItineraryLocation = ({
       aria-label={place.stepAriaLabel}
     >
       {renderMeta(place.mainLabel, place.subLabel)}
-      <Component {...hrefProps} aria-label={place.actionAriaLabel}>
+      <Component aria-label={place.actionAriaLabel} {...hrefProps}>
         {hasTime && renderTime(place.isoDate, place.time)}
         <div className="kirk-itineraryLocation-roadContainer" aria-hidden="true">
           <Bullet
--- a/src/_internals/itineraryLocation/ItineraryLocation.unit.tsx
+++ b/src/_internals/itineraryLocation/ItineraryLocation.unit.tsx
@@ -23,14 +23,12 @@ const placeWithProximity: Place = {
 
 const placeWithLink: Place = {
   ...place,
-  actionAriaLabel: 'New page with a map',
-  href: '#test',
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@blablacar/ui-library",
-  "version": "41.9.0",
+  "version": "0.0.0-ed463c5a",
```
pith, line 3: I think it's a miss 😉

src/itinerary/Itinerary.story.mdx
```diff
@@ -6,6 +6,11 @@ import { Itinerary } from './index'
 
 <Meta title="Widgets/Itinerary" />
 
+export const MyComponent = props => {
```
pith, line 9: You might want to rollback this change?

pith (approved):

pith:
I would be nice to add a unit test to avoid regression on this cases.

nickoola (approved):

wakooka:
EDIT: We might not need the change on `ItemChoice` this was because we were trying to pass `aria-label` in `href` props while we could just pass it as a top level prop, e.g.
```
<ItemChoice aria-label="My aria label" href={<SlugLink  ... />}
```
I'll re-request your reviews once I double checked that it's working in SPA

src/_internals/itineraryLocation/ItineraryLocation.tsx
```diff
@@ -107,11 +107,13 @@ export const ItineraryLocation = ({
 
   if (!isSmall && !isEmpty(link) && typeof link !== 'string') {
     Component = link.type
+
     hasChevron = true
     hrefProps = {
       ...link.props,
       className: cc([`${baseClassName}-wrapper`, link.props.className]),
     }
+    console.log(hrefProps)
```
simonrelet, line 116: Probably a left over 😊

- [ ] Remove `console.log()`

simonrelet (approved):

wakooka:
I updated the PR and removed the change on `ItemChoice` I confirm this was not needed I also updated the tests in `ItineraryLocation` making sure that they will break if change the order of 
```
<Component {...hrefProps} aria-label={place.actionAriaLabel}>
```

src/_internals/itineraryLocation/ItineraryLocation.tsx
wakooka: haha just pushed, thanks! :)

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Unreleased
 
-- [...]
+- **[FIX]** Stopped `aria-*` props from getting overriden in `Item`
```
pith, line 3:
```suggestion
- **[FIX]** Stopped `aria-*` props from getting overriden in `ItineraryLocation`
```

pith (approved):

CHANGELOG.md
wakooka: Thanks!

nickoola (approved):

PerrineBoissieres (approved):

PR #1008, opened 2020-12-01
Remove title attr from A11yProps

nickoola:
## Description

Remove `title` attribute from `A11yProps` added in the previous release (v41.10.0).
Some components like `Profile` already use `title` prop but for something else.
So we can't provide the `title` attribute globally for now, it require a global refacto/fix.

### DIFF
--- a/src/hint/__snapshots__/Hint.unit.tsx.snap
+++ b/src/hint/__snapshots__/Hint.unit.tsx.snap
@@ -300,12 +300,10 @@ exports[`Hint Default rendering (above) 1`] = `
 <div
   aria-live="polite"
   className="c0"
-  title="Hint Title"
 >
   <aside
     className="c1 c2 bubble-arrow--above"
     id="kirk-hint-1"
-    title="Hint Title"
   >
     <p>
       <strong>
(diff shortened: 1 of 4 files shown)


src/hint/__snapshots__/Hint.unit.tsx.snap
```diff
@@ -300,12 +300,10 @@ exports[`Hint Default rendering (above) 1`] = `
 <div
   aria-live="polite"
   className="c0"
-  title="Hint Title"
```
nickoola, line 303: Normal behavior as the prop is not provided in A11yProps.

src/itemActionTitle/ItemActionTitle.unit.tsx
```diff
@@ -40,17 +40,6 @@ describe('ItemActionTitle', () => {
       expect(screen.getByRole('button', { name: 'Screen reader content' })).toBeInTheDocument()
     })
 
-    it('Should have a complementary content', () => {
```
nickoola, line 43: We could test this props when globally fixed.

pith (approved):

wakooka (approved):

PR #1011, opened 2020-12-02
Boyscout/fix item action title alignment

nickoola:
## Description

Fix `ItemActionTitle` vertical alignment

**Bug demo**
Capture d’écran 2020-12-01 à 17 40 22

## How it was tested

Locally on Storybook

**Demo with all cases:**
Capture d’écran 2020-12-02 à 08 32 46
Capture d’écran 2020-12-02 à 08 33 23
Capture d’écran 2020-12-02 à 08 34 03
Capture d’écran 2020-12-02 à 08 34 26
Capture d’écran 2020-12-02 à 08 34 49
Capture d’écran 2020-12-02 à 08 35 08

### DIFF
--- a/src/_internals/item/index.tsx
+++ b/src/_internals/item/index.tsx
@@ -32,9 +32,11 @@ const StyledItem = styled(Item)`
 
   &.kirk-item--wrappable {
     flex-flow: wrap;
+    padding-top: ${space.m};
     padding-bottom: ${space.m};
 
     .kirk-item-leftWrapper {
+      padding-top: ${space.m};
       padding-bottom: ${space.m};
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **[FIX]** Fix `ItemActionTitle` vertical alignment
 - [...]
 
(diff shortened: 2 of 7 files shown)


src/_internals/item/index.tsx
```diff
@@ -32,9 +32,11 @@ const StyledItem = styled(Item)`
 
   &.kirk-item--wrappable {
     flex-flow: wrap;
+    padding-top: ${space.m};
```
pith, line 35: I don't really understand this fix. Does it override the item padding?

```
  & {
    position: relative;
    display: flex;
    padding-top: ${space.l};
    padding-bottom: ${space.l};
```

pith (approved):

src/_internals/item/index.tsx
nickoola: Yes as explain during our last call, in case of `isWrappable` the global `16px` padding need to be setted to `8px` vertically and completed by direct child to manage vertical/horizontal case display.

wakooka: I agree, I don't think this this should be changed.

wakooka: I still don't understand why :/

PR #1015, opened 2020-12-04
update title props & layout for ColumnedContentSection

Tuxosaurus:
We're adding a "USP" block on Bus landing pages that is basically the same as what we have on the homepage (a section with 3 columns, each with an icon/image and text).

I ask and we can change the Homepage block with these adjustments too.

Changes:
- icon is smaller (but that's not defined in ui-library), positioned on the left rather than in the center (except on small devices), all media have an added margin-bottom
- subtitle is now a paragraph instead of a heading (seo optimization)
- now using typography instead of Text/Title
- fixed columns gap when displaying less or more than 3 columns (not used but illustrated in the specs)

Screen Shot 2020-12-04 at 11 42 24
Screen Shot 2020-12-04 at 11 42 40

Tested on Firefox Macos, with canary release

### DIFF
--- a/src/layout/section/columnedContentSection/ColumnedContentSection.tsx
+++ b/src/layout/section/columnedContentSection/ColumnedContentSection.tsx
@@ -4,13 +4,14 @@ import { Button, ButtonStatus } from '../../../button'
 import { Column } from '../../../layout/column'
 import { Columns } from '../../../layout/columns'
 import { SectionContentSize } from '../../../layout/section/baseSection'
-import { Text, TextTagType } from '../../../text'
-import { Title } from '../../../title'
+import { TextBody } from '../../../typography/body'
+import { TextDisplay1 } from '../../../typography/display1'
+import { TextTitle } from '../../../typography/title'
 import { StyledColumnedContentSection } from './ColumnedContentSection.style'
 
 export type ColumnedContentSectionProps = Readonly<{
   className?: string
-  title: string
+  title?: string
   topLinkLabel?: string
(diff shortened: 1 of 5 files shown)


wakooka (approved):
Suugestion:
Should we rename `minimum` to `default`?
And `simple with icons` just `with icons`?

src/layout/section/columnedContentSection/ColumnedContentSection.tsx
```diff
@@ -106,12 +107,13 @@ const renderColumnContent = (columnContent: ColumnContent, index: string): JSX.E
   return (
     <Column className="kirk-columned-content-section-column" key={index}>
       {media && renderMedia(media)}
+      <p className="kirk-columned-content-section-subtitle">
```
grosbouddha, line 110: I found this a bit weird, it does look like a title.

grosbouddha (approved):

src/layout/section/columnedContentSection/ColumnedContentSection.tsx
Tuxosaurus: The change of tag is to avoid adding a cluster of headings that might bring more noise to bots than intended.

As for the look, I kept it the same indeed.

Tuxosaurus:
> Suugestion:
> Should we rename `minimum` to `default`?
> And `simple with icons` just `with icons`?

Right, I've added proper stories:
Screen Shot 2020-12-04 at 15 11 28

PR #1019, opened 2020-12-09
[BBC-11430] Handle `defaultValue` update in `SelectField`

mcellier:
## Description

With OneAction & our kairos architecture, we can't be sure that the defaultValue will be correctly set a first render (race condition). Since defaultValue can't be updated after it's first rendering, we should use the `value`. We don't want to control this value from outside the component, so I just did as what we implemented in `TextArea` and `TextField`

## What has been done

Same as `TextArea` & `TextField`: use an internal state to handle value/default value

## Things to consider

We might want to have controlled components someday? But that might require bigger updates in all the codebase

## How it was tested

On kairos

### DIFF
--- a/src/selectField/SelectField.tsx
+++ b/src/selectField/SelectField.tsx
@@ -47,12 +47,21 @@ export const SelectField = React.forwardRef(
     const a11yAttrs = pickA11yProps<SelectFieldProps>(props)
     const baseClassName = 'kirk-selectField'
     const [hasFocus, setFocus] = useState(false)
+    const [previousDefaultValue, setPreviousDefaultValue] = useState(defaultValue)
+    const [currentValue, setCurrentValue] = useState(defaultValue)
 
     useEffect(() => {
       if (ref && !disabled && focus) {
         ref.current.focus()
       }
     }, [disabled, focus])
+
+    useEffect(() => {
+      if (defaultValue !== previousDefaultValue) {
+        setPreviousDefaultValue(defaultValue)
+        setCurrentValue(defaultValue)
+      }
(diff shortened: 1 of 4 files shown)


src/selectField/SelectField.tsx
```diff
@@ -47,13 +47,22 @@ export const SelectField = React.forwardRef(
...
 
     useEffect(() => {
       if (ref && !disabled && focus) {
         ref.current.focus()
       }
     }, [disabled, focus])
 
+    useEffect(() => {
+      if (defaultValue !== previousDefaultValue) {
+        setPreviousDefaultValue(defaultValue)
```
wakooka, line 61: Do we need to have previous default value? Why don't we compare with  `props.defaultValue` with  `currentValue`?

mcellier: If we update the value (by using the form), we might end up overriding it i guess. Except that the useEffect dependencies will prevent this?

wakooka: I'm not sure about your question. But how I understand the code is that we do
```
if (defaultValue !== previousDefaultValue) {
```
but since `previousDefaultValue` and `currentValue` hold the same value I don't see the difference with this:
```
if (defaultValue !== currentValue) {
```

mcellier: Except that `currentValue` is updated on change

wakooka: Thanks that was the bit I was missing 🤦

wakooka (approved):

PerrineBoissieres (approved):

PR #1028, opened 2020-12-21
[BBC-11573] Add placeholder prop to UneditableTextField

Tip-Sy:
## Description

This prop makes the text light grey.
The idea is to remove custom CSS from Kairos that does that.

## What has been done

Add placeholder prop to UneditableTextField

Capture d’écran 2020-12-21 à 14 53 17

## How it was tested

Locally + UT

### DIFF
--- a/src/uneditableTextField/UneditableTextField.style.tsx
+++ b/src/uneditableTextField/UneditableTextField.style.tsx
@@ -27,4 +27,8 @@ export const StyledUneditableTextField = styled.div`
     text-overflow: ellipsis;
     overflow: hidden;
   }
+
+  & .kirk-uneditableTextField-label--placeholder {
+    color: ${color.lightMidnightGreen};
+  }
 `
--- a/src/uneditableTextField/UneditableTextField.tsx
+++ b/src/uneditableTextField/UneditableTextField.tsx
@@ -10,6 +10,7 @@ export type UneditableTextFieldProps = Readonly<{
   addOn?: JSX.Element
   href?: JSX.Element | string
   ellipsis?: boolean
+  isPlaceholder?: boolean
 }>
 
 export const UneditableTextField = ({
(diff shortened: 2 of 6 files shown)


src/uneditableTextField/UneditableTextField.style.tsx
```diff
@@ -27,4 +27,8 @@ export const StyledUneditableTextField = styled.div`
     text-overflow: ellipsis;
     overflow: hidden;
   }
+
+  & .kirk-uneditableTextField-label--placeholder {
```
gabrielagabriel, line 31:
```suggestion
  & ::placeholder { 
   color: ${color.lightMidnightGreen};
  }
```

src/uneditableTextField/UneditableTextField.tsx
```diff
@@ -10,13 +10,15 @@ export type UneditableTextFieldProps = Readonly<{
   addOn?: JSX.Element
   href?: JSX.Element | string
   ellipsis?: boolean
+  isPlaceholder?: boolean
```
gabrielagabriel, line 13: I _strongly_ suggest that you use `placeholder` native attribute from HTML instead of creating a prop.
By passing a placeholder prop, it can be styled with CSS
See https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder

Tip-Sy: Unfortunately, the generated HTML is a `div`, so `::placeholder` is not applicable in this case 😬 
But as discussed with Vince, this component should be burnt, let's tackle that during Chapter!
🔥 
It could simply be a button with a special skin 🤷

grosbouddha: +1000 on removing this horrible component one day, or renaming it to something like FakeTextFieldButton (it's a button, not a textfield)

grosbouddha (approved):

src/uneditableTextField/UneditableTextField.tsx
gabrielagabriel: 😱

PR #1033, opened 2021-01-05
Create new ThemeProvider + CSS Reset/fonts

grosbouddha:
## Description
- Create new ThemeProvider and put inside the default CSS reset and typography from the main application (from Layout/style.css). Relative links to font assets were replaced by absolute links to the same assets uploaded on our GCS cdn. 
- Use this ThemeProvider in all stories by enhancing the preview generator
- Remove obsolete CSS reset and binary fonts from storybook (they were used to try to mimic the reset/fonts from the main app but are not needed anymore since we use themeProvider everywhere).

Note: The ThemeProvider is not configurable for now and only provide the default Pixar theme. Some theming logic can be easily added later (and hooks to retrieve the theme from nested children). The goal for this PR is to have a placeholder for all global CSS (reset and fonts) and to remove duplcate reset/fonts from the storybook.

Next steps:
 - Use it in main app and remove all redundant CSS reset and font
 - enjoy

## How it was tested

Locally


src/_utils/themeProvider/index.tsx
```diff
@@ -0,0 +1,203 @@
+import React, { ReactNode } from 'react'
+import { createGlobalStyle } from 'styled-components'
+
+import { color } from '../branding'
+
+const CDN_BASE_URL = 'https://cdn.blablacar.com/kairos/assets/fonts'
```
gabrielagabriel, line 6: should go to global ../branding file since we could also use other assents from there.
Also this is FONT_CDN_BASE_URL

.storybook/preview.css
```diff
@@ -0,0 +1 @@
+/* Add story preview CSS here  */
```
gabrielagabriel, line 1: I don't need it's necessary to keep this file.. we can delete it

gabrielagabriel (approved):
🥇  You rock! Thank you so much!

src/_utils/themeProvider/index.tsx
guillaumewuip, line 6:
```suggestion
const CDN_FONTS_BASE_URL = 'https://cdn.blablacar.com/kairos/assets/fonts'
```

Because of `assets/fonts'`, CDN could be use for other things in the future

guillaumewuip (approved):

wakooka (approved):
Wow! 👏

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - **[UPDATE]** Add missing exports from `PhoneField`
+- **[UPDATE]** Create new `ThemeProvider` with base app CSS reset/fonts and use it for all stories
```
pith, line 4: Maybe **NEW** ?
```suggestion
- **[NEW]** Create new `ThemeProvider` with base app CSS reset/fonts and use it for all stories
```

pith (approved):
Thanks for this PR! 
It would be nice to add this `ThemeProvider` in a "getting started" section of the documentation.

src/_utils/themeProvider/index.tsx
grosbouddha: Not completely convinced (I like to keep things local as long as possible, here it's only used in this file).
But it's easy to move around so I moved it for now.

.storybook/preview.css
grosbouddha: Removed

src/_utils/themeProvider/index.tsx
grosbouddha: Done

PR #1038, opened 2021-01-08
Update CardsSection and Add new CardsGridSection

Tuxosaurus:
To be able to show sample trips on Bus landing pages, we need a new layout section to handle multiple TripCards.
I've named it `CardsGridSection` as it is based on the existing `CardsSection`.

It behaves the same on small devices but on large devices it follows a grid approach:
Screen Shot 2021-01-08 at 11 36 57

The section also features the ability to have a title on top and a button on the bottom.

At first I tried to simply update the `CardsSection` component but it was too messy so I turned it into its own thing.
Still, I need it on a "Large" content sized section and I saw that `CardsSection` did have a `max-width` prop that wasn't used so I changed it to `contentSize` here too so it's compliant with the rest of the layout sections. 

Tested on Firefox MacOS on local and with a canary release.


src/layout/section/cardsGridSection/CardsGridSection.style.tsx
```diff
@@ -0,0 +1,102 @@
...
+  & .kirk-card:last-child::after {
+    content: '';
+    display: block;
+    position: absolute;
+    right: -${wrapperHorizontalPadding};
+    width: ${wrapperHorizontalPadding};
+    height: 1px;
+  }
+
+  @media (${responsiveBreakpoints.isMediaLarge}) {
```
gabrielagabriel, line 64: nit: By declaring breakpoints by a component we easily identify its "global" behavior 

for exemple: 

```suggestion
  & .kirk-cardsGridSection {
    // Mobile first behavior
    
     @media (${responsiveBreakpoints.isMediaLarge}) {
        // Large behavior
     }
}
  
```

src/layout/section/cardsGridSection/CardsGridSection.style.tsx
```diff
@@ -0,0 +1,102 @@
...
+    padding: ${space.l} 0;
+    box-sizing: border-box;
+    overflow: auto; /* Make Cards scrollable horizontally */
+    -ms-overflow-style: none; /* Remove scrollbar visually IE 10+ */
+    scrollbar-width: none; /* Remove scrollbar visually Firefox */
+    scroll-snap-type: x mandatory;
+    scroll-padding: ${space.xl};
+  }
+
+  & .kirk-cardsGridSection .kirk-card {
```
gabrielagabriel, line 31: If Card style needs to be redefined, it rather use styled-components syntaxe to avoid classes conflict

```suggestion
import TripCard from "../tripCard"
const Card = styled(TripCard)`
// redefined styles goes here
`
```

src/layout/section/cardsGridSection/CardsGridSection.tsx
```diff
@@ -0,0 +1,56 @@
...
+      'kirk-cardsGridSection-wrapper',
+      {
+        'kirk-cardsGridSection-wrapper--large': contentSize === SectionContentSize.LARGE,
+      },
+      className,
+    ])}
+  >
+    {title && (
+      <BaseSection contentSize={contentSize}>
+        <h2 className="kirk-cardsGridSection-title">
```
gabrielagabriel, line 41: You can convert all this on a styled-component

```suggestion
      const Title =   styled(TextDisplay1)
```

src/layout/section/cardsGridSection/CardsGridSection.tsx
```diff
@@ -0,0 +1,56 @@
...
+    className={cc([
+      'kirk-cardsGridSection-wrapper',
+      {
+        'kirk-cardsGridSection-wrapper--large': contentSize === SectionContentSize.LARGE,
+      },
+      className,
+    ])}
+  >
+    {title && (
+      <BaseSection contentSize={contentSize}>
```
gabrielagabriel, line 40: Do we need to use BaseSection here? or could we control this behavior on CSS?

src/layout/section/cardsSection/CardsSection.tsx
```diff
@@ -1,20 +1,34 @@
...
+export const CardsSection = ({
+  children,
+  className = '',
+  contentSize = SectionContentSize.SMALL,
+}: CardsSectionProps) => (
+  <StyledCardsSection
+    className={cc([
+      'kirk-cardsSection-wrapper',
+      {
+        'kirk-cardsSection-wrapper--large': contentSize === SectionContentSize.LARGE,
```
gabrielagabriel, line 27: should not be part of the API since we are able to retrieve that with CSS

gabrielagabriel (approved):
🎉 Great job. It will inspire me. I have the very same problem to solve with Highlight Section:
Title, Grid of 3 and Link/Button... 

We do need layout specs :)

src/layout/section/cardsGridSection/CardsGridSection.tsx
Tuxosaurus: I think it's better if I keep the BaseSection, to make sure I reproduce the same layout as intended, rather than recreate that layout in CSS just for that. It's all about having the same width and padding, which is what the BaseSection does.

src/layout/section/cardsGridSection/CardsGridSection.style.tsx
Tuxosaurus: TripCard items are passed through the children prop so I am not sure I can have them replaced by a styled version.

src/layout/section/cardsGridSection/CardsGridSection.style.tsx
Tuxosaurus: I'll remove the media rule for `small` and make it the default, and keep only the rule for media `large`.
So all the behavior for `large` is referenced at the same level.

wakooka (approved):

src/layout/section/cardsGridSection/CardsGridSection.style.tsx
gabrielagabriel: Indeed you would to export the styled version instead of the normal one and apply it

src/layout/section/cardsSection/CardsSection.tsx
Tuxosaurus: I'm not sure I understand how we can retrieve it with CSS? It's not the breakpoint value

src/layout/section/cardsGridSection/CardsGridSection.tsx
Tuxosaurus: I will remove the className and use `styled(TextDisplay1)` instead but I can't remove the `h2`.

PR #1044, opened 2021-01-12
[BOYSCOUT] Fix itinerary headline display

mcellier:
## Description

Fix itinerary headline broken since `SubHeader` has been normalized

## What has been done

Stop using `SubHeader` as it comes with its default margin/padding that'll have to be overriden in this component.
Used `typography/SubHeaderStrong` instead, to keep consistent on the typography side. Added a h2 to keep the layout similar without breaking SEO, and re-added margins.

## Things to consider

We should use typography components over "exposed" components to avoid these issues

## How it was tested

storybook

### DIFF
--- a/src/itinerary/Itinerary.style.tsx
+++ b/src/itinerary/Itinerary.style.tsx
@@ -20,6 +20,11 @@ export const StyledItinerary = styled.div`
     padding-top: ${space.m};
   }
 
+  & .kirk-itinerary--headline {
+    padding-top: ${space.xl};
+    padding-bottom: ${space.m};
+  }
+
   & .kirk-itinerary--highlightRoad .kirk-itineraryLocation-road {
     background-color: ${color.midnightGreen};
   }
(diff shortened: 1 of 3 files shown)


src/itinerary/Itinerary.style.tsx
```diff
@@ -20,6 +20,11 @@ export const StyledItinerary = styled.div`
     padding-top: ${space.m};
   }
 
+  & .kirk-itinerary--headline {
```
Tuxosaurus, line 23: nit: to me, the "--" syntax is for modifiers only but here it targets the headline so I'd write `kirk-itinerary-headline`
I'm not sure about the `--highlightRoad` that follows so no big deal.

Tuxosaurus (approved):

wakooka (approved):

src/itinerary/Itinerary.style.tsx
mcellier: Let's change this when we'll review this component following @wakooka design doc

PR #1051, opened 2021-01-19
Add background variation to CheckCircleIcon

mcellier:
## Description

Capture d’écran 2021-01-19 à 16 29 29

## What has been done

Added `background` prop to the icon. The name is chosen to match the pixar spec ("Check Circle + background")

## How it was tested

Kairos

### DIFF
--- a/src/icon/checkCircleIcon.tsx
+++ b/src/icon/checkCircleIcon.tsx
@@ -4,13 +4,29 @@ import styled from 'styled-components'
 import { color } from '../_utils/branding'
 import { BaseIcon, BaseIconDefaultProps, Icon } from '../_utils/icon'
 
-const CheckCircleIcon = (props: Icon) => (
+export type CheckCircleIconProps = Icon &
+  Readonly<{
+    background?: boolean
+  }>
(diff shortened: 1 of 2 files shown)


src/icon/checkCircleIcon.tsx
```diff
@@ -4,25 +4,46 @@ import styled from 'styled-components'
...
   </BaseIcon>
 )
 
 const StyledCheckCircleIcon = styled(CheckCircleIcon)`
   path {
     fill: ${props => (props.iconColor ? props.iconColor : color.green)};
   }
+
+  path:nth-child(2) {
+    fill: ${color.white};
```
wakooka, line 39: nit: Not fan of this selector, add it as an attribute directly since this value should never change?

wakooka (approved):

src/icon/checkCircleIcon.tsx
mcellier: Not working as an attribute, it's overriten by the styles.

pith (approved):

PR #1057, opened 2021-01-25
Fix `ItemChoice` addons types

mcellier:
## Description

Now matching `Item` types and allow strings

### DIFF
--- a/src/itemChoice/ItemChoice.tsx
+++ b/src/itemChoice/ItemChoice.tsx
@@ -24,8 +24,8 @@ export type ItemChoiceProps = A11yProps &
     labelInfo?: React.ReactNode
     data?: string
     dataInfo?: string
-    leftAddon?: JSX.Element
-    rightAddon?: JSX.Element
+    leftAddon?: React.ReactNode
+    rightAddon?: React.ReactNode
     className?: string
     href?: string | JSX.Element
     status?: ItemStatus
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **[FIX]** Fix `leftAddon` and `rightAddon` types on `ItemChoice` to allow strings
 [...]
 
 # v45.1.0 (20/01/2021)


src/itemChoice/ItemChoice.tsx
```diff
@@ -24,8 +24,8 @@ export type ItemChoiceProps = A11yProps &
     labelInfo?: React.ReactNode
     data?: string
     dataInfo?: string
-    leftAddon?: JSX.Element
-    rightAddon?: JSX.Element
+    leftAddon?: React.ReactNode
```
gabrielagabriel, line 27: ❤️

gabrielagabriel (approved):

Tuxosaurus (approved):

src/itemChoice/ItemChoice.tsx
wakooka: Should you use `ItemProps['leftAddon']`? I don't know actually know what the best practice is here, this will avoid to repeat ourselves for a prop that only passes through to `Item`. WDYT?

grosbouddha: Let's use a dedicated type maybe (e.g. type LeftAddonType) if we really don't want to duplicate (I found ItemProps['leftAddon'] not really readable).
To be fair, since this ReactNode is internal to this file and the file is still small, i don't feel the urge to create this abstract type for now to reduce this small duplication. Your call folks.

grosbouddha (approved):

src/itemChoice/ItemChoice.tsx
mcellier: IMHO, in the end (it doesn't even matter 🎵 ), we should probably just extend parts of the base `Item` typings. But that's part of another refacto I guess

wakooka: Thanks for the answers 👌  it was more to open the discussion rather than a comment specific to this PR because I see this pattern coming back again a lot.

PR #1056, opened 2021-01-22
[PROPIM-558] NEW HighlightSection

gabrielagabriel:
## Description

screencapture-localhost-4000-2021-01-22-12_55_21

## What has been done

- Add attribute/props `hidden` instead of `class` to hidden an `Item`
- Use `styled-components` syntaxe

## Things to consider

This is a DRAFT
Missing:
- RTL

PR #1062, opened 2021-01-29
[BBC-12076] SpecialOffer component

guillaumewuip:
> BBC-12076 

## Description

This PR add a `SpecialOffer` component to kirk for the pixarisation of the Good Deal page.

There is no special Figma for this component at the moment (Jonas seems to be against considering a component for Pixar if it's not used for the 3 platforms on day one) but you can find the Figma for the Good Deal project here: https://www.figma.com/file/zX4OiDroAATUOJ4yl4ZEuO/Pixarization-final?node-id=0%3A4248

Here are some example of the variations of the component:

On desktop:
image

On small screens:
image

What we're sure is that the component should have: 
- a picture (url and alt)
- a title
- a description
- 0 or 1 "info" links
- 0, 1 or 2 Call to Action (they can be links or buttons for custom registration methods like Vinci and Fishka)

Still WIP

## DEMO

Here used inside a `BaseSection` with a divider:

Screenshot 2021-02-01 at 11 12 57

## TESTED
 
UT


src/specialOffer/SpecialOffer.story.mdx
```diff
@@ -0,0 +1,101 @@
+import { Meta, Story, Canvas, ArgsTable } from '@storybook/addon-docs/blocks'
+
+import { SpecialOffer } from './index'
+
+<Meta title="SpecialOffer" />
+
+# **SpecialOffer**
```
guillaumewuip, line 7: Below are three possible APIs for the component: 

1) Compound version
In my mind this is the version that offer the better dev experience and the right amount of flexibility for the future (no fear of having a enormous list of props to maintain in 6 months to handle special conditions). But this is also the one that offer the less security (can the dev add multiple Picture ? No LinkCTA at all ?)

2) full props version
This is the full opposite of 1). Very secured from a Typescript point of view. We can even type the precise amount of links / infos we allow.

3) Mix compound / props version
I tried to propose a "take all advantages of the two others options" option but it's maybe "all the disadvantages" 🙃 


What do you think? I find it hard to choose between the options.
Also don't hesitate to propose better names!

src/specialOffer/SpecialOffer.story.mdx
```diff
@@ -0,0 +1,101 @@
...
+      <SpecialOffer.LinkCTA
+        label="Register for option 1!"
+        url="http://example.com/register1"
+      />
+      <SpecialOffer.LinkCTA
+        label="Register for option 2!"
+        url="http://example.com/register2"
+      />
+    <SpecialOffer/>
+  </Story>
```
Tip-Sy, line 88: I prefer this option personally

src/specialOffer/SpecialOffer.story.mdx
grosbouddha: I vote for 2. Compound APIs are falsely generic (but much nicer looking) and you need to write a bunch of code to assert the non-explicit contract (or to manage all the cases if you allow every combinations which seems a bad idea :) ).

guillaumewuip: Yep. I agree with the problem you just exposed and I think I would vote for option 2 too for such a small component

gabrielagabriel: I'm kinda lost since I have the very same issue on the `SEA Landing Page Grid`, as we discussed this morning.
What I understand of these components that **are not PIXAR** is that they don't need flexibility because they aren't meant to be reused in _other_ cases. cf this PR https://github.com/blablacar/ui-library/pull/1056
So, I may **vote for 2**. I'll keep Compound Way just for styled-components elements and for the exposure of the component itself, typescript. Sounds like the better of the two world too hehe

guillaumewuip: Ok let's go for option 2 ! 🚀

src/specialOffer/SpecialOffer.tsx
```diff
@@ -0,0 +1,94 @@
...
+  }
+>
+export type SpecialOfferButtonsProps = Readonly<
+  SpecialOfferBaseProps & {
+    ctaButtons: CTAButtons
+  }
+>
+
+function SpecialOffer(props: SpecialOfferLinksProps): JSX.Element
+function SpecialOffer(props: SpecialOfferButtonsProps): JSX.Element
```
guillaumewuip, line 47: [Functions overload](https://krzysztofzuraw.com/blog/2019/typescript-function-overloads) to discriminate the two cases: 
- CTAs as links
- CTAs as buttons

src/specialOffer/SpecialOffer.tsx
```diff
@@ -0,0 +1,94 @@
...
+export type SpecialOfferLinksProps = Readonly<
+  SpecialOfferBaseProps & {
+    ctaLinks: CTALinks
+  }
+>
+export type SpecialOfferButtonsProps = Readonly<
+  SpecialOfferBaseProps & {
+    ctaButtons: CTAButtons
+  }
+>
```
guillaumewuip, line 44: Maybe you have better names in mind instead of `ctaLinks` and `ctaButtons` ?

mcellier:
> There is no special Figma for this component at the moment (Jonas seems to be against considering a component for Pixar if it's not used for the 3 platforms on day one)

You might put it with our sections components. IMHO the src/ main directory is more for "basic" components

guillaumewuip:
I'm kind of lost:
- I find moving the component inside `layout/section` a bit strange as it's bit a layout nor a component that display a list of things
- there is a (partially) similar looking component in `layout/section/columns` (thanks @mcellier for the hint) but it can't be reused like that (miss at least links and buttons)
  
  image
  
In my ideal wold there would be a generic internal component for those type of cards that can be reused to produce more high-level components. Those components would be used inside layout in a compounded manner. 

What do you think it's best to do in this particular situation for Special Offer @gabrielagabriel @grosbouddha @Tip-Sy @mcellier ?

I vote for keeping this high level component, but I'm not sure about moving it inside `layout/something`

src/specialOffer/SpecialOffer.story.mdx
```diff
@@ -0,0 +1,198 @@
...
+              label: 'Register for option 1!',
+              url: 'http://example.com/register1',
+            },
+            {
+              label: 'Register for option 2!',
+              url: 'http://example.com/register2',
+            },
+          ]}
+        />
+        <ContentDivider />
```
gabrielagabriel, line 39: Would be better to split into 2 Stories if you want to show different cases

src/specialOffer/SpecialOffer.story.mdx
```diff
@@ -0,0 +1,198 @@
...
+          ]}
+        />
+      </ItemsList>
+    </BaseSection>
+  </Story>
+</Canvas>
+
+
+## Specifications
+
```
gabrielagabriel, line 61: Do you mind explaining the use case here.

src/specialOffer/SpecialOffer.style.tsx
```diff
@@ -0,0 +1,15 @@
+import styled from 'styled-components'
+
+import { space } from '../_utils/branding'
+import { normalizeHorizontally } from '../layout/layoutNormalizer'
+
+export const StyledSpecialOffer = styled.div``
+
+export const StyledSpecialOfferPicture = styled.img`
+  width: 70%;
```
gabrielagabriel, line 9: why?! hehehe

src/specialOffer/SpecialOffer.tsx
gabrielagabriel: Yeah.. Both are Actions! the only thing that differ is `href` or `onclick` . I would join this API and resolve the element to be used inside the component instead of exposing it to the API

src/specialOffer/SpecialOffer.tsx
```diff
@@ -0,0 +1,94 @@
...
+  ctaButtons,
+}: SpecialOfferBaseProps &
+  Readonly<{
+    ctaLinks?: CTALinks
+    ctaButtons?: CTAButtons
+  }>) {
+  return (
+    <StyledSpecialOffer>
+      <StyledSpecialOfferPicture src={picture.url} alt={picture.alt} />
+      <SubHeader>{title}</SubHeader>
```
gabrielagabriel, line 63: Make sure title is available to avoid rendering a empty element

src/specialOffer/SpecialOffer.tsx
```diff
@@ -0,0 +1,94 @@
...
+      <Paragraph>{description}</Paragraph>
+      {informationLink && (
+        <Disclaimer useInfoIcon>
+          <Button status={ButtonStatus.UNSTYLED} href={informationLink.url}>
+            {informationLink.label}
+          </Button>
+        </Disclaimer>
+      )}
+      {ctaLinks && (
```
gabrielagabriel, line 72: you could have a function to resolve that outside the render method

src/specialOffer/SpecialOffer.tsx
guillaumewuip: Typescript is here to secure that. Adding a special check (`if` of `&&`) would make the code more complex without adding any more security, right?

src/specialOffer/SpecialOffer.story.mdx
```diff
@@ -0,0 +1,198 @@
...
+import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs/blocks'
+
+import { ContentDivider } from '../divider/contentDivider'
+import { ItemsList } from '../itemsList'
+import { BaseSection } from '../layout/section/baseSection'
+import { SpecialOffer } from './index'
+
+<Meta title="Widgets/SpecialOffer" />
+
+# **SpecialOffer**
```
gabrielagabriel, line 11: Hey, I was thinking that this component looks a lot like a Thumbnails/ Card pattern component. Do you think we could rename it ?!

https://getbootstrap.com/docs/5.0/components/card/
https://getbootstrap.com/2.2.0/components.html#thumbnails

src/specialOffer/SpecialOffer.style.tsx
guillaumewuip: It's approximately the ratio I see on the [Figma](https://www.figma.com/file/zX4OiDroAATUOJ4yl4ZEuO/Pixarization-final?node-id=1763%3A0). 

Why it the best way to handle this value?

src/specialOffer/SpecialOffer.story.mdx
guillaumewuip: The internals look very similar but I'm not confortable renaming it.

In my mind a good Thumbnails/Card component is generic enough to be used in all situations with slots/placeholder that are agnostics of their real content.

This is not the situation here. You can see with the typings of the component that it is very locked to the need of the Good Deals page (0 or 1 informationLink), picture cannot be a `JSX.Element`, 1 or 2 links or 1 or 2 buttons, special size for the image. Those are specificities that we don't want for a generic Thumbnails/Card component I think.

I'm not saying a Thumbnails/Card component is a bad idea but that maybe there is two levels?
- a generic Thumbnails/Card component that we can use here and elsewhere. This one would be internal to kirk.
- more high level exposed components that embed business rules but used the generic component internally

What do you think? I was talking with Michael about this this morning and my feeling is that it's not a straightforward idea for the SPA team. Maybe kirk is not thought this way at all and that I'm not aligned with what it should expose.

src/specialOffer/SpecialOffer.style.tsx
```diff
@@ -0,0 +1,15 @@
+import styled from 'styled-components'
+
+import { space } from '../_utils/branding'
+import { normalizeHorizontally } from '../layout/layoutNormalizer'
+
+export const StyledSpecialOffer = styled.div``
+
+export const StyledSpecialOfferPicture = styled.img`
```
gabrielagabriel, line 8: you could use `figure` to wrap the image itself

src/specialOffer/SpecialOffer.tsx
guillaumewuip: I'm not sure of seeing what you mean here. Do you think of the function called by the render method that, giving the full props, will returns links or buttons?

grosbouddha:
From the description:

> 0 or more "info" links

it should be '0 or 1 "info" link' ?

src/specialOffer/SpecialOffer.style.tsx
grosbouddha: The percentage-based value will look bad on very large screens.
Should we combine it with a max-width: Xpx (maybe 400)?

guillaumewuip:
> From the PR description:
> 
> > 0 or more "info" links
> 
> it should be '0 or 1 "info" link' ?

yep sorry, error on my side

src/specialOffer/SpecialOffer.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { Paragraph } from '../paragraph'
+import { SubHeader } from '../subHeader'
+import { StyledSpecialOffer, StyledSpecialOfferPicture } from './SpecialOffer.style'
+
+type Picture = Readonly<{
+  url: string
+  alt: string
+}>
+
+type Link = Readonly<{
```
grosbouddha, line 15: i would merge Link and Button to simplify the code (maybe ActionButton as a tentative name)
```
type ActionButton = {
  label: string
  url?: string
  onClick?: Function
}
```
As well, you might want to have both onClick and href at the same time one day or you might want to mix Link/Button cases. So writing more code to separate them sounds useless and less flexible.
This will divide by 2 the template code to render this and remove the unusual method overload to trigger the correct rendering.
As well, having actions with both optional href/onClick pattern is used in other places in the code.

I would as well remove the constraint on the array size (1 or 2), and just take an array of any size, render them all. That would be the responsibility of the caller to give the right amount; the component can render 0 or 40 ActionButtons.

grosbouddha:
> I'm kind of lost:
> 
> * I find moving the component inside `layout/section` a bit strange as it's bit a layout nor a component that display a list of things
> * there is a (partially) similar looking component in `layout/section/columns` (thanks @mcellier for the hint) but it can't be reused like that (miss at least links and buttons)
>   image
> 
> In my ideal wold there would be a generic internal component for those type of cards that can be reused to produce more high-level components. Those components would be used inside layout in a compounded manner.
> 
> What do you think it's best to do in this particular situation for Special Offer @gabrielagabriel @grosbouddha @Tip-Sy @mcellier ?
> 
> I vote for keeping this high level component, but I'm not sure about moving it inside `layout/something`

I would keep it as a component too, not a layout item: You still need to put it inside a layout section.

src/specialOffer/SpecialOffer.tsx
guillaumewuip: > i would merge Link and Button to simplify the code (maybe ActionButton as a tentative name)
> 
> ```
> type ActionButton = {
>   label: string
>   url?: string
>   onClick?: Function
> }
> ```

I'm not confortable allowing `{ label }` as a valid action. So maybe this type?

```
type ActionButton = {
  label: string
  url: string
} | {
  label: string
  onClick: Function
} | {
  label: string
  url: string
  onClick: Function
}
```

I'm still not confortable with this type to merge both link and button as it's complex and allow something we don't need for now (`url` and `onClick` at the same time) but we will have to maintain. I was surprised we can use a Button as a link in the first place actually.

> As well, you might want to have both onClick and href at the same time one day or you might want to mix Link/Button cases. 

Ok. I'm more into "small API first, open the API after if needed". As both url + onClick combo and mixing Link/Button are not a current need, I see them as *bugs* that Typescript can prevent. 

> I would as well remove the constraint on the array size (1 or 2), and just take an array of any size, render them all. That would be the responsibility of the caller to give the right amount; the component can render 0 or 40 ActionButtons.

Idem, those contraints have been discussed with Heloise and Naima so my position is to embed them in code, for the knowledge to written clearly and not forgotten or bypassed by the caller. The "small API first, open the API after if needed" will allow us to open for more cases if needed for free.

What do you think? Have you a strong opinion for this `ActionButton`?

grosbouddha: I don't feel strongly about these; let's keep the constraints level high, in particular if you're confident that it should not move too much.

grosbouddha (approved):

src/specialOffer/SpecialOffer.tsx
Tip-Sy: +1 for the factorization

I like it this way, simple enough IMHO:
```
type ActionButton = {
  label: string
  url?: string
  onClick?: Function
}
```
Which would give a valid button (with no action, I admit it's strange but at least there is no error/crash)
```
            <Button key={button.label}>
              {button.label}
            </Button>
```

Then if you want to enforce the type with the complete API why not, it's also fine, as you wish 👍

Tip-Sy (approved):

src/specialOffer/SpecialOffer.tsx
guillaumewuip: I would stay with the link/button distinction as I'm not confortable allowing users of the component to create something broken

src/specialOffer/SpecialOffer.tsx
gabrielagabriel: that means title is **_required_** than

src/specialOffer/SpecialOffer.story.mdx
gabrielagabriel: may we take that to chapter / kick discution since we have a lot of people involved?

src/specialOffer/SpecialOffer.tsx
gabrielagabriel: Since I've proposed in another thread I also vote for it :) hihiihihi

src/specialOffer/SpecialOffer.tsx
guillaumewuip: Yes it is, as typed here: https://github.com/blablacar/ui-library/blob/59e96c6cea8051973c77aedc5af820ef6ee3e41d/src/specialOffer/SpecialOffer.tsx#L10-L44

Everything is required appart from `informationLink`

src/specialOffer/SpecialOffer.tsx
guillaumewuip: Ok I propose  this

```suggestion
export type SpecialOfferLinksProps = Readonly<
  SpecialOfferBaseProps & {
    actions: CTALinks | CTAButtons
  }
>
```

but keeping the `CTALinks` and `CTAButtons` separate types

src/specialOffer/SpecialOffer.story.mdx
guillaumewuip: Okey, I will remove `BaseSection` to have only `SpecialOffer` here

src/specialOffer/SpecialOffer.style.tsx
guillaumewuip: I've used 

```css
  width: 70%;
  max-width: 400px;
```

let me know if it's OK for you

src/specialOffer/SpecialOffer.story.mdx
guillaumewuip: I've added a few words. Is it OK for you?

src/specialOffer/SpecialOffer.tsx
guillaumewuip: I think it's not necessary anymore with the `actions` renaming there is only one case to handle

src/specialOffer/SpecialOffer.story.mdx
guillaumewuip: Yup I think it's an important discussion but one that couldn't be just be done in a meeting. The best way to do in mind mind would be to open an ADR document for review.

src/specialOffer/SpecialOffer.tsx
guillaumewuip: > Since I've proposed in another thread I also vote for it :) hihiihihi

Yup. Actually I see two points in this discussion: the renaming of `ctaLinks` and `ctaButtons` to `actions` (done, thanks!) and the merge of the two types (I prefer to not introduce a type that allow something broken). I think we came up with a good tradeoff, thanks everyone!

src/specialOffer/SpecialOffer.style.tsx
guillaumewuip: Done, thanks!

src/specialOffer/SpecialOffer.story.mdx
```diff
@@ -0,0 +1,207 @@
...
+<Canvas>
+  <Story name="no CTAs">
+    <SpecialOffer
+      picture={{
+        url: 'https://cdn.blablacar.com/kairos/assets/build/images/blablasure_logo-459e4b952a6cc04b72ea0e12f98ec673.svg',
+        alt: 'The BlaBlaSur offer logo',
+      }}
+      title="A wonderful offer"
+      description="Once you receive your first rating, receive a 30€ voucher on Total spring green electricity offer for your home. So come back here to enjoy it !"
+      actions={[]}
```
guillaumewuip, line 72: I'm forced to use an empty `[]` instead of not providing the `actions` prop here because storybook replace the `undefined` value by a function from `addon-actions` 🤷‍♂️ I haven't found how to disable this.

src/specialOffer/SpecialOffer.style.tsx
```diff
@@ -0,0 +1,47 @@
...
+  max-width: 50%;
+
+  @media (${responsiveBreakpoints.isMediaSmall}) {
+    max-width: 100%;
+    min-width: 100%;
+  }
+`
+
+export const StyledSpecialOfferFigure = styled.figure`
+  width: 70%;
```
guillaumewuip, line 23: No need for a `max-width` anymore for desktop size screens because we switch to a 2-columns layout in that case.

src/specialOffer/SpecialOffer.style.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export const StyledSpecialOfferContent = styled.div`
+  flex: 1;
+  align-items: flex-start;
+  max-width: 50%;
+
+  @media (${responsiveBreakpoints.isMediaSmall}) {
+    max-width: 100%;
+    min-width: 100%;
+  }
```
guillaumewuip, line 19: I'm not sure it's the best way to do this. Don't hesitate to propose something else here 😁

src/specialOffer/SpecialOffer.tsx
```diff
@@ -0,0 +1,80 @@
...
+        )}
+      </StyledSpecialOfferText>
+      <StyledSpecialOfferButtons>
+        {(actions as Array<Link | Button>).map(action => (
+          <StyledSpecialOfferButtonLine key={action.label}>
+            {'onClick' in action ? (
+              <Button onClick={action.onClick}>{action.label}</Button>
+            ) : (
+              <Button href={action.url}>{action.label}</Button>
+            )}
```
Tip-Sy, line 74: I like that 👍 Good tradeoff indeed

src/specialOffer/SpecialOffer.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+type CTALinks = Array<Link>
+type CTAButtons = Array<Button>
+
+export type SpecialOfferProps = Readonly<{
+  picture: Picture
+  title: string
+  description: string
+  informationLink?: Link
+  actions?: CTALinks | CTAButtons
```
Tip-Sy, line 39: I'm wondering: there is no way to mix up links + buttons in the actions?
I don't know if we want to add this flexibility but you could remove your `CTALinks` & `CTAButtons` types and do this:
```suggestion
  actions?: Array<Link | Button>
```
But maybe you force link-only or button-only actions on purpose, if so you can ignore my comment 👍

src/specialOffer/SpecialOffer.unit.tsx
```diff
@@ -0,0 +1,124 @@
+import React from 'react'
+
+import { fireEvent, render, screen } from '@testing-library/react'
+
+import { SpecialOffer } from './SpecialOffer'
+
+const picture = {
+  url:
+    'https://cdn.blablacar.com/kairos/assets/build/images/blablasure_logo-459e4b952a6cc04b72ea0e12f98ec673.svg',
+  alt: 'The BlaBlaSur offer logo',
```
Tip-Sy, line 10: nit: `BlaBlaSure`*

src/specialOffer/SpecialOffer.style.tsx
Tip-Sy: I'm no expert but I would have done it like this as well 😛

src/specialOffer/SpecialOffer.style.tsx
Tip-Sy: Is this "magic value" coming from somewhere else? Maybe we could factorize/centralize this for other kinds of desktop layout 🤷

Tip-Sy (approved):
LGTM! Thanks for your explanations/documentations and all the hard work 💪

grosbouddha (approved):

src/specialOffer/SpecialOffer.tsx
guillaumewuip: Yes, as for now we want to be sure we don't mix the two, I will keep the initial typing, thanks for your suggestion!

src/specialOffer/SpecialOffer.tsx
```diff
@@ -0,0 +1,84 @@
...
+        {informationLink && (
+          <Disclaimer useInfoIcon>
+            <Button status={ButtonStatus.UNSTYLED} href={informationLink.url}>
+              {informationLink.label}
+            </Button>
+          </Disclaimer>
+        )}
+      </StyledSpecialOfferText>
+      <StyledSpecialOfferButtons>
+        {(actions as Array<Link | Button>).map(action => (
```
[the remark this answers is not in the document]
guillaumewuip: Doing so I'm forced to introduce a third (weaker) type because of 
image

I would need something like
```ts
type LinkOrButtonOrJustLabel = Readonly<{
  label: string
  url: string | undefined
  onClick: string | undefined
}>
```

Then map on the `actions` props to have an array of props for `<Button />`

```ts
  const buttonPropsList: Array<{
    label: string
    url: string | undefined
    onClick: string | undefined
  }> = (actions as  as Array<Link | Button>).map(action => ({ url: undefined, onClick: undefined, ...action })
```

The central issue here is me not wanting to expose a weak type `LinkOrButtonOrJustLabel` in the API of the component to be sure nobody pass just a label 🙄

I propose to stay like it is. At least we're hiding this design consideration inside the component, which is [good](https://prl.ccs.neu.edu/img/p-tr-1971.pdf)

src/specialOffer/SpecialOffer.story.mdx
```diff
@@ -0,0 +1,208 @@
+import { action } from '@storybook/addon-actions'
+import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs/blocks'
+
+import { SpecialOffer } from './index'
+
+<Meta title="Widgets/SpecialOffer" />
```
gabrielagabriel, line 6:
```suggestion
<Meta title="Pages/SpecialOffer" />
```

src/specialOffer/SpecialOffer.style.tsx
gabrielagabriel: I would rather use mobile online declaration

src/specialOffer/SpecialOffer.style.tsx
gabrielagabriel: SpecialOffer layout is the very same layout that we can find at `SuccessModal`
On SuccessModal the  Media x Body proportion was 1/3 - 2/3. Which means 7-% is indeed a magic number.
Would be important to keep this on track when we are going to realign layouts.

src/specialOffer/SpecialOffer.tsx
```diff
@@ -0,0 +1,78 @@
...
+          {informationLink && (
+            <Disclaimer useInfoIcon>
+              <Button status={ButtonStatus.UNSTYLED} href={informationLink.url}>
+                {informationLink.label}
+              </Button>
+            </Disclaimer>
+          )}
+        </StyledSpecialOfferText>
+        <StyledSpecialOfferButtons>
+          {(actions as Array<Link | Button>).map(action => (
```
gabrielagabriel, line 65: nit: Declare this function outside render

src/specialOffer/SpecialOffer.style.tsx
```diff
@@ -0,0 +1,48 @@
...
+`
+
+export const StyledSpecialOfferFigure = styled.figure`
+  width: 70%;
+  padding-top: ${space.xl};
+  padding-bottom: ${space.m};
+  margin: 0;
+
+  ${normalizeHorizontally}
```
gabrielagabriel, line 29: Why do you need normalization?

gabrielagabriel (approved):

src/specialOffer/SpecialOffer.style.tsx
guillaumewuip: It's for the `figure` to have the same margins as the `SubHeader` or the `Paragraph` below.

With normalization: 

image


Without:

image

src/specialOffer/SpecialOffer.style.tsx
guillaumewuip: It's a magic value but for now I don't think it's used elsewhere. Let's wait for a new usage of this one before trying to factorize it.

src/specialOffer/SpecialOffer.tsx
guillaumewuip: Mmh I don't want to declare a function `renderButton :: action -> JSX.Element` outside of `SpecialOffer` because it would not be a real component (no props signature). Also the performance impact of having an arrow function in `Array.map` should be very small (JS runtime have usually no issue creating lot of functions and the array is always very small).

PR #1063, opened 2021-01-29
Add illustrations

mcellier:
## Description

Add library illustration to kirk (with a great inspiration of what we have for icons)

screencapture-localhost-4000-2021-01-29-15_11_34

## What has been done

- Exported illustrations from figma
- Optimized them though svgomg
- Removed the background
- Handle the sizing rule (100% with max 33% of viewport height)

## Things to consider

- These illustrations are not exhaustive: I imported every one of them from the lib, but we have more on kairos.
- Some of them have small inconsistencies (transparent t-shirt for instance) and should be fixed 
- We'll need to update at least `IllustratedSection` and `SuccessModal` to replace image configuration as string by JSX.Element (to pass an `<Illustration />`, `<Avatar />` or `` element). This will be a breaking change in a future PR. 

## How it was tested

Storybook (and kairos by modifying a bit `IllustratedSection`)


src/_utils/illustration.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+const StyledIllustration = styled.svg`
+  max-height: 33vh;
+  width: 100%;
+  object-fit: contain;
+`
+
+export const Illustration = ({ children, title = '' }: BaseIllustrationProps): JSX.Element => (
+  <StyledIllustration
+    viewBox="0 0 375 204"
```
gabrielagabriel, line 22: would declare as default props

gabrielagabriel (approved):

gabrielagabriel:
Thanks a lot for this PR. Its like a dream coming true.

src/_utils/illustration.tsx
mcellier: the viewBox is here to have the basic aspect ratio. If we have a width under 375 it'll break using min-width. We should not override these imho

Tuxosaurus (approved):
FYI I tested the page on both Chrome and Firefox, saw the same issues as what's on your screenshot.
It's kind of bad to feature them in storybook when they can't be used as is, but at least we have something to show Designers and see how we can improve things so we can simply export from Figma

mcellier:
> FYI I tested the page on both Chrome and Firefox, saw the same issues as what's on your screenshot.
> It's kind of bad to feature them in storybook when they can't be used as is, but at least we have something to show Designers and see how we can improve things so we can simply export from Figma
> 
> EDIT: ok we found what was wrong with the images

Thanks to @Tuxosaurus eagle eyes, we spotted the problem. 
SVG illustrations contains masks using one-letter id. We had several `<mask id="a">` generating collisions when using all these illustrations in a single page. We probably won't have seen the issue on kairos, but i prefixed the mask ids to avoid the issue on storybook. 

screencapture-localhost-4000-2021-01-29-16_14_32

PR #1065, opened 2021-01-29
[PROPIM-678] SEA Page

gabrielagabriel:
## Description

2 Column Grid System to SEA Landing Pages
Pages will be composed on the App. The Grid Layout is only to "hold" the Grid logic of the page

https://www.loom.com/share/baa17ea75ceb43e39560f5f0c315944f

## What has been done

- Use CSS Grid to handle Grid

## Things to consider

- Exporting only public API
- Exported `Elements` are only used inside the `Component`
- Specific layout

## How it was tested

- RTL won't help to test this component since it is a CSS one. Would be better to screenshot on the APP


src/layout/section/landingGridSection/LandingGridSection.style.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+import { color, responsiveBreakpoints } from '../../../_utils/branding'
+
+const Grid = styled.div`
+  display: grid;
+  height: 100vh; //@todo: strech
+  grid-template-areas:
+    'figure'
+    'body';
+  grid-template-rows: minmax(33vh, 33vh) auto;
```
simonrelet, line 11: Isn't `minmax(33vh, 33vh)` the same as `33vh`?

src/layout/section/landingGridSection/LandingGridSection.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import { HeroSection, HeroSectionProps } from '../heroSection'
+import { LandingGridSectionElements } from './LandingGridSection.style'
+
+export type LandingGridSectionProps = HeroSectionProps &
+  Readonly<{
+    children: React.ReactNode
+  }>
+
+export const LandingGridSection = ({
```
grosbouddha, line 11: I would drop the *Grid* from the naming (folder and file name), the fact that it uses a grid internally is an implementation detail.
Other way to phrase it: if we start using grids for other components or section, do we need to add grid in their name ?

src/layout/section/landingGridSection/LandingGridSection.tsx
```diff
@@ -0,0 +1,28 @@
...
+  children,
+  heroText,
+  heroImageUrlLarge,
+  heroImageUrlSmall,
+}: LandingGridSectionProps) => (
+  <LandingGridSectionElements.Grid>
```
grosbouddha, line 17: I'm still not convinced by this API with nested children items.
How does this work if I write that in kairos:
```
<LandingGridSectionElements.Grid>
    <LandingGridSectionElements.Figure>
      fist figure
    </LandingGridSectionElements.Figure>
    <LandingGridSectionElements.Body>{children}</LandingGridSectionElements.Body>
    <LandingGridSectionElements.Body>stuff</LandingGridSectionElements.Body>
    <LandingGridSectionElements.Figure>
      <HeroSection
        heroText={heroText}
        heroImageUrlSmall={heroImageUrlSmall}
        heroImageUrlLarge={heroImageUrlLarge}
        bottomElement={null}
      />
    </LandingGridSectionElements.Figure>
  </LandingGridSectionElements.Grid>
```

src/layout/section/landingGridSection/LandingGridSection.style.tsx
```diff
@@ -0,0 +1,46 @@
...
+  background-color: ${color.white};
+
+  @media (${responsiveBreakpoints.isMediaLarge}) {
+    grid-template-areas: 'body figure';
+    grid-template-columns: repeat(2, 1fr);
+    grid-template-rows: auto;
+  }
+`
+const Body = styled.div`
+  grid-area: body;
```
grosbouddha, line 21: This implementation seems to rely on the fact that the component takes the full page viewport (no header/footer). how does this work with pages with a header or sticky footer ?
If this component is a section, we should be able to stack it with other sections and use it on pages with headers.

if this is not a section but a page layout/template, we have an issue since we want to consolidate them and we should loop in @wakooka.

src/layout/section/landingGridSection/LandingGridSection.tsx
grosbouddha, line 11: I would drop the Landing* from the name too. The component name should be descriptive of what it's doing not where it's used. We could have a quick async brainstorm for a good name in slack and other team members.

grosbouddha (changes requested):

src/layout/section/landingGridSection/LandingGridSection.style.tsx
```diff
@@ -0,0 +1,46 @@
+import styled from 'styled-components'
+
+import { color, responsiveBreakpoints } from '../../../_utils/branding'
+
+const Grid = styled.div`
+  display: grid;
+  height: 100vh; //@todo: strech
```
wakooka, line 7: We have many issue with 100vh on mobile, we might want to avoid this value.

src/layout/section/landingGridSection/LandingGridSection.style.tsx
wakooka: Thanks for the ping, I think we need thorough testing on this especially (Safari Mobile, IE11) once the Layout will be implemented we might have conflicts with it:
https://codesandbox.io/s/adoring-fermi-7xkuo

src/layout/section/landingGridSection/LandingGridSection.tsx
```diff
@@ -0,0 +1,28 @@
...
+    <LandingGridSectionElements.Body>{children}</LandingGridSectionElements.Body>
+    <LandingGridSectionElements.Figure>
+      <HeroSection
```
grosbouddha, line 20: We should not have a section in a section.

grosbouddha (changes requested):

wakooka (changes requested):
IMHO the PR is trying to be too generic and it might break some of the Layout work that has been done, not sure but this needs to be tested thoroughly. For now I'd rather have something more specific if you want to move fast in your project. And then after that, we can think about a good grid system with a design docs maybe? 

During the work on Layout, we identified 3 variants on large media which could be included in the design docs:
- `small`: max-width: 662px (1 column)
- `large`: max-width: 1016px; (2 columns 33% / 67%)
- `full`: full width (2 columns 50%)

src/layout/section/landingGridSection/LandingGridSection.tsx
gabrielagabriel: You should not use `LandingGridSectionElements`in Kairos

src/layout/section/landingGridSection/LandingGridSection.style.tsx
gabrielagabriel: It supposed to have no header and no footer indeed. No need to incapsulate on a "Layout" component.
But I'll check it out with your implementation @wakooka . Would be better if the POC would be on this library, we could test the integration with Storybook

src/layout/section/landingGridSection/LandingGridSection.tsx
gabrielagabriel: This is really a grid only for these pages. I've even thought about naming it Page instead. But sure.. open for naming conventions

src/layout/section/landingGridSection/LandingGridSection.tsx
gabrielagabriel: Why does?
https://html.spec.whatwg.org/multipage/sections.html#headings-and-sections

Anyway, I'm considering dropping `HeroSection` since it doesn't cover 100 of my case

gabrielagabriel:
> IMHO the PR is trying to be too generic and it might break some of the Layout work that has been done, not sure but this 
This PR is nothing but generic. It really treats SEA Landing Page grid system. and we have had no intention to treat any cases of grid outside of what has been designed for these pages. We have no design specs for that.

src/layout/section/landingGridSection/LandingGridSection.tsx
gabrielagabriel: The main thing that this "component does is to set a "Grid Layout" for this page. Should I call it **Layout** then?

src/layout/section/landingGridSection/LandingGridSection.style.tsx
```diff
@@ -0,0 +1,46 @@
+import styled from 'styled-components'
+
+import { color, responsiveBreakpoints } from '../../../_utils/branding'
+
+const Grid = styled.div`
+  display: grid;
```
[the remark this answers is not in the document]
gabrielagabriel: Fun Fact: 11 was the first one to implement grids
https://caniuse.com/css-grid

The only issue is with "grid-area". I'll check our users browser to make sure we wont stop the progress of the internet /our applications for real old browsers. We can even think about using some kind of modernize for identifying that or even feature-queries. Thanks for the tips, I'll make sure the layout is rendered correctly

src/_utils/themeProvider/index.tsx
```diff
@@ -107,6 +107,10 @@ const CssResetGlobalStyles = createGlobalStyle`
   body {
     line-height: 1;
     color: ${DEFAULT_TEXT_COLOR};
+    height: 100%;
```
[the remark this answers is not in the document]
gabrielagabriel: ok.. I'll remove that

src/layout/section/landingGridSection/LandingGridSection.style.tsx
wakooka: The point of having the "Layout" component is to be used everywhere, with or without header/footer this is the view on kairos to decide.

> > IMHO the PR is trying to be too generic and it might break some of the Layout work that has been done, not sure but this ...
> 
> @wakooka
> This PR is nothing but generic. It really treats SEA Landing Page grid system. and we have had no intention to treat any cases of grid outside of what has been designed for these pages. We have no design specs for that.

OK I see, thanks for clarifying and I was off on Friday so maybe that's why I don't have the full picture. When I said generic it's mainly because the naming and because it's relying on some css value on body, #root, etc...

wakooka:
> > IMHO the PR is trying to be too generic and it might break some of the Layout work that has been done, not sure but this ...
> 
> @wakooka
> This PR is nothing but generic. It really treats SEA Landing Page grid system. and we have had no intention to treat any cases of grid outside of what has been designed for these pages. We have no design specs for that.

OK I see, thanks for clarifying and I was off on Friday so maybe that's why I don't have the full picture. When I said generic it's mainly because the naming and because it's relying on some css value on body, #root, etc...

src/layout/section/landingGridSection/LandingGridSection.tsx
guillaumewuip: `SEALandingLayout` maybe? The fact that this use a Grid Layout could change in the future. It's like a design implementation detail.

src/layout/section/landingGridSection/LandingGridSection.tsx
guillaumewuip: > You should not use LandingGridSectionElementsin Kairos

Are they exported?

gabrielagabriel: @guillaumewuip  Unfortunatelly. Is there a way to use it inside on component.tsx file without exporting it from style.tsx?

gabrielagabriel: @grosbouddha  @guillaumewuip  

Resolved here: [index.tsx](https://github.com/blablacar/ui-library/blob/b7558ae466b9c9e7cd6ff378aa0022bde23129ea/src/layout/section/SEALandingLayout/index.tsx#L1)

src/layout/section/landingGridSection/LandingGridSection.tsx
gabrielagabriel: Resolved here: [Renaming](https://github.com/blablacar/ui-library/pull/1065/commits/69e4198ccf19034c72563e2762b99240d8fa1c93)

gabrielagabriel: I've followed @guillaumewuip  [naming suggestion ](https://github.com/blablacar/ui-library/pull/1065/commits/69e4198ccf19034c72563e2762b99240d8fa1c93)

Let me know if it that ok for you @grosbouddha

src/index.tsx
```diff
@@ -31,6 +32,8 @@ export * from './layout/section/illustratedSection'
 export * from './layout/section/mediaSection'
 export * from './layout/section/slideSection'
 export * from './layout/section/textFieldsSection'
+export * from './layout/section/SEALandingLayout'
```
Tuxosaurus, line 35: The build is failing on this test, I suspect a case issue
Screen Shot 2021-02-03 at 14 20 17

gabrielagabriel: Thanks for looking into it! I really had no idea why is it falling :S

src/layout/section/SEALandingLayout/SEALandingLayout.style.tsx
```diff
@@ -0,0 +1,46 @@
...
+    grid-template-rows: auto;
+  }
+`
+const Body = styled.div`
+  grid-area: body;
+
+  @media (${responsiveBreakpoints.isMediaLarge}) {
+    align-self: center;
+    justify-self: center;
+    max-width: 23.4375rem; //magical number: iPhone6 width
```
wakooka, line 26: I don't understand the comment about iPhone 6 width. We are in a media large media query, but the comment is about iPhone 6

gabrielagabriel: Yes. The design has used the max-size of the content as a mobile one. So no changes. This is the same size as 
`componentSizes.searchOverlayWidth`. It keeps happing over and over again in the app.

src/layout/section/landingGridSection/LandingGridSection.style.tsx
gabrielagabriel: I 've moved 100vh to a media query big screens I hope would help and be able to use it

src/index.tsx
gabrielagabriel: Still failing 🦺

gabrielagabriel:
@wakooka @grosbouddha  CODE REVIEW status:

-  Renamed the component to SPA "section" convention
- Used a business term SEA to indicate that is not a generic component
- Can't use `HeroSection` since I can't redefine large screens image ratio
- Do not export any Elements to encapsulate on the component.
- Still using css-grid since we don't have our Layout Component ready yet. We can update any thing that this component require after

src/layout/section/landingGridSection/LandingGridSection.tsx
gabrielagabriel: HeroSection was dropped

src/layout/section/seaSection/SeaSection.style.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { TextSubHeader } from '../../../typography/subHeader'
+
+const CONTENT_WIDTH = '375px' // iPhone6 width same as componentSizes.searchOverlayWidth
+
+const Wrapper = styled.div`
+  display: grid;
+  grid-template-areas:
+    'media'
+    'body';
+  grid-template-rows: minmax(33vh, 33vh) auto;
```
Tuxosaurus, line 13: wasn't `minmax(33vh, 33vh)` changed with a previous feedback?

src/layout/section/seaSection/SeaSection.story.mdx
```diff
@@ -0,0 +1,97 @@
...
+  </Story>
+</Canvas>
+
+## Specifications
+
+SEA Landing Pages responsive 2 Column Vertical Grid
+
+**Large**
+
+- `Media` tracker has 100%;
```
Tuxosaurus, line 78: `100%` of what?

Nit: I would also replace `tracker` with `element`. Not sure what you mean with "tracker" here.

src/layout/section/seaSection/SeaSection.style.tsx
```diff
@@ -0,0 +1,67 @@
+import styled from 'styled-components'
+
+import { color, font, responsiveBreakpoints, space } from '../../../_utils/branding'
+import { TextSubHeader } from '../../../typography/subHeader'
+
+const CONTENT_WIDTH = '375px' // iPhone6 width same as componentSizes.searchOverlayWidth
```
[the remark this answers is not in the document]
wakooka: I'm not a fan of sharing a variable just because they have the same value. Are we sure it was intended on the design side? If we update the `searchOverlayWidth` for a specific reason de we want it to update this screen as well? I would keep it local `CONTENT_WIDTH` variable here.

src/layout/section/seaSection/SeaSection.style.tsx
```diff
@@ -0,0 +1,67 @@
...
+const Wrapper = styled.div`
+  display: grid;
+  grid-template-areas:
+    'media'
+    'body';
+  grid-template-rows: minmax(33vh, 33vh) auto;
+  background-color: ${color.white};
+
+  @media (${responsiveBreakpoints.isMediaLarge}) {
+    height: 100vh; //@todo: strech to parent
```
wakooka, line 17: This is a new file but it has a `@todo`, could you do it in this PR?

src/layout/section/seaSection/SeaSection.story.mdx
```diff
@@ -0,0 +1,97 @@
...
+- `Body` tracker has max-with `375px` and is vertical/horizontally centered
```
wakooka, line 79: Does that mean that if my screen is 500px for instance, I will only get content on 375px wide?

src/layout/section/seaSection/SeaSection.style.tsx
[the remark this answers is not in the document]
Tuxosaurus: Alright, I've checked where `searchOverlayWidth` is used and yes it is definitely not link to a device size, so I agree, let's keep `CONTENT_WIDTH`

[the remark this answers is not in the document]
wakooka: I think many of the entries in `componentSizes` object shouldn't be there as they are not global

src/layout/section/seaSection/SeaSection.style.tsx
gabrielagabriel: Infortunally I can't because **height** in CSS depends on the parent. I can't define the height for `body` or `root` since its part of `Layout` 🤷‍♀️

src/layout/section/seaSection/SeaSection.story.mdx
gabrielagabriel: `Tracker` is the term used to place any element inside a grid area. 

Is that too generic for you?

Tuxosaurus: Oh you mean `Track` then, that's ok

src/index.tsx
Tuxosaurus: 🟢

src/layout/section/seaSection/SeaSection.style.tsx
wakooka: ok then 
- [ ] remove comment

src/layout/section/seaSection/SeaSection.style.tsx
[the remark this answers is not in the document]
wakooka: Can you revert based on what we said?

src/layout/section/seaSection/SeaSection.story.mdx
wakooka: It's probably a mistake in the design?  I don't think we had this behaviour on any page.
The width should grow with the screen when you are under 800px in small media mode.
Wdyt?

src/layout/section/seaSection/SeaSection.style.tsx
```diff
@@ -0,0 +1,68 @@
+import styled from 'styled-components'
+
+import { color, componentSizes, font, responsiveBreakpoints, space } from '../../../_utils/branding'
+import { TextSubHeader } from '../../../typography/subHeader'
+
+// Common content width with used on the app
+const CONTENT_WIDTH = componentSizes.searchOverlayWidth
```
wakooka, line 7: See https://github.com/blablacar/ui-library/pull/1065#discussion_r570897730

src/layout/section/seaSection/SeaSection.style.tsx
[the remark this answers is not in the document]
gabrielagabriel: Ok! I'll revert and let the comment to keep a track on where we are using same values (looking for patterns)

src/layout/section/seaSection/SeaSection.style.tsx
```diff
@@ -0,0 +1,67 @@
...
+  background-image: ${props => `url(${props.small})`};
+
+  @media (${responsiveBreakpoints.isMediaLarge}) {
+    grid-row-start: 1;
+    grid-column-start: 2;
+    background-image: ${props => `url(${props.large})`};
+  }
+`
+
+const Leading = styled(TextSubHeader).attrs({ isInverted: true })`
```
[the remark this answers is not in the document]
gabrielagabriel: Designers usually use `Leading` term to represent the biggest item in the page, although the right term in typography is `Display` 

I've update the API to represent HTML => Going for `Heading` thanks for the tips

Tuxosaurus (approved):

src/layout/section/seaSection/SeaSection.story.mdx
gabrielagabriel: @wakooka  Check it out [the layout](https://www.figma.com/file/zMeyZPxqu1SaosHAVuz6qs/Landing-Pages?node-id=150%3A648) it can better understand why the content has this `max-width`

guillaumewuip (approved):

wakooka (approved):

grosbouddha:
As discussed, let's rename from layout/section/seaSection/ to layout/page/seaPage since the sea component is a full page, non-stackable component (sections are stackable).
As well, let's make sure this new page layout is added to the in-progress layout taxonomy (cc @wakooka)

grosbouddha:
We should move src/layout/section/seaSection/carpool-coral-bus-large.svg outside of the seaPage folder, probably in a global illustration folder.
It can be done in a follow up PR.

grosbouddha (approved):

Tuxosaurus (approved):

PR #1069, opened 2021-02-01
[PROPIM-690] New Showcase/Marquee component

Tuxosaurus:
https://user-images.githubusercontent.com/373381/106726233-7004c900-660a-11eb-95ba-52c3efe645d2.mp4

## How it was tested

Tested locally on Firefox MacOS


src/showcase/Showcase.story.mdx
```diff
@@ -0,0 +1,34 @@
...
+      <div>through all children</div>
+      <div>regardless of their number</div>
+      <div>or their type</div>
+      <Paragraph>(here's a Paragraph)</Paragraph>
+    </Showcase>
+  </Story>
+</Canvas>
+
+## Specifications
+
```
gabrielagabriel, line 23: Missing specs.. 
```suggestion
Showcase is **marquee** component
```

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 - [...]
 - **[UPDATE]** `Storybook` version `6.1.15`
+- **[NEW]** Add `Showcase`
```
gabrielagabriel, line 5: I would call it `Marquee`

gabrielagabriel (approved):
🌈

src/showcase/Showcase.tsx
```diff
@@ -0,0 +1,25 @@
...
+export type ShowcaseProps = Readonly<{
+  children: Array<React.ReactNode>
+}>
+
+export const Showcase = (props: ShowcaseProps): JSX.Element => {
+  const { children } = props
+
+  return (
+    <StyledShowcase>
+      {children.map((child, index) => {
```
mcellier, line 14: nit: You might want to filter falsy elements, as we often have the following pattern:

```
{condition && ( <Element/> )}
```

src/showcase/Showcase.unit.tsx
```diff
@@ -0,0 +1,20 @@
...
+  it('Should render', () => {
+    render(
+      <Showcase>
+        <div>fhtagn</div>
+        <div>cthulhu</div>
+        <Paragraph>Paragraph</Paragraph>
+      </Showcase>,
+    )
+    expect(screen.getByRole('list')).toBeInTheDocument()
+    expect(screen.getAllByRole('listitem')).toHaveLength(3)
```
mcellier, line 18: Did you test to improve the test by checking the visibility of elements over time?

You can try a combination of `waitFor()` and `toBeVisible()`

CHANGELOG.md
```diff
@@ -12,10 +13,6 @@
 # v46.1.0 (27/01/2021)
 
 - **[NEW]** Add `PaypalIcon`
-  <<<<<<< HEAD
```
pith, line 15: nit: rebase issue

Tuxosaurus: yep, rebased already

gabrielagabriel: It's my fault, It will be fix on this https://github.com/blablacar/ui-library/pull/1070

src/showcase/Showcase.tsx
grosbouddha: ```{children.filter(Boolean).map((child, index) => {```

src/showcase/Showcase.story.mdx
```diff
@@ -0,0 +1,34 @@
...
+import { Showcase } from './index'
+
+<Meta title="Widgets/Showcase" />
+
+# Showcase
+
+<Canvas>
+  <Story name="Default">
+    <Showcase>
+      <div>Lists and loops</div>
```
grosbouddha, line 13: Do we need this very wide HTML-based API? Can we just have a list of stings ?

src/showcase/Showcase.story.mdx
```diff
@@ -0,0 +1,34 @@
...
+import { Paragraph } from '../paragraph'
+import { Showcase } from './index'
+
+<Meta title="Widgets/Showcase" />
+
+# Showcase
+
+<Canvas>
+  <Story name="Default">
+    <Showcase>
```
grosbouddha, line 12: Not sure about the a11y of this component (right now, it's not too good). Having moving UI pieces is usually annoying for impaired users (since it's hard to make it accessible) but I guess I can look somewhere else.

This remarks apply for SEO crawlers a11y too: ~~They will only have access to the first item in the showcase when crawling this page.~~ It's all opacity based so it should be okish for SEO. (https://stackoverflow.com/questions/3593158/does-google-index-pages-with-opacity0-or-hidden-or-displaynone)

Tuxosaurus:
Sorry, I will take the comments into account but I will need to rework the animation after we checked with the UX: it's not just fading, it's also sliding

CHANGELOG.md
Tuxosaurus: Now that we know it actually slides, I'll go for Marquee (will change when it officially makes it into the Design System)

src/showcase/Showcase.story.mdx
Tuxosaurus: We are presented with two use cases, one that slides between a new custom component that will include an itinerary, and another that slides between subheader components. So I figured I would not restrict what you can pass to the Showcase/Marquee component.

Great thinking on pausing the animation on hover to read it. It's not in the specs (well there's not much specs anyway) but I'll add it.

wakooka (approved):

src/showcase/Showcase.story.mdx
Tuxosaurus: Added a line, but we don't really have specs for this.

src/showcase/Showcase.story.mdx
Tuxosaurus: This is not purely for SEO, rather for SEA. Nevertheless, for another component we checked with our expert to see if "display: none" was to be avoided and they said it was OK, as far as all the content is there on the first render. Further improvements might come later.

As for A11Y, I would agree but right now we're only replacing a legacy page "as is". I do hope that more time is spent on designing a whole new project on the matter of SEA pages but it's not going to happen soon IMO.

src/showcase/Showcase.unit.tsx
Tuxosaurus: I tried your suggestion but couldn't make it more so far...

Tuxosaurus: As seen together, after multiple attempts, I added a TODO

mcellier (approved):

PR #1074, opened 2021-02-03
add new TripCardSample component

Tuxosaurus:
Screen Shot 2021-02-03 at 16 43 39

Context: SEA landing page (slideshow between 3 "tripCardSample")
Screen Shot 2021-02-03 at 16 42 10

## Things to consider

The name, it's not that good, so I welcome suggestions.
Also it may be moved in a "SEA" subfolder if it's what we want, since it is not meant to be used widely outside of these Landing pages.

Tested on Firefox MacOS

PR #1076, opened 2021-02-04
update TripCardSample with better style

Tuxosaurus:
Missing update from the feedback I got on https://github.com/blablacar/ui-library/pull/1074


gabrielagabriel (approved):
🥇

PR #1079, opened 2021-02-04
update style and story to match use case

Tuxosaurus:
Screen Shot 2021-02-04 at 17 39 06

Fixed horizontal spacing to match Paragraph, SubHeader, etc.

## Things to consider
I had to paste the style for "normalizeHorizontally" cause it conflicted with the styled component's props

## How it was tested

Tested on Firefox MacOS, storybook and canary

### DIFF
--- a/src/marquee/Marquee.style.tsx
+++ b/src/marquee/Marquee.style.tsx
@@ -1,5 +1,7 @@
 import styled from 'styled-components'
 
+import { normalizeHorizontally, NormalizeProps } from '../layout/layoutNormalizer'
+
 // Total animation duration:
 // 3400ms fully visible + 320ms slideInOut
 const animationDuration = 3720
@@ -12,10 +14,13 @@ export const StyledMarquee = styled.ul`
   }
 `
 
-export const StyledMarqueeItem = styled.li<{
-  position: number
-  totalItems: number
-}>`
+type MarqueeItemProps = NormalizeProps &
+  Readonly<{
+    position: number
+    totalItems: number
+  }>
+
+export const StyledMarqueeItem = styled.li<MarqueeItemProps>`
   /* animation steps defined by how many items are passed */
   @keyframes slideInOutWait {
     0% {
@@ -54,4 +59,6 @@ export const StyledMarqueeItem = styled.li<{
   animation-duration: ${props => `${animationDuration * props.totalItems}ms`};
   animation-iteration-count: infinite;
   animation-delay: ${props => `${animationDuration * props.position}ms`};
+  width: 100%;
+  ${normalizeHorizontally};
 `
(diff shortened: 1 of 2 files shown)


mcellier (approved):

src/marquee/Marquee.style.tsx
```diff
@@ -54,4 +56,10 @@ export const StyledMarqueeItem = styled.li<{
   animation-duration: ${props => `${animationDuration * props.totalItems}ms`};
   animation-iteration-count: infinite;
   animation-delay: ${props => `${animationDuration * props.position}ms`};
+  width: calc(100% - (${horizontalSpace.outer}*2));
+  /* cannot use normalizeHorizontally directly here because it conflicts with props */
```
gabrielagabriel, line 60: nit: use `&&` to increase specificity and thus remove !**important**

```suggestion
&& {
 padding-left: ${horizontalSpace.inner};
  padding-right: ${horizontalSpace.inner};
  margin-left: ${horizontalSpace.outer};
  margin-right: ${horizontalSpace.outer};
}
```

gabrielagabriel (changes requested):
Remove `!important`

src/marquee/Marquee.style.tsx
Tuxosaurus: `!important` here is copy/pasted from https://github.com/blablacar/ui-library/blob/master/src/layout/layoutNormalizer.tsx#L18

So I'm not sure duplicating the class is really better than "staying coherent"

I don't have a strong opinion so I'll wait for @wakooka's review if possible

wakooka: I don't understand the comment "it conflict with props", what was the issue?

Tuxosaurus: It wouldn't pass the linter since it's using it's own props that is not in the declared props for this component.
Screen Shot 2021-02-05 at 10 21 02


But for some reason I stopped there while I could simply add the missing prop. I've just updated the PR, it's now working fine with `normalizeHorizontally`

src/marquee/Marquee.style.tsx
```diff
@@ -15,6 +17,7 @@ export const StyledMarquee = styled.ul`
 export const StyledMarqueeItem = styled.li<{
   position: number
   totalItems: number
+  hasHorizontalSpacing?: boolean
 }>`
```
wakooka, line 21: `& NormalizeProps` instead of repeating it

wakooka (approved):

src/marquee/Marquee.style.tsx
Tuxosaurus: Thanks, still some getting used to manipulating types on my side.
I changed it, looks cleaner now.

Tuxosaurus:
> Remove `!important`

It seems I can't mark the requested change as "Done", but it is. Can you check @gabrielagabriel ?

gabrielagabriel (approved):
💯

PR #1084, opened 2021-02-08
add a11y to DropdownButton and HamburgerButton

Tuxosaurus:
Added A11Y props to DropdownButton. I will need to add a dropdown that toggle a menu with 2 items, Login & Signup.
I want this dropdown to read something like "Connect" with screen readers but right now the API doesn't allow for this.

I also moved the DropdownButton out of the "Forms" category back to "Widgets" as it feels more appropriate regarding the way we use it.

Tested locally on Firefox MacOS with VoiceOver and a canary release.

### DIFF
--- a/src/dropdownButton/DropdownButton.story.mdx
+++ b/src/dropdownButton/DropdownButton.story.mdx
@@ -4,7 +4,7 @@ import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs/blocks'
 import { Avatar } from '../avatar'
 import { DropdownButton } from './index'
 
-<Meta title="Forms/DropdownButton" />
+<Meta title="Widgets/DropdownButton" />
 
 # **DropdownButton**
 
--- a/src/dropdownButton/DropdownButton.story.mdx
+++ b/src/dropdownButton/DropdownButton.story.mdx
@@ -4,7 +4,7 @@ import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs/blocks'
 import { Avatar } from '../avatar'
 import { DropdownButton } from './index'
 
-<Meta title="Widgets/DropdownButton" />
+<Meta title="Menu/DropdownButton" />
 
 # **DropdownButton**
 
@@ -23,9 +23,9 @@ import { DropdownButton } from './index'
 
 ```jsx
 import { DropdownButton } from '@blablacar/ui-library/build/DropdownButton'
-  <DropdownButton onClick={action('onClick')} open={false}>
-    <Avatar />
-  </DropdownButton>
+<DropdownButton onClick={action('onClick')} open={false}>
+  <Avatar />
+</DropdownButton>
 ```
 
 <ArgsTable of={DropdownButton} />
--- a/src/dropdownButton/DropdownButton.tsx
+++ b/src/dropdownButton/DropdownButton.tsx
@@ -9,7 +9,7 @@ import { StyledDropdownButton } from './DropdownButton.style'
 export type DropdownButtonProps = A11yProps &
   Readonly<{
     onClick: (event: React.MouseEvent<HTMLElement>) => void
-    children: JSX.Element | string
+    children: React.ReactNode
     open?: boolean
     className?: string
     iconPosition?: 'left' | 'right'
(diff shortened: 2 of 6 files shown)


nickoola:
The A11y props addition seems ok for me.

nickoola (approved):

Tuxosaurus:
> The A11y props addition seems ok for me.
> Could you add a demo (Loom) to show how it works with VoiceOver?
> 
> Other remark not related to your change, on prod it seems to be outputted as a `<a href="#" role="button">` strange as it seems not possible in the component code.
> Screenshot 2021-02-09 at 09 37 20

Thanks for your review.
Your screenshot is from the V3, on SPA we do have a button.
I'll add a demo but also I will need to update the "HamburgerButton" too as it is used in the same way on small devices.

src/dropdownButton/DropdownButton.story.mdx
```diff
@@ -4,7 +4,7 @@ import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs/blocks'
 import { Avatar } from '../avatar'
 import { DropdownButton } from './index'
 
-<Meta title="Forms/DropdownButton" />
+<Meta title="Widgets/DropdownButton" />
```
gabrielagabriel, line 7: DropdownButton is only used for the menu. I do agree that it's not a form then, but it's not a generic component either. Could you move closer to where is applied?

```suggestion
<Meta title="Menu/DropdownButton" />
```

src/dropdownButton/DropdownButton.tsx
```diff
@@ -2,37 +2,38 @@ import React from 'react'
...
-  onClick: (event: React.MouseEvent<HTMLElement>) => void
-  children: JSX.Element | string
-  open?: boolean
-  className?: string
-  iconPosition?: 'left' | 'right'
-}>
+export type DropdownButtonProps = A11yProps &
+  Readonly<{
+    onClick: (event: React.MouseEvent<HTMLElement>) => void
+    children: JSX.Element | string
```
gabrielagabriel, line 12:
```suggestion
    children: React.node
```

src/dropdownButton/DropdownButton.tsx
```diff
@@ -2,37 +2,38 @@ import React from 'react'
...
-  >
-    <button aria-expanded={open} type="button" onClick={onClick}>
-      {iconPosition === 'left' && <ChevronIcon iconColor={color.lightMidnightGreen} down />}
-      {children}
-      {iconPosition === 'right' && <ChevronIcon iconColor={color.lightMidnightGreen} down />}
-    </button>
-  </StyledDropdownButton>
-)
+export const DropdownButton = (props: DropdownButtonProps) => {
+  const { open = false, children, onClick, className = '', iconPosition = 'right' } = props
```
gabrielagabriel, line 19: IMO destructuring props on function declarations is easier to read the attributes of the function itself.
I would suggest to keep as before

Tuxosaurus: Not in this case. In fact most of the components using the A11Yprops use what I wrote because it's more readable.

By destructuring in the signature you have to repeat the list of props:
`
export const DropdownButton = ({
  open = false,
  children,
  onClick,
  className = '',
  iconPosition = 'right'
}: DropdownButtonProps) => {
  const a11yAttrs = pickA11yProps<DropdownButtonProps>({ open, children, onClick, className, iconPosition, ...props})
`

src/dropdownButton/DropdownButton.story.mdx
Tuxosaurus: Done

src/dropdownButton/DropdownButton.tsx
Tuxosaurus: Done

src/hamburgerButton/HamburgerButton.story.mdx
```diff
@@ -2,7 +2,7 @@ import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs/blocks'
 
 import { HamburgerButton } from './index'
 
-<Meta title="Widgets/HamburgerButton" />
+<Meta title="Menu/HamburgerButton" />
```
gabrielagabriel, line 5: 👍

gabrielagabriel (approved):

Tuxosaurus:
@nickoola @gabrielagabriel here's the loom: https://www.loom.com/share/7b2528abfcae4fe48f696f5a315a461e
Notice that at the end, you see the tab order is not good and it's actually the case with the dropdown menu on small device (not related to this PR)

PR #1088, opened 2021-02-10
make Avatar isSmall smaller

Tuxosaurus:
<Avatar isSmall /> made it 40x40px, slightly smaller than the normal size of 48x48px.
The "isSmall" variant is not used in the application so now I'm presented with an Avatar the size of an icon (24x24px) I simply changed the values.

Use case:
Screen Shot 2021-02-10 at 15 23 48

Note that it still works with the Badge:
Screen Shot 2021-02-10 at 15 27 38

Tested locally on Firefox MacOS and storybook.

### DIFF
--- a/src/avatar/Avatar.style.tsx
+++ b/src/avatar/Avatar.style.tsx
@@ -11,8 +11,8 @@ export const StyledAvatar = styled.div`
   height: 48px;
 
   &.kirk-avatar--small {
-    height: 40px;
-    width: 40px;
+    height: 24px;
+    width: 24px;
   }
 
   &.kirk-avatar--medium {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 [...]
+- **[UPDATE]** Change size of `<Avatar isSmall />`
 
 # v47.4.0 (11/02/2021)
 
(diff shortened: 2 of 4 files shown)


wakooka (approved):
What's the use case in the product?

Tuxosaurus:
> What's the use case in the product?

To replace the hamburger menu in the top bar when using a small device (it would be used both logged in and logged out)

wakooka:
Then this should probably be updated https://www.figma.com/file/JjaALcWYWcqGtanufVLKYj/UI-%C2%B7-Pixar?node-id=1347%3A477

Tuxosaurus:
> Then this should probably be updated https://www.figma.com/file/JjaALcWYWcqGtanufVLKYj/UI-%C2%B7-Pixar?node-id=1347%3A477

Good point, I pinged Louis about it in the project's Figma

src/avatar/Avatar.style.tsx
```diff
@@ -11,8 +11,8 @@ export const StyledAvatar = styled.div`
   height: 48px;
 
   &.kirk-avatar--small {
-    height: 40px;
-    width: 40px;
+    height: 24px;
```
gabrielagabriel, line 14: Would be nice to create variables for that

PR #1089, opened 2021-02-10
Bbc 12388/fix phonefield error state

nickoola:
## Description

Fix phonefield error state display

**Before the fix:**
Screenshot 2021-02-10 at 17 55 26
Screenshot 2021-02-10 at 17 55 42

--------------------------------------------------
**After:**
Screenshot 2021-02-10 at 17 51 42
Screenshot 2021-02-10 at 17 51 56

## How it was tested

Local environment with ui-library only (Safari/Chrome/FF)

### DIFF
--- a/src/phoneField/PhoneField.style.tsx
+++ b/src/phoneField/PhoneField.style.tsx
@@ -56,15 +56,16 @@ export const StyledPhoneField = styled.div`
     padding-left: 0 !important;
   }
 
-  &.kirk-error .kirk-phoneField-wrapper--inline {
+  &.kirk-error .kirk-phoneField-wrapper--inline .kirk-phoneField-wrapper--background {
     background: ${color.lightRed};
     animation: phoneFieldError ${transition.duration.fast} ease-in-out;
   }
 
-  &.kirk-error .kirk-selectField,
-  &.kirk-error .kirk-selectField .kirk-icon,
+  &.kirk-error .kirk-selectField-background,
+  &.kirk-error select,
+  &.kirk-error .kirk-selectField-background .kirk-icon,
   &.kirk-error .kirk-textField-wrapper,
-  &.kirk-error .kirk-textField .kirk-textField-wrapper input {
+  &.kirk-error input {
     background: ${color.lightRed};
     border-color: ${color.lightRed};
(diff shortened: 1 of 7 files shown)


src/phoneField/PhoneField.style.tsx
```diff
@@ -56,23 +56,26 @@ export const StyledPhoneField = styled.div`
...
     background: ${color.lightRed};
     border-color: ${color.lightRed};
   }
 
   &.kirk-error .kirk-error-message {
     color: ${color.red};
     display: block;
-    padding: ${space.m};
+    padding-right: calc(${space.m} + ${space.xl});
+    padding-left: calc(${space.m} + ${space.xl});
```
nickoola, line 77: In this component `TextField` & `SelectField` already have `space.xl` horizontally, this padding is not provided by the upper wrapper.
So for the `error message` we need to provide this padding + his own padding.

src/phoneField/__snapshots__/PhoneField.unit.tsx.snap
```diff
@@ -60,8 +60,9 @@ exports[`PhoneField should have the expected markup in the DOM with custom setti
   outline: 0;
   box-shadow: none;
   border: none;
+  border-radius: 16px;
   cursor: pointer;
-  z-index: 1;
+  z-index: 0;
```
nickoola, line 65: Need the same deep position on the `select` as the arrow icon. Otherwise, in case of error display the `select` red background hide the icon.

src/selectField/SelectField.style.tsx
```diff
@@ -36,8 +36,9 @@ export const StyledSelectField = styled.div`
     outline: 0;
     box-shadow: none;
     border: none;
+    border-radius: ${radius.l};
     cursor: pointer;
-    z-index: 1;
+    z-index: 0;
```
nickoola, line 41: Need the same deep position on the select as the arrow icon. Otherwise, in case of error display the select red background hide the icon.

Tip-Sy (approved):
LGTM, thanks!!

mcellier (approved):
Thanks !

PR #1092, opened 2021-02-11
[BBC-12223] Brand new itinerary

mcellier:
## Description

**Warning: this PR is a big breaking change on the `Itinerary` component. We'll have to migrate on kairos, and also some ui-library components (`TripCard` for instance).** 

This PR is still a draft to start gather feedback. Some things are missing:
- [x] Add unit tests
- [x] Implement & tests a11y (with @nickoola)
- [x] Browser tests with SauceLabs
- [x] Pixar validation (with @wakooka)

Figma-style story:
screencapture-192-168-0-101-4000-iframe-html-2021-02-15-18_24_20

New component:
https://user-images.githubusercontent.com/729186/107666197-8bb54280-6c8e-11eb-9c15-d39109e3229d.mp4

Old component:
https://user-images.githubusercontent.com/729186/107666211-8f48c980-6c8e-11eb-997b-f666ef2e3ebc.mp4

## What has been done

- Created brand new components with a modular approach. See storybook for more example.
- Updated `Proximity` component to handle the "icon + text" display (was done on kairos before)

### Changes with previous itinerary
- `headline` has been removed. 
- microdata have been removed 
- Hover are now fixed! (padding + connection)
- We used small bullet before for hidden stops. It's not in the spec, we should use the standard ones. 
- Added highlighted & custom icon option for boost project

### Room for improvement
- The dotted line does not render that well (we can't control its interruption / height)
- Design lost the youtube video presenting the animation. I retrieved (more or less) the old one, but a css wizard may improve it easilly.

### Next steps
- Release and migrate all kairos itineraries 
- Remove the old itinerary and rename newItinerary 
- Update `pages` (examples) , `Stepper`, `TripCard`, `TripCardSample` components. The last 3 should take `<Itinerary>` as an `itinerary` prop, not places & stuff.
- Move `proximity` under the `newItinerary` folder once the migration is complete. This component isn't used outside the itinerary.

## How it was tested

Locally on Storybook + Sauce Labs (see comment) + Kairos


src/newItinerary/Itinerary.style.tsx
```diff
@@ -0,0 +1,21 @@
...
+  // On li sub-element to have a proper hover on clickable elements
+  > ul > li > * {
+    ${normalizeHorizontally};
+  }
+
+  // Used to display the left column. Elements without time also have this empty tag to help with the layout
+  time {
+    width: ${TIME_WIDTH};
+    padding-top: ${space.m};
+    display: ${(props: ItineraryProps) => (props.small ? 'none' : 'initial')};
```
gabrielagabriel, line 19: Do you need to redeclare type here since **StyledItinerary** is already typed?

src/newItinerary/Itinerary.tsx
```diff
@@ -0,0 +1,91 @@
...
+        prevLine = Lines.ACTIVE
+        break
+      case Connection:
+        prevLine = Lines.CONNECTION
+        break
+    }
+  }
+
+  // Next line for Place
+  if (child.type === Place && index + 1 < children.length) {
```
gabrielagabriel, line 64: Why do you prefere use a switch/case instead of a object dictionary?

src/newItinerary/Place.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { Bullet, BulletTypes } from '../bullet'
+import { Proximity } from '../proximity'
+import { ItineraryItem, ItineraryItemProps } from './internals/ItineraryItem'
+import {
+  StyledChevronIcon,
+  StyledContent,
+  StyledLabel,
+  StyledProximity,
+  StyledSubLabel,
+} from './Place.style'
```
gabrielagabriel, line 12: I know, naming things is pretty hard.. but why not associate this with your components following a compound pattern? eg: `Place.Label`

src/newItinerary/internals/ItineraryItem.style.tsx
```diff
@@ -0,0 +1,21 @@
+import styled from 'styled-components'
+
+import { color } from '../../_utils/branding'
+
+export const StyledItineraryItem = styled.li`
+  min-height: 32px;
```
gabrielagabriel, line 6: could use `const ELEMENT_HEIGHT = 32` previously defined

src/newItinerary/internals/ItineraryItem.style.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+  > div,
+  > a {
```
gabrielagabriel, line 9: avoid using tags

src/newItinerary/internals/ItineraryItem.style.tsx
```diff
@@ -0,0 +1,21 @@
...
+    display: flex;
+  }
+
+  a {
```
gabrielagabriel, line 13: avoid using tags to target elements

src/newItinerary/internals/Line.style.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+import { color, space } from '../../_utils/branding'
+import { Lines } from '../Lines'
+
+export const StyledLineWrapper = styled.div`
+  position: relative;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  min-height: 32px;
```
gabrielagabriel, line 11: Could use `const ELEMENT_HEIGHT = 32` previously declared

src/newItinerary/internals/Line.style.tsx
```diff
@@ -0,0 +1,63 @@
...
+  width: 10px;
```
gabrielagabriel, line 12: Magic number that not follow multiple of 8 (base number)

src/newItinerary/internals/Line.style.tsx
```diff
@@ -0,0 +1,63 @@
...
+const lineToBackground: Readonly<{ [key in Lines]: string }> = {
+  [Lines.NONE]: 'transparent',
+  [Lines.ACTIVE]: color.midnightGreen,
+  [Lines.INACTIVE]: color.gray,
+  [Lines.CONNECTION]: `repeat-y url("data:image/svg+xml;base64,${base64EncodedDash}")`,
+  [Lines.HIDDEN_STOPS]: color.midnightGreen,
+}
+
+export const StyledLine = styled.div<{ line: Lines }>`
+  width: 4px;
```
gabrielagabriel, line 29: could be part of a `border` token and so declared as variable to avoid magic numbers

src/newItinerary/internals/Line.style.tsx
```diff
@@ -0,0 +1,63 @@
...
+  height: 12px;
```
gabrielagabriel, line 30: Previously 10 now 12. Must discuss with design. maybe they need help to spot inconsistency

src/newItinerary/internals/Line.style.tsx
```diff
@@ -0,0 +1,63 @@
...
+  &:last-child {
+    flex: 1;
+  }
+
+  // For hidden stops, render rounded bars with some margins.
+  ${({ line }) =>
+    line === Lines.HIDDEN_STOPS &&
+    `
+    &:first-child {
+      height: 8px;
```
gabrielagabriel, line 42: now 8. Is this a progression: 8, 10, 12...

src/newItinerary/internals/Line.style.tsx
```diff
@@ -0,0 +1,63 @@
...
+      margin-bottom: 6px;
```
gabrielagabriel, line 43: should use spacing so (4 or 8)

src/newItinerary/internals/Line.style.tsx
```diff
@@ -0,0 +1,63 @@
...
+      border-radius: 0 0 4px 4px;
+    }
+
+    &:last-child {
+      margin-top: 12px;
```
gabrielagabriel, line 48: should use space :S

src/newItinerary/internals/Line.style.tsx
```diff
@@ -0,0 +1,63 @@
...
+      border-radius: 4px 4px 0 0;
+    }
+  `}
+`
+
+export const StyledBullet = styled.div`
+  // Use absolute position to make sure the bullet goes over the line nicely.
+  position: absolute;
+  top: 12px;
```
gabrielagabriel, line 57: should be calc or variables: why 12?

src/newItinerary/internals/Line.style.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  svg {
```
gabrielagabriel, line 59: avoid targeting tags

gabrielagabriel (commented):
Overall comment: would be great to discuss in chapter our "style" of exposing **styled** elements

Tuxosaurus:
Looks good to me overall, the logic is nice. And thanks for the change on Proximity!

About using `TheVoice`, I'd say `Itinerary` as a component should not handle the title. It's another component, sits on top of `Itinerary` without any specific change to its display or positioning so I'd remove the `headline` prop altogether.

I believe it is currently used on the Ride Details page only and since it's not crawled that would technically be OK. But it's also weird that we use it as an H1 for a whole page about a ride and yet that title only has the date, so IMO, regardless of SEO matters, it is a questionable use for `TheVoice` that does not meet the Pixar specs...

So yeah, I'd rather remove it from `Itinerary`

src/proximity/Proximity.style.tsx
```diff
@@ -1,16 +1,20 @@
...
 export const StyledProximity = styled.div`
-  display: inline-block;
+  display: flex;
   line-height: 0;
-
-  svg {
-    margin-top: ${space.s};
-  }
+  align-items: center;
+  padding-top: ${space.s};
```
wakooka, line 10: Have you checked if theses changes are  breaking the current `Itinerary`?

src/newItinerary/internals/Line.style.tsx
wakooka: We don't have 12 in `space` 😬 (for no good reasons)

src/newItinerary/internals/Line.style.tsx
wakooka: I'm not a fan with of factorizing things that are not (yet) in the design. Scoping this `width` here make sense to me, because it's not related to any other component.

src/newItinerary/internals/ItineraryItem.style.tsx
```diff
@@ -0,0 +1,21 @@
...
+    -webkit-tap-highlight-color: ${color.tapHighlight};
+    -webkit-touch-callout: none;
```
wakooka, line 15: It might be a good time ro reassess the usefulness of these properties

src/newItinerary/internals/ItineraryItem.style.tsx
wakooka, line 9: Not sure about the side-effect if any, but this should work
```suggestion
  > * {

File: src/newItinerary/internals/Line.style.tsx (hunk 9 above)
Comment (mcellier) [reply src/newItinerary/internals/Line.style.tsx:29]: Put a `LINE_WIDTH` constant at the top of the file

File: src/proximity/Proximity.style.tsx (hunk 16 above)
Comment (mcellier) [reply src/proximity/Proximity.style.tsx:10]: Yes that's good

File: src/newItinerary/internals/ItineraryItem.style.tsx (hunk 17 above)
Comment (mcellier) [reply src/newItinerary/internals/ItineraryItem.style.tsx:16]: @Tuxosaurus, do you have any idea?

File: src/newItinerary/Place.tsx (hunk 3 above)
Comment (mcellier) [reply src/newItinerary/Place.tsx:13]: I'll let you add a topic in the chapter. I just reused the naming "convention" we have everywhere else.

File: src/newItinerary/internals/ItineraryItem.style.tsx (hunk 17 above)
Comment (Tuxosaurus) [reply src/newItinerary/internals/ItineraryItem.style.tsx]: No, but caniuse is pretty clear about that:
https://caniuse.com/?search=callout

It may be worth keeping tap-highlight though:
https://caniuse.com/?search=tap-highlight-color

Looks like it's used on all Items

Comment (simonrelet) [reply src/newItinerary/internals/ItineraryItem.style.tsx]: This is what I found on the subject.

[**`-webkit-tap-highlight-color`**](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-tap-highlight-color)

> [6.2 Disable highlighting](https://medium.com/appscope/designing-native-like-progressive-web-apps-for-ios-1b3cdda1d0e8)
>
> 1_wnXeYxpcqK6KwgfVTNRDKQ
>
> When tapping a link in the iOS web browser, a grey box appears around the element. Although there is no simple way of disabling this effect, you may change the highlighting color to transparent, and thus effectively make it disappear. To do this for your PWA, set the `-webkit-tap-highlight-color` property to `transparent` for the desired elements (or assign it to the body element to disable link highlighting for all elements.)

Basically it's a `:active` for iOS Safari.

[**`-webkit-touch-callout`**](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-touch-callout)

> [6.3 Disable callouts](https://medium.com/appscope/designing-native-like-progressive-web-apps-for-ios-1b3cdda1d0e8)
>
> image
>
> If you tap and hold on an element in the iOS browser, it will open a callout menu (like the one above.) To disable this effect on iOS, set the `-webkit-touch-callout` property to `none` for the desired elements. Again, to disable the effect for all elements, assign the property to your body element.

Comment (mcellier) [reply src/newItinerary/internals/ItineraryItem.style.tsx]: As always, thanks @simonrelet for your explanations! 
Based on those, I think we can get rid of `-webkit-touch-callout` as it goes against user's habits. I think we can leave the other one though.

File: src/newItinerary/Itinerary.story.mdx [hunk 18]
```diff
@@ -0,0 +1,219 @@
...
+        aria-label="Pickup at 09:00 at Gare de Lyon, Paris"
+        time="09:00"
+        label="Paris"
+        subLabel="Gare de lyon"
+      />
+      <HiddenStops label="2 stopovers" aria-label="Click to show intermediate stops">
+        <HiddenStop label="Auxerre" aria-label="Bus stops in Auxerre" />
+        <HiddenStop label="Dijon" aria-label="Bus stops in Dijon" />
+      </HiddenStops>
+      <Place time="12:00" label="Lyon" subLabel="Gare Perrache" aria-label="Drop off the vehicle at 12:00 at Gare Perrache, Lyon" />
```
nickoola, line 69: `aria-label` on `Place` should only content complementary text (Arrival/Pickup/Dropoff/arrival) as other children in the `<li>` will be readed. Otherwise we would had to hide children to avoid redundant infos (`aria-hidden`).

src/newItinerary/HiddenStops.tsx
```diff
@@ -0,0 +1,46 @@
...
+  prevLine,
+  nextLine,
+  children,
+  label,
+  ...props
+}: HiddenStopsProps) => {
+  const [hidden, setHidden] = useState(true)
+
+  return (
+    <StyledHiddenStops
```
nickoola, line 26: After testing the collapse part with VoiceOver, I remark it was read 2 times and the UI doesn't allow us to follow those W3C example https://www.w3.org/WAI/GL/wiki/Using_the_WAI-ARIA_aria-expanded_state_to_mark_expandable_and_collapsible_regions as the toggle button disappear after click.
So I suggest to keep the hide/show logic for visual user only and let the sublist alway readable by screen reader, without any button purpose.
So I suggest to do:
- On `StyledHiddenStops ` remove `role="button"`
- On `StyledStopsCount as="button"` add `aria-hidden="true"`
- On the sublist `<ul>` remove `aria-hidden={hidden}`

mcellier:
Tested browsers:
- Chrome desktop Mac OS
- Chrome android
- Safari desktop 
- Edge
- IE 11

Capture from IE11:
Capture d’écran 2021-02-15 à 17 01 50

src/newItinerary/Itinerary.story.mdx
```diff
@@ -55,26 +55,25 @@ import { Addon, Connection, HiddenStop, HiddenStops, Itinerary, Lines, Place } f
 <Canvas>
   <Story name="Ride details - Bus">
     <Itinerary aria-label="Your ride from Paris to Marseille">
-      <Addon label="Lille" aria-label="Bus comes from Lille" />
+      <Addon label="Lille" aria-label="Bus comes from" />
       <Place
-        aria-label="Pickup at 09:00 at Gare de Lyon, Paris"
+        aria-label="Pickup:"
```
nickoola, line 60: Thanks for the content update 👍

nickoola (approved):

src/newItinerary/internals/Line.style.tsx
```diff
@@ -0,0 +1,65 @@
...
+  [Lines.NONE]: 'transparent',
+  [Lines.ACTIVE]: color.midnightGreen,
+  [Lines.INACTIVE]: color.gray,
+  [Lines.CONNECTION]: `repeat-y url("data:image/svg+xml;base64,${base64EncodedDash}")`,
+  [Lines.HIDDEN_STOPS]: color.midnightGreen,
+}
+
+export const StyledLine = styled.div<{ line: Lines }>`
+  width: ${LINE_WIDTH};
+  height: calc(${LINE_HEIGHT} + 1px); // Adding 1 magic px to have the line go under the bullet
```
mcellier, line 34: Without the margin:
Capture d’écran 2021-02-15 à 18 37 41

I didn't update the LINE_HEIGHT constant as it's used in various computations (that work as it)

wakooka (approved):

src/newItinerary/HiddenStops.unit.tsx
```diff
@@ -0,0 +1,47 @@
...
+      <HiddenStops label="2 stops">
+        <HiddenStop label="Paris" />
+        <HiddenStop label="Lyon" />
+      </HiddenStops>,
+    )
+    expect(screen.getByText('2 stops')).toBeInTheDocument()
+    expect(screen.getByRole('button')).toBeInTheDocument()
+
+    // Toggle stops.
+    // Note: aria-ignore makes .toBeInTheDocument() return false
```
Tuxosaurus, line 20: But you aren't using `aria-ignore` anymore, are you?

Tuxosaurus (approved):

gabrielagabriel (approved):

src/newItinerary/HiddenStops.unit.tsx
mcellier: `aria-hidden`*

PR #1093, opened 2021-02-12
Fix CheckCircleIcon display when it has a title

mcellier:
## Description

`nth-child(2)` worked as long as we didn't had a title. Otherwise, the element we wanted to style is ... third.

## What has been done

Use adjacent selector instead

## How it was tested

Kairos

### DIFF
--- a/src/icon/checkCircleIcon.tsx
+++ b/src/icon/checkCircleIcon.tsx
@@ -35,7 +35,7 @@ const StyledCheckCircleIcon = styled(CheckCircleIcon)`
     fill: ${props => (props.iconColor ? props.iconColor : color.green)};
   }
 
-  path:nth-child(2) {
+  path + path {
     fill: ${color.white};
   }
 `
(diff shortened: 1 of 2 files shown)


src/icon/checkCircleIcon.tsx
```diff
@@ -35,7 +35,7 @@ const StyledCheckCircleIcon = styled(CheckCircleIcon)`
     fill: ${props => (props.iconColor ? props.iconColor : color.green)};
   }
 
-  path:nth-child(2) {
+  path + path {
```
gabrielagabriel, line 38: sounds risky to have only the 2nd path on white.

PerrineBoissieres (approved):

src/icon/checkCircleIcon.tsx
wakooka: Should we use a classname in this case?

wakooka (approved):

PR #1097, opened 2021-02-15
[BBC-8404] Fix ScreenModal spacing

gabrielagabriel:
## Description

https://www.loom.com/share/a39ebeaa538d49029a4f0e9d38b042b5?sharedAppSource=personal_library

## What has been done

- Default story with Screen Layout without behavior and also the ability to use text knobs to interact with children

## Things to consider

- Screenshots updates on the App :(

## How it was tested

- SB


gabrielagabriel:
I'm pinguing identity because this bug has been on User Squad run board since ever

guillaumewuip (approved):

src/successModal/SuccessModal.style.tsx
```diff
@@ -90,7 +93,7 @@ const SuccessTitle = styled.h1`
   flex: 1;
   justify-content: center;
   margin: 0; /*reset h1*/
-  padding: ${space.xl} 0 32px; /* TheVoice spacing*/
+  padding: ${space.xl} ${space.xl} 32px; /* TheVoice spacing*/
```
wakooka, line 96: Should we use `TheVoice` instead?

src/successModal/SuccessModal.style.tsx
```diff
@@ -81,6 +81,9 @@ const Content = styled.div`
 
   @media (${responsiveBreakpoints.isMediaLarge}) {
     height: auto;
+    /* gutter */
+    margin-left: ${space.l};
+    max-width: 537px; /*Magic number from spec*/
```
wakooka, line 86: Hmm it's hard to tell if this is the value we should rely on because they use an arbitrary screen size.
Can you double check with design?

wakooka (approved):

src/successModal/SuccessModal.style.tsx
gabrielagabriel: Resolved here: https://github.com/blablacar/ui-library/pull/1097/commits/f2db1bca1397f671080e9a6cbc21ada0d51a2028#diff-8414a8f16a238f7ff57de8829846aee5a2e4afda415090dd97528d8dadc09785R92

src/successModal/SuccessModal.style.tsx
gabrielagabriel: Checked with design and used component sizing instead

PR #1104, opened 2021-02-19
Fix SelectField and PhoneField down arrow

nickoola:
## Description

Fix unclickable down arrow on `SelectField` & `PhoneField`

## Things to consider

- Keep the `<select>` tag above the `<svg>` tag (higher z-index)
- Keep the `<select>` tag transparent in case of error display to let the `<svg>` down arrow visible

## Demo
With the fix:
https://www.loom.com/share/a0333adb84af432d805beaa8d9984320

## How it was tested

Local environment with ui-library only (Chrome/FF/Safari)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **[FIX]** Fix `SelectField` and `PhoneField` unclickable down arrow
 [...]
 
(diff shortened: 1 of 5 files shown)


pith (approved):

pith:
Is it still a draft?

Tuxosaurus (approved):

PR #1111, opened 2021-02-25
rename HighlightSection props for improved usage

Tuxosaurus:
Changed the name of the highlighted items so it's less tied to product notions.

The highlights in HighlightSection were split in "axes" and "destinations", with "axes" being required.
It happens that the product now features blocks with only "destinations" so to make it less confusing, I'm renaming "axes => main" and "destinations = other" (you may suggest better names!)

I've also updated the data we use for tests and stories as it was confusing me a bit. Now test "destinations" are Spanish cities.

## How it was tested

Locally on Firefox MacOS

### DIFF
--- a/src/layout/section/highlightSection/HighlightSection.story.mdx
+++ b/src/layout/section/highlightSection/HighlightSection.story.mdx
@@ -3,7 +3,7 @@ import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs/blocks'
 import { RideAxis } from '../../../_utils/rideAxis'
 import { HighlightSection } from './index'
 
-export const axes = [
+export const featured = [
   {
     id: '11',
     label: 'Toulouse',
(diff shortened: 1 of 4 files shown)


src/layout/section/highlightSection/HighlightSection.story.mdx
```diff
@@ -3,7 +3,7 @@ import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs/blocks'
 import { RideAxis } from '../../../_utils/rideAxis'
 import { HighlightSection } from './index'
 
-export const axes = [
+export const main = [
```
gabrielagabriel, line 6: featured

gabrielagabriel (commented):
more

gabrielagabriel (approved):
Thanks for the update. I would go for "featured and more" terms, as you have used previously on the original code

mcellier (approved):

src/layout/section/highlightSection/HighlightSection.story.mdx
Tuxosaurus: Agreed!

Tuxosaurus:
> Thanks for the update. I would go for "featured and more" terms, as you have used previously on the original code

I think I'll use "featured" and "optional". I'm not 100% comfortable with "more" since it can be confused with the last "See more results" item that closes the list (and maybe the "see more/less" button)

PR #1112, opened 2021-02-26
Normalize TimePicker

mcellier:
## Description

Horizontal normalization of `TimePicker` component. 

Capture d’écran 2021-02-26 à 16 52 37
Capture d’écran 2021-02-26 à 16 52 48

## What has been done

- Horizontal normalization
- Removed unused (and un-specified) `small` prop
- Fixed hover color when disabled

## Things to consider

Main change: component should take 100% of screen width on small devices, we were fixing it at 320px. Now this size is applied on large devices, and the component centers itself.

## How it was tested

Storybook + time step in publication flow


src/timePicker/TimePicker.style.tsx
```diff
@@ -53,6 +64,10 @@ export const StyledTimePicker = styled.div`
     border-color: transparent;
   }
 
+  &[aria-disabled='true']:hover {
+    background-color: ${color.white};
+  }
```
wakooka, line 69: not sure to understand this rule?

wakooka (approved):

src/timePicker/TimePicker.style.tsx
mcellier: We had an issue when the field is disabled, we were still applying a gray background on hover. The text also being gray, we just saw a gray block

wakooka: It seems that this `disabled` prop does not implemented in the design system and that we don't use it on SPA side. I'll be in favor of removing this. WDYT?

mcellier: Do we have a disabled state for all form elements? For consistency's sake I'll like to have the same everywhere, but as we this TimePicker only appears alone in One Action screens I guess we can remove it. @gabrielagabriel WDYT?

src/timePicker/TimePicker.style.tsx
```diff
@@ -1,26 +1,37 @@
...
-    font-weight: 500;
-    line-height: 1;
-    background-color: ${color.white};
-    color: ${color.midnightGreen};
-    border: 1px solid ${color.gray};
-    border-radius: 2em;
-    box-sizing: border-box;
-    cursor: pointer;
-    padding: ${inputBorderSize.focus};
+    left: 50%;
```
gabrielagabriel, line 14: `margin: auto` does make the job.

src/timePicker/TimePicker.style.tsx
```diff
@@ -1,26 +1,37 @@
...
   }
+`
+
+export const StyledTimePicker = styled.div`
+  position: relative;
+  width: 100%;
+  display: inline-flex;
+  justify-content: center;
+  align-items: center;
+  font-size: 4em;
```
gabrielagabriel, line 25: This font-size doesn't follow our scale. Do you think you could come closer by using `brand` or `xxl` sizes?

src/timePicker/TimePicker.style.tsx
gabrielagabriel: Its indeed an global issue on realign libraries. PIXAR doesn't have hover, focus and disabled states. 
I think we can treat with other components (when spec will be created)

gabrielagabriel (approved):

src/timePicker/TimePicker.style.tsx
mcellier: It didn't, that's why I used this

src/timePicker/TimePicker.style.tsx
mcellier: Pixar specs says its xxl (82px). But it doesn't fit on small screens, meaning we should apply the following rule:

`It can’t go on more than 1 line. If data is too long we reduce the font size (see edge case).`

I'll leave this for the realign libraries epic. I'll create a jira

PR #1113, opened 2021-02-26
Rename title prop on Hint, Profil, TripCard and QrCard components

nickoola:
## Description

Rename the `title` prop to `mainTitle` on components which use `A11yProps`.
In  `Hint`, `Profil`, `TripCard` and `QrCard` cases the `title` prop is not used for html `title="Complementary info"` attribute. So it block `A11yProps` enhancements.

The goal in a following PR is to be able to enhance the `A11yProps` by adding the `title` prop for html `title` attribute.

## How it was tested

Local environment with ui-library


mcellier:
What about `label` instead of `mainTitle`?

nickoola:
> What about `label` instead of `mainTitle`?

I think `label` can conflict in other component cases and not enough specific as the prop concern the main text info.
Probably better with `mainText`, `mainInfo` or `mainInfoText` that could be reuse for every component.
What do you think @gabrielagabriel?

mcellier (approved):

gabrielagabriel:
I'm taking the point as a subject to Pixar Chapter to align with other platforms. 
If you are in a hurry you can merge that and we talk about the subject later on **Realign Libraries Topic**

nickoola:
> I'm taking the point as a subject to Pixar Chapter to align with other platforms.
> If you are in a hurry you can merge that and we talk about the subject later on **Realign Libraries Topic**

@gabrielagabriel Ok thanks for your answer 👍

PR #1114, opened 2021-03-01
Normalize DatePicker

mcellier:
## Description

DatePicker horizontal normalization 

Capture d’écran 2021-03-01 à 09 50 47

## How it was tested

Storybook + kairos (publish flow, edit flow, search page)

### DIFF
--- a/src/datePicker/DatePicker.style.tsx
+++ b/src/datePicker/DatePicker.style.tsx
@@ -1,19 +1,19 @@
 import styled from 'styled-components'
 
-import { color, font, space } from '../_utils/branding'
+import { color, font, horizontalSpace, space } from '../_utils/branding'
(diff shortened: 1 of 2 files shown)


wakooka (approved):

src/datePicker/DatePicker.style.tsx
```diff
@@ -124,12 +124,12 @@ export const StyledDatePicker = styled.div`
   &.kirk-datepicker-horizontal .kirk-datepicker-previous-month {
     position: absolute;
     top: -12px;
```
gabrielagabriel, line 126: can we convert that into a variable to understand where this value comes from?

gabrielagabriel (approved):

PR #1118, opened 2021-03-01
[BBC-12765] Paragraph should accept ReactNode

guillaumewuip:
> BBC-12765 

## Description

This PR update `Paragraph` to accept ReactNode, which is usefull to render a paragraph with link inside.

But! As the expandable feature of `Paragraph` (when `isExpandable` is `true` only) use `substring` on the children,
I keep the `children: string` typing when `isExpandable` is `true`.

3 commits in this PR:
- 1st and 2nd about refactoring the component to use hooks and testing-library
- 3rd about enabling `ReactNode` for simple `Paragraph`'s `children`

Out of scope:
- accessibility updates: see https://blablacar.atlassian.net/browse/BBC-12577

## TESTED
UT

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 [...]
 
+- **[UPDATE]** Allow `Paragraph` to receive `ReactNode`
+
 # v49.0.0 (25/02/2021)
 
 - **[UPDATE]** Horizontal Normalization on `SearchRecap`
--- a/src/paragraph/Paragraph.tsx
+++ b/src/paragraph/Paragraph.tsx
@@ -1,4 +1,4 @@
-import React, { PureComponent } from 'react'
+import React, { useState } from 'react'
 import cc from 'classcat'
 
 import { Button, ButtonStatus } from '../button'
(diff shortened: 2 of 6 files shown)


src/paragraph/Paragraph.tsx
```diff
@@ -7,59 +7,68 @@ import { StyledParagraph } from './Paragraph.style'
...
-        {readMoreButton}
-      </StyledParagraph>
-    )
+  return (
+    <StyledParagraph className={cc(className)} role="presentation">
+      <Text itemProp={itemProp} tag={TextTagType.PARAGRAPH} newlineToBr>
+        {content}
+      </Text>
+      {!isExpanded && (
+        <Button status={ButtonStatus.UNSTYLED} onClick={expandParagraph} className="mt-s">
```
mcellier, line 60: We should avoid these kind of classname here. Can you remove this `mt-s` and add the margin in `Paragraph.style.tsx`

mcellier (approved):
You need to update the CHANGELOG too

src/paragraph/Paragraph.tsx
```diff
@@ -7,59 +7,68 @@ import { StyledParagraph } from './Paragraph.style'
...
-      <StyledParagraph className={cc(className)} role="presentation">
-        <Text {...(itemProp && { itemProp })} tag={TextTagType.PARAGRAPH} newlineToBr>
-          {content}
-        </Text>
-        {readMoreButton}
-      </StyledParagraph>
-    )
+  return (
+    <StyledParagraph className={cc(className)} role="presentation">
+      <Text itemProp={itemProp} tag={TextTagType.PARAGRAPH} newlineToBr>
```
gabrielagabriel, line 56: Since you are on the refactor mode, could you please consider using typographic components `TextBody` instead of `Text`. 

```
<TextBody as="p" />
```

You don't need the `newlinetoBr` attr since  `TextBody` already treat breaking lines with CSS

`white-space: pre-line;`

src/paragraph/Paragraph.tsx
```diff
@@ -7,59 +7,68 @@ import { StyledParagraph } from './Paragraph.style'
...
+export type ParagraphExpandableProps = ParagraphBaseProps &
+  Readonly<{
+    isExpandable: true
+    expandLabel: string
+    children: string
+  }>
+
+export type ParagraphProps = ParagraphSimpleProps | ParagraphExpandableProps
+
+const truncateText = (text: string) => `${text.substring(0, DEFAULT_MAX_CHAR_SIZE)}…`
```
gabrielagabriel, line 29: Since you are refactoring this component, could you please review #a11y with @nickoola and behavior with @Tuxosaurus  for SEO.

Thanks :)

src/paragraph/Paragraph.tsx
gabrielagabriel: Missing some ARIA controls. Please check with @nickoola  to help you out

src/paragraph/Paragraph.tsx
nickoola: I think this PR is not a refacto but the ability to use inline node for the text part.
The expand mechanism for a11y can be improved but require discussion with Pixar as not consistent between components. We need to convince designers to keep the expand button instead of hide it after click.
Here is the corresponding jira for this task [BBC-12577](https://blablacar.atlassian.net/browse/BBC-12577) 
So for A11y no change asked on this PR.

src/paragraph/Paragraph.tsx
nickoola: @gabrielagabriel you right, there is also the focus to manage after click on the expand button.
But the a11y issue here is that we hide the button after click. 
Before adding some stuff for a11y we should ask Pixar (Jonas) if we can update the specs to keep the expand button visible. Like for the `HighlightSection`.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 [...]
 
+- **[NEW]** Allow `Paragraph` to receive `ReactNode`
```
gabrielagabriel, line 5: Update
```suggestion
- **[UPDATE]** Allow `Paragraph` to receive `ReactNode`
```

gabrielagabriel (approved):

pith (approved):
Nice refactoring 👏

PR #1127, opened 2021-03-03
Use new itinerary in TripCard + remove microdata

mcellier:
## Description

Use new `Itinerary` in `TripCard`

Capture d’écran 2021-03-03 à 13 58 52

## What has been done

See related FADR.
I had to update TripCard related sections + some example pages, leading to a big number of impacted files.

## Things to consider

A PR will be opened on kairos side to create the seo component.

## How it was tested

Storybook + kairos


src/tripCard/TripCard.style.tsx
```diff
@@ -53,10 +53,8 @@ export const StyledTripCard = styled(Card)`
 
   & .kirk-tripCard-itinerary {
     min-width: 0; /* reset flex default to allow shrinking below content width */
-  }
-
-  &.kirk-tripCard--with-price .kirk-tripCard-itinerary {
-    padding-right: ${space.m};
+    position: relative;
+    left: -${space.xl};
```
mcellier, line 57: As we did for other component, the negative margin is not a nice solution but at least consistent. Once we finish the normalisation, we'll remove all !important stuff and be able to override styles nicer.

src/tripCard/TripCard.example.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react'
+
+import { WarningIcon } from '../icon/warningIcon'
+import { Address, Itinerary } from '../newItinerary'
+import { TripCard } from '.'
+
+// Needs to be outside of the .mdx file, cause the parser changes object types
+// and mess with the iti display
```
pith, line 8:
```suggestion
// and mess with the itinerary display
```

src/tripCard/TripCard.example.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react'
+
+import { WarningIcon } from '../icon/warningIcon'
+import { Address, Itinerary } from '../newItinerary'
+import { TripCard } from '.'
+
+// Needs to be outside of the .mdx file, cause the parser changes object types
```
pith, line 7: So there are no possibility to play with knobs?

guillaumewuip (approved):

src/tripCard/TripCard.style.tsx
pith: Could add that in a css comment? It would be nice to explain why we need this.

src/tripCard/TripCard.tsx
```diff
@@ -40,7 +38,7 @@ export type Driver = User & {
 export type TripCardProps = A11yProps &
   Readonly<{
     href: string | JSX.Element
-    itinerary: Place[]
+    itinerary: JSX.Element
```
pith, line 41: I'm sorry to reopen a debate, but I'm wondering if we could keep a strong typing here (to allow only Itinerary). I know it's meant to add meta tags, but as our microdata info are not about the itinerary but about the trip itself, why not have an `SEOTripCard`, instead of a SEOItinerary. What do you think?

pith: Or the TripCard could have a microdata props which would accept an array of meta tags.

src/tripCard/TripCard.unit.tsx
```diff
@@ -21,22 +22,12 @@ function createPassengers(count: number): User[] {
...
-      },
-      {
-        mainLabel: 'Bordeaux',
-        subLabel: 'Gare Bordeaux Saint-Jean',
-        time: '12:00',
-        isoDate: '2017-12-11T12:00',
-        distanceFromPoint: '8km',
-      },
-    ],
+    itinerary: (
```
pith, line 25: Could you add some tests to check that the itinerary is displayed? I see no test related to it in this file.

src/tripCard/TripCard.style.tsx
wakooka: Should we add a comment in the code to easily identify/grep the negative margin we will need to update once all components are normalized?

src/tripCard/TripCard.tsx
```diff
@@ -91,15 +88,11 @@ export const TripCard = (props: TripCardProps) => {
     price,
     flags = {},
     titles = {},
-    metaUrl = null,
     statusInformation = null,
     badge = null,
     title = null,
   } = props
   const a11yAttrs = pickA11yProps<TripCardProps>(props)
-  const departure = itinerary[0]
```
pith, line 100: Thanks for removing this! 🙏

wakooka (approved):

src/tripCard/TripCard.example.tsx
mcellier: Not with SmartKnobs. But we can add manual knobs. It was broken on local storybook

src/tripCard/TripCard.tsx
mcellier: Take a look at the kairos PR first and we'll discuss about it after

wakooka: > Or the TripCard could have a microdata props which would accept an array of meta tags.
If we do this, it's going to be hard to change if in the future we need micro data without a `TripCard`. In both cases we're using props to render microdata inside a UI components which is not great but I would go with the smallest component unit possible which is `Itinerary`. Even though the current solution has its own drawbacks (typing), I like it more because it helps moving forward with the `NewItinerary` which is the main goal here and we can think of another scalable solution after that.

PR #1132, opened 2021-03-05
update Menu and ItemsList to accept single child item

Tuxosaurus:
## Description

We're soon going to update the TopBar component and as I'm investigating what could be done I stumbled upon potential uses cases where we only have 1 entry in a menu, or at least that we render multiple items in a single component that **Menu** should accept.

```
<Menu>
    <ItemsRenderedThroughAnotherComponent />
</Menu>
```

## How it was tested

Locally on Firefox MacOS with canary release


src/itemsList/ItemsList.tsx
```diff
@@ -38,6 +40,22 @@ export const ItemsList = (props: ItemsListProps) => {
   } = props
   const a11yAttrs = pickA11yProps<ItemsListProps>(props)
 
+  function isSingleChild(item: ChildrenType): item is ItemsListChild {
```
mcellier, line 43: You might want to try something like:

```
const items = Array.isArray(children) ? children : [children]
const filteredChildren = children.filter(Boolean)
```

That way you won't have to keep the next 10 lines

src/itemsList/__snapshots__/ItemsList.unit.tsx.snap
```diff
@@ -796,6 +796,326 @@ button:focus:not(.focus-visible).c1 {
 </ul>
 `;
 
+exports[`ItemsList Should render an unordered list with only one element 1`] = `
```
mcellier, line 799: nit: If it's not too long, you might want to migrate this to RTL.

mcellier (approved):

src/itemsList/ItemsList.tsx
Tuxosaurus: Done, I had try this but was misusing isArray. Works fine!

src/itemsList/__snapshots__/ItemsList.unit.tsx.snap
Tuxosaurus: Done

gabrielagabriel (approved):
Great work.
Are you thinking about open a PR to another type of elements, example a list `ul` or a `div` instead?

Tuxosaurus:
> Great work.
> Are you thinking about open a PR to another type of elements, example a list `ul` or a `div` instead?

For now I have no plan but I am thinking of looking at the usage of `nav` first.

PR #1133, opened 2021-03-08
Use transient props in Address component for styled-comp props

pith:
## Description

The `Address` component is currently throwing a lot of warnings due to the `highlighted` props which is only used by styled-component but is not meant to be rendered in the DOM. To remove these warnings this PR uses the [transient props syntax of Styled-component][1].

```
Warning: Received `false` for a non-boolean attribute `highlighted`.

If you want to write it to the DOM, pass a string instead: highlighted="false" or highlighted={value.toString()}.

If you used to conditionally omit it with highlighted={condition && value}, pass highlighted={condition ? value : undefined} instead.
```

[1]: https://styled-components.com/docs/api#transient-props

## What has been done

Prefix the highlighted props with a `$` as explained in the styled-component documentation.

## Things to consider

No change expected, except the removal of the warnings.

## How it was tested

Running storybook locally

image


wakooka (approved):
Neat! I didn't know about this one. Thanks

mcellier (approved):
Didn't know that either. Could you talk a bit about it in the next chapter meeting to explain this to everyone?

gabrielagabriel (approved):
Quite interesting, are this case happening on other components? 
I've seen a component that needed the same kind of treatment that I can't recall right now on some of your PRs @mcellier 

Maybe we could use the props direct on the component itself  (Don't know if it compiles)

```
export const StyledLabel = styled(TextTitleStrong).attrs({highlighted: props.highlighted})<{ highlighted?: boolean }>`
  ${props => (props.highlighted ? `color: ${color.blue}` : '')};
```

Thanks for the tips, though

PR #1136, opened 2021-03-11
Introduce CardsStackSection

mcellier:
## Description

Add `CardsStackSection` to handle horizontal normalisation of TripCards. Also handles margin between them.

localhost_9000_rides (1)

## Things to consider

- The 8px margin is based on Search figma. **Nothing is specified in pixar**.
- Tested in Your Rides, as search results will need a bigger refactoring

## How it was tested

Storybook + Your rides


wakooka (approved):

src/layout/section/cardsStackSection/CardsStackSection.story.mdx
```diff
@@ -0,0 +1,40 @@
...
+      <TripCard {...tripCardConfig} />
+      <TripCard {...tripCardConfig} />
+      <TripCard {...tripCardConfig} />
+    </CardsStackSection>
+  </Story>
+</Canvas>
+
+## Specifications
+
+Used to display stacked TripCards.
```
gabrielagabriel, line 27: IMHO it can be applied to stack any list of components. 

Don't really understand why you need it instead of applying **spacing rules**. We should avoid having new components that are nor Pixar nor Product for now... 

Or... maybe you are treating this as a "Layout"?? hehehe

mcellier: I explained it this morning during stand-up: to normalize TripCards, we can either:

- Put margins on the `TripCard` elements. We can't use the default `normalizeHorizontally` method as it's adding padding, and here it's meant to be outside the component. Moreover, adding those will completely mess up the spacings in `CardsSection` and `CardsGridSection` that we're using (and we'll have to override everything)
- In fact, search results and YourRides are the only cards not using a section, so I though "why not creating a section for those also". On the plus side, no override is needed on existing sections. 

I got approvals from most of the team then.

gabrielagabriel (approved):

PR #1137, opened 2021-03-11
Export itemActionStatus

wakooka:
## Description

`ItemActionStatus` was not exported and we need it on SPA side.


mcellier (approved):
You should add it to `itemAction/index` also

PR #1138, opened 2021-03-11
Allow polymorphism of SubHeader component

mcellier:
## Description

Allow polymorphism for SubHeader component. Will help seach & seo pages to use normalize components with the right margins. 

## What has been done

- Stopped using <Title> for this component
- Use typography component for text

## Things to consider

In theory we might be able to use `forwardedAs` from `styled-components` but i didn't succeed. Any idea?

## How it was tested

Storybook


mcellier:
PS: if that's all right for you, I'll do the same for `TheVoice`

src/subHeader/SubHeader.tsx
```diff
@@ -3,9 +3,10 @@ import React from 'react'
 import { StyledSubHeader } from './SubHeader.style'
 
 export type SubHeaderProps = Readonly<{
-  children: string
+  children: React.ReactNode
+  as?: keyof JSX.IntrinsicElements | React.ComponentType<any> // from styled-components types
```
gabrielagabriel, line 7: this is native on styled-component we should not use it here

mcellier: Didn't work as it's wrapping another styled component. That's why they put the forwardedAs but I couldn't make it work. Any idea?

Tuxosaurus: I remember I tried but couldn't use it out of the box
Screen Shot 2021-03-11 at 17 20 14

Tuxosaurus (approved):

wakooka (approved):

gabrielagabriel (approved):

PR #1141, opened 2021-03-15
import TopBar style from main app

Tuxosaurus:
The current TopBar has lots of style rules in the main app that should come from the ui-library component.
With that PR, only 2 rules remain, waiting for a more important refactoring of the component itself.

Also removed unnecessary style and added close() to the dimmer which is now included along with the drawer.

Tested locally on Firefox MacOS, with canary release

### DIFF
--- a/src/_utils/branding.ts
+++ b/src/_utils/branding.ts
@@ -140,6 +140,14 @@ export const componentSizes = {
     small: '56px',
     large: '72px',
   },
+  headerWidth: {
+    small: '100%',
+    large: '1280px',
+  },
+  drawerWidth: {
+    small: '100%',
+    large: '350px',
+  },
 }
 
 export const modalSize = {
--- a/src/topBar/TopBar.style.tsx
+++ b/src/topBar/TopBar.style.tsx
@@ -1,39 +1,35 @@
 import styled from 'styled-components'
 
-import { color } from '../_utils/branding'
+import { color, componentSizes, responsiveBreakpoints } from '../_utils/branding'
 
-export const StyledTopBar = styled.header`
-  & {
-    position: relative;
-    width: 100%;
-    background-color: ${color.white};
+export const StyledTopBar = styled.header<{ zIndex?: number }>`
+  position: fixed;
+  width: 100%;
+  background-color: ${color.white};
+  display: flex;
+  align-items: center;
+  top: 0;
+  height: ${componentSizes.headerHeight.small};
+  z-index: ${props => props.zIndex || 3}; /* z-index overridden in main application, above drawer */
+
+  .kirk-topBar-inner {
     display: flex;
-    align-items: center;
+    width: 100%;
+    margin: auto;
+    font-weight: 500;
   }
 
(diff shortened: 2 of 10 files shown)


mcellier (approved):

src/topBar/drawer/Drawer.style.tsx
```diff
@@ -1,15 +1,20 @@
...
-export const StyledDrawer = styled.aside`
   &.kirk-drawer--open {
     position: fixed;
     top: 0;
     left: 0;
     bottom: 0;
     right: 0;
-    z-index: 1;
+    z-index: ${props =>
+      props.zIndex || 2}; /* z-index overridden in main application, above dimmer - below topBar */
```
wakooka, line 15: Shouldn't we use the `$zIndex` notation here?

Also I'm not sure we want to have the default to `2` here, we could simply remove the default no?

Tuxosaurus: Changed to use transient props (not used to it).

For the default, I thought I'd rather have defaults that work within the topBar component (as TopBar, Drawer, Dimmer are related) than making it a required prop. No real opinion.

Tuxosaurus: alright, made it a required prop (and also removed unnecessary style and added close() to the dimmer which is now included along with the drawer)

wakooka (approved):

src/_utils/branding.ts
```diff
@@ -140,6 +140,14 @@ export const componentSizes = {
     small: '56px',
     large: '72px',
   },
+  headerWidth: {
+    small: '100%',
+    large: '1280px',
+  },
+  drawerWidth: {
+    small: '100%',
+    large: '350px',
```
gabrielagabriel, line 149: it feels like 375px (iphone8 default)

src/topBar/TopBar.style.tsx
```diff
@@ -1,47 +1,56 @@
...
     transform: translate(-50%, -50%);
     text-align: center;
   }
 
-  & .kirk-topBar-right {
+  .kirk-topBar-right {
     justify-content: flex-end;
   }
+
+  @media (${responsiveBreakpoints.isMediaLarge}) {
```
gabrielagabriel, line 44: nit: apply responsive to the item that is changing behavior to better readability

gabrielagabriel (approved):

src/topBar/TopBar.tsx
```diff
@@ -1,34 +1,25 @@
...
   className?: string
+  zIndex: number
   leftItem?: ReactNode
   rightItem?: ReactNode
   centerItem?: ReactNode
-  fixed?: boolean
-  bgTransparent?: boolean
-  bgShadedTransparent?: boolean
-  innerWrapperClassName?: string
+  hasScrolled?: boolean
```
Tuxosaurus, line 12: I've added this `hasScrolled` prop to bring the "u-separator-bottom" style to ui-library.

PR #1143, opened 2021-03-17
[BBC-12726] Node 15 + NPM 7 + update peer deps

mcellier:
## Description

Upgrade to node 15 / npm 7 to match the target version before intenalizing the lib.

## Things to consider

Also updated peer deps numbers to match kairos.

## How it was tested

Canary release by @guillaumewuip


package.json
```diff
@@ -8,8 +8,8 @@
     "url": "git@github.com:blablacar/ui-library.git"
   },
   "engines": {
-    "node": "^12.16.1",
-    "npm": "^6.13.4"
+    "node": "^15.11.0",
+    "npm": "^7.6.0"
```
guillaumewuip, line 12:
```suggestion
    "node": "^12.16.1 | ^15.11.0",
    "npm": "^6.13.4 | ^7.6.0"
```

Maybe this to be backward compatible?

mcellier: Trying this syntax but it doesn't seems to work. 
I got the following working:

```
"node": ">=12.16.1 <=v15.11.0",
"npm": ">=6.13.4 <=7.6.0"
```

WDYT?

mcellier: 👌

mcellier: You're right. Moreover, the `.nvmrc` ensure the version so we're good IMHO.

guillaumewuip (approved):

PR #1145, opened 2021-03-18
Moved ModalFog component as a generic component

wakooka:
## Description

Moved ModalFog component as a generic component so we can use it anywhere.

Kapture 2021-03-18 at 17 16 46

## What has been done

- Created `Fog`
- Unit tests to make sure we can't click on elements underneath `Fog` when it is in loading state 
 
## Things to consider

- Added a polyfill on `inert` attr for a11y since it's still an experimental feature in all browsers

## How it was tested

Storybook + Unit test

### DIFF
--- a/src/fog/Fog.tsx
+++ b/src/fog/Fog.tsx
@@ -12,6 +12,8 @@ export type FogProps = Readonly<{
 export const Fog = ({ isLoading, children }: FogProps) => (
   <StyledFogContainer>
     <StyledFog $isLoading={isLoading} aria-hidden="true" />
-    <div ref={node => node && isLoading && node.setAttribute('inert', '')}>{children}</div>
+    <div ref={node => node && isLoading && node.setAttribute('inert', '')} aria-hidden={isLoading}>
+      {children}
+    </div>
   </StyledFogContainer>
 )
--- a/src/fog/Fog.tsx
+++ b/src/fog/Fog.tsx
@@ -12,7 +12,13 @@ export type FogProps = Readonly<{
 export const Fog = ({ isLoading, children }: FogProps) => (
   <StyledFogContainer>
     <StyledFog $isLoading={isLoading} aria-hidden="true" />
-    <div ref={node => node && isLoading && node.setAttribute('inert', '')} aria-hidden={isLoading}>
+    <div
+      ref={
+        node => node && (isLoading ? node.setAttribute('inert', '') : node.removeAttribute('inert'))
+        // eslint-disable-next-line react/jsx-curly-newline
+      }
+      aria-hidden={isLoading}
+    >
       {children}
     </div>
   </StyledFogContainer>
--- a/src/fog/Fog.tsx
+++ b/src/fog/Fog.tsx
@@ -1,0 +1,21 @@
+import React, { Fragment } from 'react'
+
+import 'wicg-inert'
+
+import { StyledFog, StyledFogContainer } from './Fog.style'
+
+export type FogProps = Readonly<{
+  isLoading?: boolean
+  children: React.ReactNode
+}>
+
+export const Fog = ({ isLoading, children }: FogProps) => (
+  <StyledFogContainer>
+    <Fragment>
+      <StyledFog $isLoading={isLoading} aria-hidden="true" />
+      <section ref={node => node && isLoading && node.setAttribute('inert', '')}>
+        {children}
+      </section>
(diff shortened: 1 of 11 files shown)


nickoola:
Really cool new component! 
I'll checkout your branch to see if the `inert` attribute is correctly interpreted by screen reader (if it skip content as expected).
For me it's a layout behavior and should be part of Pixar Layout specs. Is it the case in your confluence doc?

src/fog/Fog.unit.tsx
```diff
@@ -0,0 +1,32 @@
...
+  it('should not be able to click on an interactive element', () => {
+    const onClick = jest.fn()
+    render(
+      <Fog isLoading>
+        <button onClick={onClick}>Button</button>
+      </Fog>,
+    )
+
+    const button = screen.getByRole('button')
+    expect(onClick).not.toHaveBeenCalled()
```
wakooka, line 30: Oops I forgot to `fireEvent.click(button)`  that's why the is passing 🤦‍♂️

wakooka:
> Really cool new component!
> I'll checkout your branch to see if the `inert` attribute is correctly interpreted by screen reader (if it skip content as expected).

It should be ok because I'm also adding a `aria-hidden` to true

> For me it's a layout behavior and should be part of Pixar Layout specs. Is it the case in your confluence doc?

For now it's way too early to think about this component as part of the layout specs. The layout specs today are there for having sticky & position logic basically, but it  might evolve in the future in something more detailed.

src/fog/Fog.unit.tsx
wakooka: I updated the test in my last commit, since I'm using `aria-hidden`, the button shouldn't even be "readable"

nickoola:
> > Really cool new component!
> > I'll checkout your branch to see if the `inert` attribute is correctly interpreted by screen reader (if it skip content as expected).
> 
> It should be ok because I'm also adding a `aria-hidden` to true
> 
> > For me it's a layout behavior and should be part of Pixar Layout specs. Is it the case in your confluence doc?
> 
> For now it's way too early to think about this component as part of the layout specs. The layout specs today are there for having sticky & position logic basically, but it might evolve in the future in something more detailed.

Everything works fine with Screen reader.

Just have a question on this extra markup, I remark there is an empty `<div aria-hidden="true">` before the `inert` wrapper:
Screenshot 2021-03-19 at 11 01 54
Could is be removed?

wakooka:
>  Could it be removed?

@nickoola  no, this element is the one on top with the semi-opacity (see gif)

nickoola (approved):

src/fog/Fog.tsx
```diff
@@ -0,0 +1,24 @@
...
+  isLoading?: boolean
+  children: React.ReactNode
+}>
+
+export const Fog = ({ isLoading, children }: FogProps) => (
+  <StyledFogContainer>
+    <Fragment>
+      <StyledFog $isLoading={isLoading} aria-hidden="true" />
+      <div
+        ref={node => node && isLoading && node.setAttribute('inert', '')}
```
simonrelet: Don't we want to remove the `inert` attribute if `isLoading` is `false`?

src/fog/Fog.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export type FogProps = Readonly<{
+  isLoading?: boolean
+  children: React.ReactNode
+}>
+
+export const Fog = ({ isLoading, children }: FogProps) => (
+  <StyledFogContainer>
+    <StyledFog $isLoading={isLoading} aria-hidden="true" />
+    <div ref={node => node && isLoading && node.setAttribute('inert', '')} aria-hidden={isLoading}>
```
simonrelet, line 15: Don't we want to remove the `inert` attribute when `isLoading` is set back to `false`?

simonrelet (approved):

src/fog/Fog.tsx
wakooka: By looking at the devTools it does, but I'm wondering if it's the way storybook renders the component. Maybe I should add a else with an explicite `removeAttribute`

simonrelet: Maybe it's because React set all the attributes during render (remove old ones) and then the ref callback is called.

wakooka: Done 👍

PR #1146, opened 2021-03-19
Allow polymorphism of TheVoice component

mcellier:
## Description

Allow polymorphism for TheVoice component. Will help seach & seo pages to use normalize components with the right margins. 

## What has been done

- Stopped using <Title> for this component
- Use typography component for text

## Things to consider

In theory we might be able to use `forwardedAs` from `styled-components` but i didn't succeed. Any idea?

## How it was tested

Storybook

### DIFF
--- a/src/theVoice/TheVoice.tsx
+++ b/src/theVoice/TheVoice.tsx
@@ -1,21 +1,17 @@
 import React, { ReactNode } from 'react'
-import cc from 'classcat'
 
 import { StyledTheVoice } from './TheVoice.style'
 
 export type TheVoiceProps = Readonly<{
+  children: ReactNode
   id?: string
   className?: string
-  children: ReactNode
   isInverted?: boolean
+  as?: keyof JSX.IntrinsicElements | React.ComponentType<any> // from styled-components types
 }>
 
-export const TheVoice = ({ id, className, children, isInverted }: TheVoiceProps) => (
-  <StyledTheVoice
-    id={id}
-    className={cc([className, { 'kirk-thevoice--inverted': isInverted }])}
-    headingLevel="1"
-  >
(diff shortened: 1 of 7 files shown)


src/theVoice/TheVoice.tsx
```diff
@@ -1,21 +1,17 @@
...
 
 import { StyledTheVoice } from './TheVoice.style'
 
 export type TheVoiceProps = Readonly<{
+  children: ReactNode
   id?: string
   className?: string
-  children: ReactNode
   isInverted?: boolean
+  as?: keyof JSX.IntrinsicElements | React.ComponentType<any> // from styled-components types
```
guillaumewuip, line 10: Maybe can we only allow some tags like `h1 | h2 | h3` ?

mcellier: We're using a default styled-component feature here. We might render it as p, span or div, it doesn't matter IMHO.

Tuxosaurus: IMO if we're gonna add "as" to a component then it should behave like the other instances with that prop, so no restriction I guess.

Tuxosaurus (approved):
Thanks also for the id on Text, I'm guessing we'll need that soon™

src/theVoice/TheVoice.tsx
guillaumewuip: Ok. I'm more in favour of providing only what is needed right now and edit it for new usage latter. Smallest API first for an easy maintenance and robustness in the long term.

For example are we sure we want to allow `span` or `div` here? And even if we want to allow them, is it not more explicit to have them listed in the type like `| 'span' // you can use span here for <reason>` ?

That's not blocking :)

PR #1148, opened 2021-03-23
[PROPIM-797] Banner Section

gabrielagabriel:
## Description

Brand New Component for Marketing purpose 

**Layout:**
BBC_Banner_AppDownload_Final_Mobile_380x155_EN

**WIP**
Screenshot 2021-03-23 at 13 53 53

## What has been done

- Component using ui-library components to avoid technical debt on the App

## Things to consider

- Not Pixar (but use all components from ui-library)
- Further information is requested to understand how to deal with Media (QRCode Card) and Buttons


src/banner/Banner.style.tsx
```diff
@@ -0,0 +1,51 @@
+import styled from 'styled-components'
+
+import { color, componentSizes, space } from '../_utils/branding'
+import { TextDisplay1 } from '../typography/display1'
+import { TextTitle } from '../typography/title'
+
+export const StyledBanner = styled.section`
+  background-color: ${color.boostBlue};
+  padding: ${space.xxl} ${space.xl};
+  height: 19.625rem; //314px magic number
```
Tuxosaurus, line 10: Maybe this should be a `min-height` so that if the title is too long the banner can stretch. I'd favor pixels (height of the phone image + top margin)

src/banner/Banner.style.tsx
```diff
@@ -0,0 +1,51 @@
...
+  background-image: url('./phone-frame.svg');
+`
+/** LAYOUT */
+export const Content = styled.div`
+  display: flex;
+  margin-left: auto;
+  margin-right: auto;
+  max-width: ${componentSizes.largeSectionWidth};
+`
+export const Media = styled.div``
```
Tuxosaurus, line 39: If the `height` is unfixed as I wrote above, we probably need to have `align-self: end;` for Media so it sticks to the bottom

src/banner/Banner.tsx
```diff
@@ -0,0 +1,34 @@
...
+  copy: Copy
+  actions?: CTALinks | CTAButtons
+  qrcode?: QrCardProps
+}>
+
+export const Banner = ({ copy, actions, qrcode }: BannerProps) => (
+  <StyledBanner>
+    <BannerElements.Content>
+      <BannerElements.Body>
+        <BannerElements.Heading as="h1">{copy.heading}</BannerElements.Heading>
```
Tuxosaurus, line 21: Do we know if we want a `h1` here?

src/banner/Banner.tsx
```diff
@@ -0,0 +1,34 @@
...
+        <BannerElements.Paragraph as="p">{copy.paragraph}</BannerElements.Paragraph>
+        {actions && <BannerElements.Actions>{actions}</BannerElements.Actions>}
+      </BannerElements.Body>
+      <BannerElements.Media>
+        {qrcode && (
+          <BannerElements.Frame>
+            <QrCard imageUrl={qrcode.imageUrl} mainTitle={qrcode.mainTitle} />
```
Tuxosaurus, line 28: It seems the `Card` component used by `QrCard` only outputs a `li`, it might need refactoring if we want to use it as a standalone card with a `div`.

gabrielagabriel: Yes.. the DOM isn't what we are looking for here. Mainly the visual part, but I would not include that on our MVP.

src/banner/Banner.style.tsx
gabrielagabriel: I have a tendency to use rem since when you update your font-size base in the browser the bloc keeps the proportion defined.

src/banner/Banner.style.tsx
```diff
@@ -0,0 +1,51 @@
...
+  margin-top: ${space.xl};
+
+  > button:first-child {
+    margin-right: ${space.l};
+  }
+`
+
+export const Frame = styled.div`
+  padding: ${space.l} ${space.xxl} 0 ${space.xxl};
+  background-image: url('./phone-frame.svg');
```
Tuxosaurus, line 30: That's frustrating, I can't get the image to be displayed this way using css. Might need some config twisting out of my knowledge

src/banner/Banner.style.tsx
Tuxosaurus: As you want, it's only been used once in the app.

Also regarding the height, makes me think we don't know what it is supposed to look like on small devices. I imagine either a flex-wrap to flip the phone below the text, or we simply the phone (if someone is browsing the site on their phone, they'd not scan the code - or maybe to show someone else but that's a rare scenario)

PR #1149, opened 2021-03-23
Add title and tabIndex to A11yProps

nickoola:
## Description

Add title and tabIndex to A11yProps

## Demo

**If we add a `title` & `tabIndex` prop to the `Button` component:**
Screenshot 2021-03-23 at 14 42 10

## How it was tested

Unit tests + Storybook + Canary release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **[BREAKING CHANGE]** Add `title` and `tabIndex` props to `A11yProps` interface and `pickA11yProps` function
 - [...]
 
 # v56.0.0 (22/03/2021)
--- a/src/_utils/interfaces.ts
+++ b/src/_utils/interfaces.ts
@@ -11,6 +11,8 @@ export interface CommonFieldsProps {
 export interface A11yProps {
   id?: string
   role?: string
+  title?: string
+  tabIndex?: number
   'aria-label'?: string
   'aria-labelledby'?: string
   'aria-describedby'?: string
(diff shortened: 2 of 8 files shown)


src/itemChoice/ItemChoice.tsx
```diff
@@ -36,7 +36,6 @@ export type ItemChoiceProps = A11yProps &
     onFocus?: (event: React.FocusEventHandler<HTMLElement>) => void
     onMouseDown?: (event: React.MouseEvent<HTMLElement>) => void
     onDoneAnimationEnd?: () => void
-    tabIndex?: number
```
nickoola, line 39: Provided by `A11yProps`

src/button/Button.tsx
```diff
@@ -24,7 +24,6 @@ export type ButtonProps = A11yProps &
     type?: string
     href?: string | JSX.Element
     className?: string
-    title?: string
```
nickoola, line 27: Provided by A11yProps

src/_utils/interfaces.ts
```diff
@@ -11,26 +11,38 @@ export interface CommonFieldsProps {
...
     'aria-describedby',
     'aria-controls',
   ]
-  keys.forEach(k => {
+  const keysWithNumberValue: A11yKeysWithNumberValue[] = ['tabIndex']
+
+  keysWithStringValue.forEach(k => {
+    returnValue[k] = source[k]
+  })
+  keysWithNumberValue.forEach(k => {
```
nickoola, line 43: `title` prop can't have a `number` value for example, so I split key props in two list to solve ts issue.

mcellier: Using something like `source.reduce()` might be a bit nicer

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **[UPDATE]** Add `title` and `tabIndex` props to `A11yProps` interface and `pickA11yProps` function
```
mcellier, line 3: It's more a breaking change, as some typings are being changed

mcellier (approved):

wakooka:
Out of scope of this PR: I think we should change our approach like we're going to do on TextField props closer to HTML attributes and using `InputHTMLAttributes<HTMLInputElement>` type (I have a branch on this hopefully I can open it this week). Using `AriaAttributes` type from React could clean this custom file, and then we can extend most of our components with `AriaAttributes` type

wakooka (approved):

nickoola:
> Out of scope of this PR: I think we should change our approach like we're going to do on TextField props closer to HTML attributes and using `InputHTMLAttributes<HTMLInputElement>` type (I have a branch on this hopefully I can open it this week). Using `AriaAttributes` type from React could clean this custom file, and then we can extend most of our components with `AriaAttributes` type

Yes in the end we should more categorise attribute we want to provide to components instead of a complete list.
It could be a second step.

CHANGELOG.md
nickoola: Yes it'll require some components prop change on Kairos side. So I pass those changes to [BREAKING CHANGES].

src/_utils/interfaces.ts
nickoola: Right it's cleaner now.

PR #1152, opened 2021-03-30
prevent itinerary from shrinking with long names

Tuxosaurus:
Video with problem & fix

https://user-images.githubusercontent.com/373381/113024550-2ad2c280-9187-11eb-8f28-a1088bc510fe.mp4

When a city name was too long and could not be naturally wrapped, it would take all the space, shrinking the time and line items. These 2 should not shrink.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
 # Unreleased
 
+- [...]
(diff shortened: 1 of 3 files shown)


mcellier (approved):

pith (approved):

pith:
Thanks for the fix!

